PR #870, opened 2015-01-26
Assignment permissions on creation

mlangens:


chcholman:
This is looking great! Couple of updates: 

* Let's remove the list item, but keep the indent for the checkboxes
* The sub-checkboxes currently appear immediately after whatever is the new interaction with the Accept Submissions checkbox, so if the box starts as checked, then un-checking it displays the sub-checkboxes. 
* When you return to the page, if the "Accept Submissions" box is marked, I should be able to see the sub-checkboxes. 
*  The settings (link, checkbox, text) here should then alter what is shown in the submission form (submissions/form).

chcholman:
Almost there! We still need the settings in this form to actually affect what's displayed in the submissions form.

chcholman:
@josankapo @mlangens Am I right that this work is already in the assignment submissions work that just got merged?

PR #894, opened 2015-01-28
sets a maxlength for tier and metric inputs

mlangens:


mlangens:
@zazaian I noticed that the entire form currently fails if any validation fails. What's the easiest way to prevent navigation away from the form if a validation fails?

mkzzn:
Hey Max -- I think that should be correct behavior for individual records, but not for all records. Is that what you're noticing? I'm not sure that we want to persist objects that have invalid attributes, but we should be able to persist others that are valid. I'm looking at your branch right now and it seems that you can save other Tiers and Metric objects even if another one happens to be invalid. Could you take another look?  @chcholman, Any thoughts on this?

chcholman:
Hey @mkzzn I think it's the way that the form validation fails that's at issue here. When we tested it yesterday, if you submit the form with data that fails the validation, you are redirected to the form and lose all of the material you entered. Ideally you'd be redirected to the form with the data still present (or be unable to leave).

mkzzn:
Ahhh, you're talking about the submit rubric page, rather than design rubric. I'd actually just been managing the enabled/disabled state of the submit button based on whether or not there was invalid data in the form. If there was invalid data or data missing, submit would be disabled until that data was present. Does that make sense? Should we allow them to submit the form anyway and then mention the errors? Seems like an extra step. At a minimum I'd suggest keeping the form submission disabled and throwing up a JS notification of the existing failed validations.

chcholman:
The latter part makes a ton of sense - the form should be disabled unless the data is present. But isn't the place where you asked  @mlangens to add validation on the design rubric page? IF you write a title that's too long, given the auto save, what's the expected functionality? We are highlighting the field, but I think right now we lose all the data entered if you navigate away. We don't have a submit button so you can't disable that.

mkzzn:
I don't think that's correct re: the design rubric page. I just took a look at the log behavior of a valid Tier when there's an invalid Tier within that same metric, and it seems to update and persist correctly. Is this different than what you're seeing? Are you keeping an eye on the dev log as you're testing here?

chcholman:
[Testing]

User story: As an instructor I need to be able to create rubrics
- I created a new metric and updated the title - characters stopped after 30 characters
- I created a new tier and updated the title - characters stopped after entering 30
[good to go]

User story: As an instructor I want to be able to grade students using rubrics
- I selected the tier with the longest name as their criteria score, and entered text feedback at that metric. Successfully saved
[good to go]

This looks set!

PR #897, opened 2015-02-11
fix issue where rubric was awarding badges for all tiers and metrics reg...

zazaian:
...ardless of which one was actually selected


chcholman:
This is good to go!

PR #898, opened 2015-02-13
2015 02 11  improve mass grading workflow and bugs  tkt#2687

chcholman:


chcholman:
[Testing bug: duplicate student name display in quick grade form]

[User Story] As an instructor, I need to be able to grade all students for a single assignment very quickly
1. I clicked "Quick Grade" from the assignment index page by a blog post in the EDUC222 sample data set
2. The first student "Brandon Angel" appears listed twice 
3. Page refresh: gone
4. Check another assignment - same duplication, same student 

@zaz Will you check into this when you get a chance?

PR #984, opened 2015-05-28
Memcache

mlangens:


mlangens:
Hey @JangoSteve I thought I'd ask you about this here. I'm referencing https://signalvnoise.com/posts/3113-how-key-based-cache-expiration-works because I think this will be the easiest way to implement caching. I have a bunch of log output I wanted to show you, and I've made some gists. 

Here is the first load: https://gist.github.com/mlangens/bc9fe7080cc986437406
The second: https://gist.github.com/mlangens/f302a4a6f379ca7c1a80
And the diff: https://gist.github.com/mlangens/0ebe820182060f3fb348

From what I've seen, any update on these models will expire the existing keys and rewrite. As of yet, I have not seen any stale data in my testing. I just wanted to run this by you before I take off with it!

cc: @chcholman

PR #997, opened 2015-06-12
Assignment show

mlangens:


mlangens:
Local before / after load times
screenshot 2015-06-15 13 42 35
screenshot 2015-06-15 13 41 34

PR #1036, opened 2015-08-06
Forgot password

jwright:
This implements the forgot password functionality for a gradecraft login.

This uses [Sorcery's reset password](https://github.com/NoamB/sorcery/wiki/Reset-password) functionality to send password reset instructions to a registered email address which contains a link with a password reset token. The user visits that link and they can change their password (with a password confirmation). If the passwords match, the user is logged in and taken to their dashboard.

Password reset tokens expire after 8 hours. If the user visits their password reset link after the 8 hours, they are redirected to the password reset page to request a new password reset token.

I added some new validations to the `User` object to ensure that the password is confirmed and the password confirmation exists on all existing users.

All edges cases are tested within a controller spec: https://github.com/UM-USElab/gradecraft-development/blob/be4976c4c31a94c3c70dc92d00ab446cb5bb2457/spec/controllers/passwords_controller_spec.rb

I used the existing password reset email with a few modifications.

This is the password reset form:

screen shot 2015-08-06 at 11 52 27 am

This is the change password form:

screen shot 2015-08-06 at 11 53 00 am

I added a forgot password link within the GradeCraft login tab:

screen shot 2015-08-06 at 11 52 02 am

Note, that the password reset form also exists on the `/login` page. This works as well.

screen shot 2015-08-06 at 11 54 57 am

Also note that none of the flash messages appear on non-authenticated pages. This means that the user does not see the flash notices and errors on the new password page, the root page, and the password change page. I assumed that this is a known bug and if not, an issue should be opened for this.

Closes 33a on the Summer punchlist.


jwright:
@chcholman This is ready for review

chcholman:
Hi Jamie, 

This looks fantastic. Three (tiny) updates to the email: 
- Can you switch it from being "Hello, user.email" to being "Hello, user.first_name"
- Can you check the link embedded in the email? It may be fine in production, but what I got in testing was the following: http:/&#47;gradecraft:3000/passwords/Xybi4XVyWqLs6NBzEk8h/edit
- Can you remove the period at the end of the link, so that if someone tries to copy and paste it they don't accidentally include it? 

Excellent catch on the missing alerts. I'm working on those  layouts given all of the foundation work right now so I'll make sure they get added back in!

Thanks so much,
Cait

jwright:
@chcholman Thanks for the feedback. I made the adjustments. Great catches.

PR #1040, opened 2015-08-08
Activating new accounts

jwright:
This pull request allows staff to create new user accounts (with `/users/new`) without specifying a password for the new user. A random password is generated for the new user and an activation email is sent to the user for them to complete their account.

New users are not allowed to sign up without activating their accounts.

There is a [migration](https://github.com/UM-USElab/gradecraft-development/blob/0d3dad27064b6427d557466eba9db606b4147965/db/migrate/20150806205557_sorcery_user_activation.rb) to add the Sorcery user activation submodule fields to the `User`.

All of the users that were created during the setup are also activated now.

This is the new user create view (with the removal of the password fields):

screen shot 2015-08-07 at 10 24 20 pm

This is the email that the new user gets when they get created:

screen shot 2015-08-07 at 10 24 50 pm

This is the form that the new user goes to in order to activate their account:

screen shot 2015-08-07 at 10 25 14 pm

Closes #33b on on the punch list

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -104,10 +104,15 @@ class UsersController < ApplicationController
   def activated
     @token = params[:token]
     @user = User.load_from_activation_token(@token)
-    @user.update_attributes params[:user]
-    @user.activate!
-    auto_login @user
-    redirect_to dashboard_path, notice: "Welcome to GradeCraft!"
+
+    redirect_to root_path, alert: "Invalid activation token. Please contact support to request a new one." and return unless @user
+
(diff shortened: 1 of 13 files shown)


chcholman:
@jwright Nice! This is looking great. 

I think it may need an additional skip_before_filter :activate at the top of the users controller - I tested by copying the link into a new browser where I wasn't logged in, and was redirected to the homepage. Otherwise it's good to go!

jwright:
@chcholman Nice catch. I updated the controller. This is ready for another review.

chcholman:
Perfect! Merging in.

PR #1041, opened 2015-08-10
Importing users

jwright:
Fixes the student import and sends activation emails to all students that are imported sucessfully.

The import (`/users/import`) was not currently working because it had some (presumably) old code for assigning a `student` role to a `User`. This was changed to build a `CourseMembership` for a `student` instead. **Please review this code carefully as I made some assumptions of who is performing this import and that they have a `current_course`.**

The import now specifies a temporary password for each `student` imported and sends out the activation email for each imported `User`.

The import now also renders a `import_results` view that displays who was imported and if there were any issues with imported users. This used to just blow up the import if the `User` could not be imported. 

This is the imported results page (with a successful imported user):

screen shot 2015-08-10 at 3 39 30 pm

This is the imported results page (if there are any unsuccessful imported users):

screen shot 2015-08-10 at 3 38 47 pm

Closes #33c

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -146,6 +146,14 @@ class UsersController < ApplicationController
       @unsuccessful = []
       CSV.foreach(params[:file].tempfile, headers: true, skip_blanks: true) do |row|
         team = Team.find_by_course_and_name current_course.id, row[4]
+        team ||= Team.create course_id: current_course.id, name: row[4] do |t|
+          unless t.valid?
+            @unsuccessful << \
+              { data: row.to_s, errors: t.errors.full_messages.join(", ") }
+            next
+          end
+        end
+
         user = User.create do |u|
           u.first_name = row[0]
           u.last_name = row[1]
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -1,4 +1,6 @@
 class UsersController < ApplicationController
+  include UsersHelper
+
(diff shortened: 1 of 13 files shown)


jwright:
@chcholman Ready for review

chcholman:
What have you tested in regards to import file structure and type? I've just tried to import a .csv file with and without header rows, and am getting the following error: 
screen shot 2015-08-10 at 4 06 25 pm
screen shot 2015-08-10 at 4 15 20 pm

jwright:
@chcholman I use the csv file format that was there generally there before and it was setup with the following fields (no header rows):

```
first_name, last_name, username, email
```

Does this work? Should we support a different set of fields?

chcholman:
Awesome, that format worked perfectly.

Yes, we should do the following: 
1) Allow for a header row
2) Import an additional fifth column "Team Name". We should look for a team belonging to the course with the name that's present, and if it's not already created it should be added. If it is present, their team membership should be created. 

Thanks!

jwright:
Got it! I will make the adjustments.

jwright:
- [X] Allow headers
- Add team name as last import column
- Create team for `current_course` and `student` if team does not exist for the `current_course`
- Add validation rule for case insensitive unique name with `course` scope

jwright:
@chcholman This is ready for another review. It now creates the team and I extracted the logic of importing into a new object.

chcholman:
Looks perfect, thanks so much @jwright!

PR #1081, opened 2015-08-25
Fix edit profile

jwright:
The ability to update a profile (via `/users/edit_profile`) was broken due to the fact that a password was not required.

In order to update a profile without specifying a password each time, the `password` and `password_confirmation` fields were removed from the parameters if they were not blanked.

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -75,6 +75,7 @@ class UsersController < ApplicationController
     if @user.is_student?(current_course)
       @user.teams.set_for_course(current_course.id, params[:user][:course_team_ids])
     end
+
     @user.update_attributes(params[:user])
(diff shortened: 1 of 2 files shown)


chcholman:
Just tested this by: 

1. Logged in as a hermione.granger@hogwarts.edu
2. Navigated to the 'edit_profile' page. 
3. Tried to upload a new avatar .png file 
4. Clicked submit - was redirected to the form with an error message that my profile could not be successfully updated. 
5. Retested and confirmed that if I do include a password/confirmation, the form does successfully submit. 

Can you take a look @jwright?

jwright:
@chcholman Hmmm...I tried it and it worked for me. Is there something I am not trying? Here is a screencast by using Hermione.

update-profile

PR #1097, opened 2015-08-27
Self log logic

chcholman:


chcholman:
Thanks to @jwright we're switching to using just the select form, no more button!

PR #1102, opened 2015-08-27
Weight fix

josankapo:


chcholman:
Wow do I love the upgrade to the style display on this @josankapo!

PR #1111, opened 2015-08-31
Replace all forms with errors with common partial

jwright:
Uses a common partial (`app/views/layouts/alerts`) to display model errors instead of having each form display a different partial.

This replaces all current views that display model errors without using the common partial.


jwright:
@chcholman Ok. I fixed the issue with assignments. It was an issue with the following code:

```
local_assigns[:model].present?
```

I changed it to the following and it worked :confused::

```
local_assigns.has_key?(:model)
```

I also updated the new/edit forms that contained the form so that the alert layout was not rendered multiple times.

PR #1114, opened 2015-09-02
Image upload validation

jwright:
This adds file size validation to most of the image file uploaders. The following have validations added to them:

* Courses (media)
* Assignments (media and thumbnail)
* Events (media and thumbnail)
* Challenges (media and thumbnail)

The file size validator allows a maximum size validation on the uploaders. In this case, it is 2 MB.

In addition, I also did a major overall and clean up of the uploaders. The duplicate uploaders were eliminated in favor of a base uploader for images and thumbnails. Concerns for the media and thumbnail images were created and used in the respective models.

The views were consolidated as well into partials.


chcholman:
Thanks so much for this overhaul! 

I tried to locally upload two images (both the media image and the thumbnail) to an assignment, and ran into this issue: 

screen shot 2015-09-03 at 6 34 25 am

I have: run the migrations, run bundle install and restarted the server.

jwright:
@chcholman I could not replicate this locally but I believe I know what the issue was regardless. I moved the MiniMagick mixin back to the `ImageUploader` instead of just on the `ThumbnailUploader`. Can you verify that you don't run into the issue anymore?

chcholman:
@jwright helps if you have ImageMagick installed ;)

PR #1117, opened 2015-09-05
Allow Administrators access to all courses

jwright:
This PR adds a course searcher for administrators so that they can find all of the courses from their name for easy switching to that course.

Here is a sample of it in action:

change-course

In addition, a `User` nows has a `admin` boolean attribute which is used to assign the admin role to various users. When a `Course` is created, all admins are added to that course by default with `CourseMembership` attributes. This is done so that administrators have access to all courses.

In order to deploy, please run the following tasks on staging and production:

- [ ] rake db:migrate
- [ ] rake courses:update_admins


chcholman:
@jwright Should rake courses:update_admins work locally?

PR #1121, opened 2015-09-08
Predictor grey section

josankapo:


chcholman:
Will you test this out in EDUC222 (sample data)? I'm seeing the predictor loading note stay in place, and an error on the attempt to load predictor_badges.

PR #1127, opened 2015-09-10
Flag students

jwright:
Allows staff members to flag various students. A staff member can flag a student from the leaderboard, student profile, and the student index views. A new menu item was added so that the staff member can view a list of the flagged students.

flagged-users

screen shot 2015-09-10 at 4 31 53 pm

screen shot 2015-09-10 at 4 32 34 pm

* Only allows staff members to flag students
* Only can assign a flagged to student to a course if they belong to a course
* NOTE: I am using the unobtrusive js method for flagging students. I don't believe this has been used before in the app (responding with view.js.haml) so let me know if you don't want me to use this and I will change it
* NOTE: There could be a possible "loop hole" where the student is flagged to a specific course but then they are removed from this course. They would still be contained within this courses' flagged users. I can create an `before_destroy` callback on a `CourseMembership` to remove this flagged student but I did not know how often that happens. (I try to avoid callbacks as much as possible).

### DIFF
--- a/app/assets/stylesheets/_buttons_and_icons.sass
+++ b/app/assets/stylesheets/_buttons_and_icons.sass
@@ -98,6 +98,12 @@ tbody
   text-align: center
   float: right
 
+.flagged-user-icon > .fa-flag
+  color: $color-grey-6
+
+  &.flagged
+    color: $color-orange-1
+
 @media (max-width: 1000px)
   .hide-for-small
     display: none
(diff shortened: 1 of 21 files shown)


chcholman:
A couple of minor UI revisions: 

- Let's switch from the star to the flag icon: http://fortawesome.github.io/Font-Awesome/icon/flag/ 
- Let's move the feature to the far left of the table (the farthest left item on both tables) and remove the 'button' styling (just the icon)
- The 'flagged' state should be orange, and the off state should be a grey (probably somewhere around #999) 

Thank you!

jwright:
I made the changes mentioned above. It looks much better and is more discoverable! Thanks.

screen shot 2015-09-11 at 6 39 18 pm

screen shot 2015-09-11 at 6 40 25 pm

screen shot 2015-09-11 at 6 39 23 pm

screen shot 2015-09-11 at 6 39 05 pm

chcholman:
Thanks so much for this @jwright!

PR #1139, opened 2015-09-18
VT - started Shibboleth integration using SAML. Configuration is set …

prabode:
Cait, Here is the basic configuration for SAML with service provider endpoints. This should give U-M IT working urls they requested.


chcholman:
Thanks @prabode!

PR #1146, opened 2015-09-21
Color palette improvements

zazaian:
Added Sass map collections for all sets of colors. Refactored definitions of .color-blue-5 selectors through sass map iterators. Restyle /color_theme.html with white background and selectors associated with each color. Add helpers for defining sub-groups of colors on the themes page. Rewrite color_theme.html template using helpers top DRY up code.


zazaian:
Need to rebase against master before submitting pull request. Will re-submit after rebasing.

PR #1147, opened 2015-09-21
Presenter introduction for assignments show

jwright:
This PR is an early, proof of concept for the introduction of the presenter pattern within our application. I believe that using a presenter pattern with reduce the number of bugs in the system, allow us to change things more rapidly and with higher confidence, and reduce the maintenance cost.

We have a *lot* of logic in views. This is not good because the cost and complexity required to test views. View tests are expensive because of the long runtime and the high amount of setup. Moving the logic in the views to a `Presenter` will enable us to test more easily and thus reduce the amount of bugs that get introduced due to the regression tests. This will also allow us to remove the current view tests when replaced with presenters which will in turn, speed up the test runs.

Since presenters are easy to test, we can use TDD techniques to write the display logic. This will lead to better, more maintainable since TDD is a benefit for that. We can use test doubles to test our collaborators in the presenter so we will not need active record in most cases nor will we need many rails related items. This makes for extremely fast tests and a fast feedback cycle for red, green, refactor tdd loop.

This will allow us to reduce the amount of code in the controllers. Controllers are just meant to translate a request into some sort of business logic and ensuring that the user has access to that request. That is it. Anything else is not the job of the controller. Controller methods should be between 1 - 4 lines and that is it. We have 100 line controller methods. These huge methods are mostly just getting data ready for the views. We can move this responsibility to the presenter.

Lastly, we are breaking [Sandy Metz's rule](https://robots.thoughtbot.com/sandi-metz-rules-for-developers) of only having one variable declared for views. We create variables all over the place for our views. This is not very maintainable and very error prone. By only creating one variable in the controller (e.g. `presenter`), we can use that one variable in the view and ensure that the collaborators are all of the presenter, which is tested.

* Views that need presenter

- [x] `app/views/grades/_staff_view.html.haml` -> uses `app/views/grades/_group_show` && `app/views/grades/_individual_show`
- [x] `app/views/grades/show.html.haml` -> uses `app/views/grades/_group_show`  && `app/views/grades/_individual_show`
- [x] `app/views/submissions/show.html.haml` -> uses `app/views/submissions/_student_show`


app/presenters/assignment_presenter.rb
```diff
@@ -0,0 +1,21 @@
...
+require_relative "view_context"
+
+class AssignmentPresenter < Presenter::Base
+  include Presenter::ViewContext
+
+  def assignment
+    properties[:assignment]
+  end
+
+  def title
```
jwright: Logic for the title in the view here (https://github.com/UM-USElab/gradecraft-development/pull/1147/files#diff-40ad16c98b5bf678334ab92a012cdc8bL3) was moved into the presenter.

This eliminated the need for the assigning of the title in the controller here (https://github.com/UM-USElab/gradecraft-development/pull/1147/files#diff-001c0f1c1de81d202565b6bed95af0eaL32) and the testing of that in a view spec (https://github.com/UM-USElab/gradecraft-development/pull/1147/files#diff-3479cfb0d55185dc825353f908183a60L38) and the controller spec (https://github.com/UM-USElab/gradecraft-development/pull/1147/files#diff-29d4fa26b86a9198ccd330100c5091aeL49)

app/controllers/assignments_controller.rb
```diff
@@ -74,6 +73,8 @@ def show
       else
         @grades_for_assignment = @assignment.all_grades_for_assignment
       end
+
+      render :show, AssignmentPresenter.build({ assignment: @assignment, view_context: view_context })
```
jwright: Renders with options to make a `presenter` local.

app/views/assignments/show.html.haml
```diff
@@ -1,4 +1,4 @@
 - if current_user_is_staff?
-  = render "view_assignment_management"
+  = render partial: "view_assignment_management", locals: { presenter: presenter }
```
jwright: Pass the `presenter` local to the partial

spec/presenters/assignment_presenter_spec.rb
```diff
@@ -0,0 +1,28 @@
+require "rspec/core"
+require "./app/presenters/assignment_presenter"
+
+describe AssignmentPresenter do
+  let(:assignment) { double(:assignment, name: "Crazy Wizardry", pass_fail?: false, point_total: 5000)}
+  let(:view_context) { double(:view_context) }
```
jwright: The `view_context` can be any test double which makes testing the presenters easy without requiring additional libraries, slowing down the test.

spec/presenters/assignment_presenter_spec.rb
```diff
@@ -0,0 +1,28 @@
+require "rspec/core"
```
jwright: No requiring Rails!

jwright:
Jazzy :notes:

chcholman:
@jwright this is so cool to see come together!

Notes so far: 

1. The table is slightly misaligned, I think because of this line: 
https://github.com/UM-USElab/gradecraft-development/blob/presenter-introduction-for-grades-show/app/views/assignments/individual/_table_body.html.haml#L39

the issue is that you're checking for whether the assignment accepts submissions, but there are a number of buttons within that block that need to display regardless of that state. 

2. The "Student Weightable" columns should be displayed immediately next to the raw score: 
https://github.com/UM-USElab/gradecraft-development/blob/presenter-introduction-for-grades-show/app/views/assignments/individual/_table_body.html.haml#L30

jwright:
@chcholman This has been updated to fix the issues found above.

chcholman:
Next up!

If you run the new sample data, and go to the this assignment: 
http://localhost:5000/assignments/2

You should be able to see data on the analytics tab.

chcholman:
Okay @jwright next note. This line: 

https://github.com/UM-USElab/gradecraft-development/blob/presenter-introduction-for-grades-show/app/views/assignments/_view_assignment_grades.html.haml#L12

Needs to check if a student's grade is NOT available. We show students a rubric preview if they don't yet have a grade. If they do have a grade, a filled out rubric appears on their results page.

chcholman:
Believe it or not, ready to go @jwright!!

PR #1148, opened 2015-09-21
Color palette improvements

zazaian:
Creates a slightly better UI for anyone interested in reviewing color swatches through the /color_theme.html page. Condenses bulk color-variable definitions in to simple maps grouped by color, and adds helpers for defining groups etc on the color_theme.html page. I think the new color map definitions signal a better way to group and define colors for existing palettes, regardless of how they're ultimately used or organized.


chcholman:
@zazaian @josankapo refactored slightly to use the page layout and fonts already in place and make more color blocks appear in a single view. There's a lot of value in the way Jonathan had it with the transparent bkg for comparing, so I'm going back and forth on that and the size of the blocks themselves.

zazaian:
I can strip out the white background for now. Ideally it'd just be a button to change the color of the background but probably not worth spending more time on ATM.

chcholman:
Cool. I'd like @josankapo to have a chance to check it out tomorrow before we do anything else.

zazaian:
Sounds good.

josankapo:
I Added a quick comment explaining that the ColorPaletteHelper is not used in GC but only on the color dev page, and added back in the placeholders so that relative values were aligned.  This should be good enough for now.  Probably would be best to actually create an entire Style Guideline page for the app which includes the palette if we find it's worth keeping this page in the long term.

PR #1151, opened 2015-09-22
Fixes validation / input for grade scheme high range

mlangens:


chcholman:
@mlangens This is looking great! Currently, the submit button doesn't redirect to the display page. Can you update that?

PR #1152, opened 2015-09-23
VT - refactored to load saml cert files using files in the system to …

prabode:
…fix production server's env setting issues.


prabode:
@chcholman I am merging this to fix saml certificates issues on the production servers.

PR #1153, opened 2015-09-23
removed additional caching in predictor

josankapo:
I found another instance of caching attached to the predictor and removed it.


chcholman:
I get a failing spec when I test this - will you look into this @josankapo?

josankapo:
My specs passed before I made the pull request — which one fails for you?

chcholman:
n/m my fault!

PR #1155, opened 2015-09-25
VT - Made saml cert files optional to run the application locally.

prabode:
Technical debt, I should have made these optional in the first place. But now saml will throw errors if you try to access the saml urls without setting the cert files.


chcholman:
Awesome, thanks so much @prabode

PR #1170, opened 2015-10-01
Sorts assignments ascending alphanumerically with custom sort function

mlangens:


chcholman:
Hi Max, thanks so much for working on this!

1) When you submit the PR, can you document where it is that I should be testing to see the effects?
2) I added some sample data that can help us build around this a bit. Will you test and confirm that this work successfully allows both sort and inverse sort?

Thank you!

mlangens:
@chcholman I updated the function to sort in reverse. Also, the new sample data looks great. Let me know if there is anything else I can do here!

chcholman:
Thanks @mlangens!

PR #1194, opened 2015-10-07
Testing login paths for all roles

chcholman:
This PR adds feature tests confirming that all user roles are successfully routed to the correct landing page path under the condition that the timeline is turned off.

### DIFF
--- a/spec/features/logging_in_spec.rb
+++ b/spec/features/logging_in_spec.rb
@@ -1,6 +1,6 @@
 require "spec_helper"
 
-feature "logging in" do
+feature "logging in", focus: true do
   let(:password) { "p@ssword" }
 
   context "as a student" do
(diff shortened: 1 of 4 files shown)


jwright:
Hey Cait. This looks great but I feel like testing all of the roles may be overkill. We are not getting much out of testing login for all the roles except that the user is directed to a different url based on their role. These tests would be better served as controller tests. Integration (feature) tests should just serve as "smoke tests". Do all the pieces work together? Does this sound like a reasonable approach?

I think the only 2 login feature tests that are left are the happy path and unhappy path for kerebos login.

spec/features/logging_in_spec.rb
```diff
@@ -1,6 +1,6 @@
 require "spec_helper"
 
-feature "logging in" do
+feature "logging in", focus: true do
```
jwright: We want to make sure that `focus: true` does not get checked in since CI will only run that test and skip the rest. It's great for development though.

chcholman:
@jwright Awesome! Thank you so much for the feedback. I built these out in response to the context 'as a student' - so does that context stay in place, but we ignore the other roles? Do we abstract the role from the context? And then be converted to User Sessions controller tests confirming the routing per role?

So sorry about the focus:true inclusion. I did wonder why it passed so quickly :P

jwright:
@chcholman I think the context would stay in place so that the test reads what it is doing. Since we need to create a user of a certain role, then the test output is just telling us what we are doing. Reading the output would read well:

```
logging in
  as a student
    with a password successfully
```

For the `UserSessionsController` specs, you will probably want to add context there as well so the test output reads something like:

```
  UserSessionsController
    #create
      as a student
        redirects to the syllabus page
      as an admin
        redirects to the analytics page
```

chcholman:
@jwright Take a look when you get a chance! I:

* Removed the role-specific login testing 
* Added a logout feature test

Next step will be adding controller tests for the various roles/routes. Right direction?

spec/features/logging_out_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+    end
+
+    scenario "successfully" do
+
+      within("#account-info") do
+        click_link "Logout"
+      end
+
+      expect(current_path).to eq root_path
+      expect(page).to have_content "You are now logged out."
```
jwright: Can we use `have_notification_message` here? Check out https://github.com/UM-USElab/gradecraft-development/blob/role-authentication/spec/support/has_notification_message_matcher.rb

jwright:
@chcholman This looks great. I just had one comment/question?

chcholman:
Fun! Updated per your request @jwright! Confirm that it's good to go?

PR #1197, opened 2015-10-08
Make specs awesome again

jwright:
- [X] Create base spec helper
- [x] Use base spec helper for testing lib
- [x] Create `active_record_spec_helper`
- [x] Speed up `spec/models`
- [x] Speed up `spec/mailers`
- [x] Speed up `spec/validators`
- [x] Speed up `spec/helpers`
- [x] Speed up `spec/controllers`
- [x] Rename current spec helper to rails_spec_helper

1. What is the purpose of this spec?
1. Is this testing what it is supposed to?
1. Is this spec worth it?
1. Was this spec left over from TDD?
1. Can this spec be better served as something else?
1. Is this subject under test supposed to be responsible for this?

BEFORE 🙏:  1477 examples, 5 min. 57 sec.
AFTER 💇: 1281 examples 3 min. 34 sec.

Here is an example of a sped up test:

model_spec


jwright:
@chcholman Holy epic rebase Batman™. This is finally ready for review.

chcholman:
@jwright Can you get it to pass the Codeship tests?

PR #1208, opened 2015-10-23
Course switch test

chcholman:
This feature tests to confirm that users can select a course for which they have a membership from the dropdown menu in the top nav, click on it, and switched into that context. I considered adding a test to confirm they couldn't access a course they don't have a membership for, but decided that made more sense to do as a controller test (there's nothing for them to click on in the interface - you'd have to do it at the session level). 

@jwright will you review?


spec/features/switching_course_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+
+    before { visit root_path }
+
+    before(:each) do
+      LoginPage.new(user).submit({ password: password })
+    end
+
+    scenario "successfully" do
+
+      within("#course-list") do
```
jwright: This is very good :+1:

spec/features/switching_course_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+  let(:password) { "p@ssword" }
+
+  context "as a student" do
+    let(:course_1) { create :course }
+    let!(:course_membership) { create :student_course_membership, user: user, course: course_1 }
+    let(:course_2) { create :course }
+    let!(:second_course_membership) { create :student_course_membership, user: user, course: course_2 }
+    let(:user) { create :user, password: password }
+
+    before { visit root_path }
```
jwright: This `before` can be merged with the `before(:each)` below. These 2 methods run at the same time.

spec/features/switching_course_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+
+    before(:each) do
+      LoginPage.new(user).submit({ password: password })
```
jwright: Can we use a `Sorcery` helper here instead of manually logging in each time?

jwright: ICYMI: I added this support helper in the PR #1320

PR #1334, opened 2015-11-06
1230 um user import

### LINKED ISSUE #1230: UM user import
We need to be able to quickly import the roster of courses that are at UM. They should 1) know that these users are to be authenticated via Shibboleth, and so not create/trigger password creation process, and 2) mark as active immediately.

jwright:
Allows a staff member to import UM students. These students login via the UM login method and therefore they should not have a password created for them. They also do not need to have both an email and a username. These are generated for them based on what is not provided.

An option to send a welcome email to these users was created and a new email was created for them.

screen shot 2015-11-06 at 12 23 49 pm

import

Closes #1230


zazaian:
Sexy.

chcholman:
@jwright can I confirm a piece of the interaction - if I mark the import as UM, the checkbox "Send Welcome Email To New User Accounts" is disabled. But we created a new email for them - is that going to automatically send? I'd want them to have the choice of sending or not, I think.

jwright:
@chcholman So you want the ability to send welcome emails to user's even if they are not marked as UM students?

PR #1344, opened 2015-11-10
1277 Fix tier badges icon bug

### LINKED ISSUE #1277: Default badge icon route broken in production

jwright:
Replaced `img(src="{{}}")` with ng-src for images that are data-bound to urls.

There were some 404 errors in the error log for requests to something like:

```
/assignments/1807/grade/%7B%7Bbadge.icon%7D%7D
```

This was because something else was failing on the page and the view was trying to get the request for `/assignments/1807/grade/{{badge.icon}}` instead of the databound value.

Using ng-src [should solve this issue](http://stackoverflow.com/a/27554935) as it does not set the `src` attribute if it's not a valid url and the browser will not make the request for it (the attribute is not rendered).

Fixes #1277

### DIFF
--- a/app/views/grades/_rubric_edit.html.haml
+++ b/app/views/grades/_rubric_edit.html.haml
@@ -60,7 +60,7 @@
               {{tier.description}}
             .row.badge-row
               .tier-badge-image(ng-repeat="(tierBadgeId, tierBadge) in tier.badges")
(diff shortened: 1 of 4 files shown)


jwright:
@acousticrobot I wanted to tag you in this PR and see if you agreed with this solution for Angular.

PR #1361, opened 2015-11-13
1349 refactor assignment grade import

### LINKED ISSUE #1349: Refactor Assignment#grade_import and add spec
The grade_import method on the Assignment model could use some light refactoring as well as a spec to test output. It's currently being used in the assignment exports / smart archiver code and would add integrity to that functionality if this were refactored.

jwright:
Refactored both the `AssignmentsController#import_grades` and `AssignmentsController#export_grades` to use exporter objects.

Created `GradeExporter` and `AssignmentExporter` with specs and removed the corresponding methods from the `Assignment` object.

Closes #1349

### DIFF
--- a/app/controllers/assignments_controller.rb
+++ b/app/controllers/assignments_controller.rb
@@ -204,9 +204,9 @@ class AssignmentsController < ApplicationController
   end
 
   def export_grades
-    @assignment = current_course.assignments.find(params[:id])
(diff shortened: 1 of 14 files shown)


jwright:
@zazaian Tagging you in case you want to review this PR as well

jwright:
@chcholman I think that is great. Thanks. Changing now.

PR #1367, opened 2015-11-14
1341 sorting scores not working

### LINKED ISSUE #1341: Sorting scores not working
Something is broken in our dynatable sorting mechanisms - sorting by score disables the JS for the table

jwright:
Modified the dynatable reader for scores so that it sorts on a `numericScore` value and updated the column to sort on that new `numericScore` column. The issue was that sometimes the score was a string and sometimes the score was a number. When dynatable tried to treat it as a string, `toLowerCase` would not work on a number.

The writers were removed since the readers for score now return the score as it was written but they sort on the `numericScore` column. All of the score columns in all dynatables were updated.

Fixes #1341

### DIFF
--- a/app/assets/javascripts/dynatable.js
+++ b/app/assets/javascripts/dynatable.js
@@ -22,11 +22,8 @@ $('table.nopage_dynatable').dynatable({
       return Number(el.innerHTML) || 0;
     },
     score: function(el, record) {
-      if($.trim(el.innerHTML) == '') {
-        return el.innerHTML;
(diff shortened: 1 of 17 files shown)


chcholman:
@jwright Yes, we need to handle displaying a 0 if that's the score, and a blank if there's not.

jwright:
@chcholman I removed the old way with the writer and now it sorts on a new column in the record. I manually tested all of the tables.

Issue #1366, opened 2015-11-14
Refactor & test Assignment model


chcholman:

jwright:
Current grade from CodeClimate is an [`F`](https://codeclimate.com/github/UM-USElab/gradecraft-development/Assignment)

Current code coverage is at [`65.8%`](https://codeclimate.com/github/UM-USElab/gradecraft-development/coverage/564792a41787d77c48031115)

PR #1386, opened 2015-11-17
1366 refactor and test assignment

jwright:
Refactors the `Assignment#open?` method to reduce complexity and increase readability for work on #1366.

Test cases were written for all the use cases that were being checked for within the `Assignment#open?` method and then, one by one, methods were extracted. The `opened?`, `overdue?`, and `accepting_submissions?` methods were extracted and tested to help increase the readability.

### DIFF
--- a/app/models/assignment.rb
+++ b/app/models/assignment.rb
@@ -399,24 +399,21 @@ class Assignment < ActiveRecord::Base
     end
   end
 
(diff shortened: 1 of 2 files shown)


jwright:
screen shot 2015-11-17 at 11 33 00 am

PR #1422, opened 2015-11-25
1391 new um user

### LINKED ISSUE #1391: Add UM user in interface

jwright:
Allows a staff member to create a new user account for a UM student.

Introduces a new tabbed UI so the staff member can choose if they are adding an internal (UM) user or a GradeCraft user. The form adjust itself to display the appropriate form fields.

new-user

Created a new service that removes the complicated branching logic from the controller and moves it into a service that runs a bunch of small actions that execute a bit of functionality. This was enabled by using [LightService](https://github.com/adomokos/light-service).

The complicated logic for the user import was also replaced with the new service. This removed a lot of duplication.

Closes #1391


chcholman:
@jwright I think I might be missing something - I've installed the gem, restarted the server, and am looking at the user/_new form, but I don't see the tabbed UI. Is it possible you haven't committed that part of the work?

jwright:
@acousticrobot @zazaian @chcholman I wanted to bring your attention to this PR and point out the services architecture from using [LightService](https://github.com/adomokos/light-service).

The service is basically a container that runs a bunch of actions. You can see the service here: https://github.com/UM-USElab/gradecraft-development/pull/1422/files#diff-f27408b80b411149b9ac1778ee51f300

The actions are created as classes under the services directory. You can see them here: https://github.com/UM-USElab/gradecraft-development/tree/1391-new-um-user/app/services/creates_new_user

You can see that the actions are really small, concrete actions. They are not large and should not be more than 5 lines IMO or the action is doing too much.

For testing the services, you should just basically test that the service runs the appropriate actions. The test for the service is here: https://github.com/UM-USElab/gradecraft-development/blob/1391-new-um-user/spec/services/creates_new_user_spec.rb

The actions are really easy to test since they are small concrete actions: You can see them here: https://github.com/UM-USElab/gradecraft-development/tree/1391-new-um-user/spec/services/creates_new_user_spec

LightService is cool because it has a concept of a `context` which can add things to it while each action finishes. The actions can expect or promise certain objects which the tests can easily test for.

NOTE: The service and the actions should be named as verbs as they execute functionality rather than represent objects.

If you have any thoughts, I'd love to hear them!

jwright:
@chcholman I have not committed that yet. I am working on it now.

chcholman:
@jwright This looks awesome. I've got one new element that needs to be added to the external user creation process, but am going to break that out into a new ticket. Okay for me to merge this, and leave the branch open for conversation with @acousticrobot and @zazaian?

jwright:
@chcholman That sounds good. I am ok with merging this in if you want to. I can change it later if we decide a different direction

jwright:
@chcholman This is ready for review

PR #1446, opened 2015-11-29
Add EventLogger subclass for Login Events, rescue with #perform on redis error

### LINKED ISSUE #1443: Reactivate last login record

chcholman:


app/controllers/application_controller.rb
```diff
@@ -154,12 +154,11 @@ def pageview_logger_attrs
     }
   end
 
-  # NOTE: does this need to be re-activated?
   # Tracking course logins
   def log_course_login_event
-    # membership = current_user.course_memberships.where(course_id: current_course.id).first
-    # Resque.enqueue(EventLogger,'login', course_id: current_course.id, user_id: current_user.id, student_id: current_student.try(:id), user_role: current_user.role(current_course), last_login_at: membership.last_login_at.to_i, created_at: Time.now)
-    # membership.update_attribute(:last_login_at, Time.now)
+    membership = current_user.course_memberships.where(course_id: current_course.id).first
```
zazaian: I believe this will work fine as-is, but needs to be upgraded with an EventLogger class to match the other event loggers. I think it would be fine to open this up now and create a separate issue for building a new LoginEventLogger class.

chcholman:
@zazaian Yes, that was what we'd agreed you would do on this branch.

PR #1453, opened 2015-12-02
Private methods at bottom

chcholman:


zazaian:
This is one of the funniest commit messages of all time.

PR #1457, opened 2015-12-02
1366 refactor and test assignment

jwright:
Moves some logic out of `Assignment` and into a `Gradable` module.

I am unsure if we want these changes. This will reduce the complexity of `Assignment`, moving it from a `F` -> `D` but it does hide away some of the functionality of `Assignment` into a module.

I am ok with closing this PR without merging but I wanted to get feedback from the people™.


jwright:
@chcholman @acousticrobot I cleaned up the module a bit. I removed the creation of hashes from the model and moved to the presenter. I also removed some unused methods :grinning: and changed the name of the `all_grades_for_assignment` method.

Ready for review again.

chcholman:
Looks good to me - @acousticrobot?

PR #1464, opened 2015-12-05
1276 assignments controller

### LINKED ISSUE #1276: Test coverage for Assignments Controller

jwright:
This improves the test coverage and increases the GPA grade for the `AssignmentsController`.

Reduced the line count for all methods (except `export_submissions`) to be below 8 lines which should improve complexity.

`export_submissions` was not included as it is fixed in the [`smart_archiver`](https://github.com/UM-USElab/gradecraft-development/tree/smart_archiver) branch.

Closes #1276

### DIFF
--- a/app/controllers/assignment_types_controller.rb
+++ b/app/controllers/assignment_types_controller.rb
@@ -1,4 +1,5 @@
 class AssignmentTypesController < ApplicationController
+  include SortsPosition
 
   before_filter :ensure_staff?, :except => [:predictor_data]
   before_action :find_assignment_type, only: [:show, :edit, :update, :export_scores, :all_grades, :destroy]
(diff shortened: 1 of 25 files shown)


jwright:
@chcholman This seems like a very weird method as it updates the `id` column for assignments (I didn't know you could do that :/). I can't find anywhere where it is being used. Do you know if it is being used off-hand?

https://github.com/UM-USElab/gradecraft-development/blob/master/app/controllers/assignments_controller.rb#L123

chcholman:
@jwright Ah. No, that's not the goal... It's definitely being used - it is supposed to update the _position_ of the assignment and can be manipulated by loading the assignment index, and drag/drop reordering the assignment in the list. You'll find the same 'sort' method on badges, and assignment types.

jwright:
@chcholman Ok. I will update it and write a test for it :)

jwright:
@chcholman Also, I read it wrong. It is updating the `position` attribute.

chcholman:
@jwright (phew)

PR #1513, opened 2015-12-15
hotfix: check permissions for grade predicted score

acousticrobot:
* Fixes a but where professors were able to see student grades
  * Passes current_user into PredictedAssignmentCollection, PredictedAssignment, PredictedGrade
  * Handles permissions within these models
  * Simplifies logic: predicted score and permission to update are only granted to current student

### DIFF
--- a/app/models/predicted_assignment_collection.rb
+++ b/app/models/predicted_assignment_collection.rb
@@ -1,19 +1,24 @@
 class PredictedAssignmentCollection
   include Enumerable
 
-  attr_reader :assignments, :user
+  attr_reader :assignments, :current_user, :student
 
-  def initialize(assignments, user)
+  def initialize(assignments, current_user, student)
     @assignments = pluck_attributes assignments
-    @user = user
+    @current_user = current_user
+    @student = student
   end
 
   def each
-    assignments.each { |assignment| yield PredictedAssignment.new(assignment, user) }
+    assignments.each { |assignment| yield PredictedAssignment.new(assignment, current_user, student) }
   end
 
   def [](index)
     to_a[index]
+  end
+
+  def permission_to_update?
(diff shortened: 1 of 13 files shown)


app/models/predicted_assignment_collection.rb
```diff
@@ -1,21 +1,26 @@
...
-    assignments.each { |assignment| yield PredictedAssignment.new(assignment, user) }
+    assignments.each { |assignment| yield PredictedAssignment.new(assignment, current_user, student) }
   end
 
   def [](index)
     to_a[index]
   end
 
+  def permission_to_update?
+    @current_user == @student
```
jwright: @acousticrobot I think it is a good practice to use accessor (especially readers) instead of instance variables when you can. This reduces unintended consequences of modifying instance variables and this corresponds with ruby by sending messages vs. accessing data.

app/models/predicted_grade.rb
```diff
@@ -8,7 +10,7 @@ def pass_fail_status
   end
 
   def predicted_score
-    grade.predicted_score if grade.student && grade.student.is_student?(grade.course)
+    grade.student == @current_user ? grade.predicted_score : 0
```
jwright: Same comment as above for this section

app/views/assignments/predictor_data.json.jbuilder
```diff
@@ -48,6 +48,6 @@ end
 json.term_for_assignment term_for :assignment
 json.term_for_pass current_course.pass_term
 json.term_for_fail current_course.fail_term
-json.update_assignments @assignments.user == current_user
+json.update_assignments @assignments.permission_to_update?
```
jwright: :+1:

PR #1545, opened 2015-12-26
1442 move exporters off course model

chcholman:
This PR moves the methods that had been living on the Course model for file downloads over to distinct exporters.


app/exporters/assignment_exporter.rb
```diff
@@ -1,5 +1,5 @@
 class AssignmentExporter
```
jwright: I feel like we should rename this to `GradeExporter` with an `export` method instead of renaming the method

app/controllers/courses_controller.rb
```diff
@@ -119,7 +119,7 @@ def destroy
   def export_earned_badges
     @course = current_course
```
jwright: If we do not have  a view or something that needs an instance variable, we should not create the instance variable. This can help us look at a glance and ensure that a variable is not exposed that is not needed. I would change this to `course` instead of `@course`.

app/exporters/course_grade_exporter.rb
```diff
@@ -1,19 +1,31 @@
 class CourseGradeExporter
```
jwright: This name is confusing for me. What is a `CourseGrade`? We don't seem to have that construct but we have this as an exporter. I am not saying that the name does not work but it was confusing at first glace.

app/exporters/course_grade_exporter.rb
```diff
@@ -1,19 +1,31 @@
...
 
   #final grades: total score + grade earned in course
-  def final_grades_for_course(course)
+  def final_grades_for_course(course, students)
```
jwright: Why not just use the `course` argument and get the students from the course?

app/controllers/info_controller.rb
```diff
@@ -134,6 +135,12 @@ def choices
     @teams = current_course.teams
   end
 
+  def assignment_structure
```
jwright: This seems to be more appropriate on an `AssignmentStructureController` with an `export` method. Actions are usually verbs but we are treating them as nouns in most of the actions on the `InfoController`. I fear that `InfoController` will soon just be a dumping ground for actions that do not belong anywhere else. I think it is ok to have small controllers with a single action.

app/exporters/assignment_structure_exporter.rb
```diff
@@ -0,0 +1,16 @@
+class AssignmentStructureExporter
+  def assignment_structure(course)
```
jwright: I feel we should rename this to `export` to explain what the method does.

app/exporters/gradebook_exporter.rb
```diff
@@ -0,0 +1,55 @@
...
+      memo << student.send(method)
+    end
+    # todo: we need to pre-fetch the course teams for this
+    student_data << student.team_for_course(course).try(:name)
+
+    # add the grades for the necessary assignments, todo: improve the performance here
+    course.assignments.inject(student_data) do |memo, assignment|
+      grade = assignment.grade_for_student(student)
+      if grade and grade.is_student_visible?
+        memo << grade.try(:raw_score)
```
jwright: This `try` is unnecessary as there is a check for the grade on line 47.

app/exporters/multiplied_gradebook_exporter.rb
```diff
@@ -0,0 +1,36 @@
...
+    end
+    # todo: we need to pre-fetch the course teams for this
+    student_data << student.team_for_course(course).try(:name)
+
+    # add the grades for the necessary assignments, todo: improve the performance here
+    course.assignments.inject(student_data) do |memo, assignment|
+      grade = assignment.grade_for_student(student)
+      if grade and grade.is_student_visible?
+        memo << grade.try(:raw_score)
+        memo << grade.try(:score)
```
jwright: This `try` and the `try` on line 27 are not necessary because there is a check for the grade on line 26.

app/models/course.rb
```diff
@@ -349,142 +349,6 @@ def ordered_student_ids
       .collect(&:id)
   end
 
-  #Export Users and Final Scores for Course
```
jwright: :+1:

app/performers/grade_export_performer.rb
```diff
@@ -27,8 +27,8 @@ def fetch_course
     Course.find @attrs[:course_id]
   end
 
-  def fetch_csv_data
-    @csv_data = @course.research_grades_csv
+  def fetch_csv_data(course)
+    @csv_data = GradesForResearchExporter.new.research_grades course.id
```
jwright: Since the method calls for an id but then fetches the course within the method, can we just pass the course into the method here to eliminate the need for another call?

spec/exporters/assignment_type_exporter_spec.rb
```diff
@@ -6,25 +6,62 @@
   let(:students) { create_list :user, 2 }
   subject { AssignmentTypeExporter.new }
 
+  before(:each) do
+    @assignment_type_1 = create(:assignment_type, course: course, name: "Charms")
```
jwright: Do the `assignment_type` need to be instance variables? Do we use them below in the tests? If not, I would just make these `let!`.

spec/exporters/assignment_type_exporter_spec.rb
```diff
@@ -6,25 +6,62 @@
...
+    @assignment_type_2 = create(:assignment_type, course: course, name: "History of Wizardry")
+    @assignment_types = course.assignment_types
+    @assignment = create(:assignment, course: course, assignment_type: @assignment_type_1)
```
jwright: Can we change these to `let!`? Let's help define the setup where a `before` block is usually used for the act phase of testing.

spec/models/course_spec.rb
```diff
@@ -701,35 +701,7 @@
       expect(subject.ordered_student_ids).to eq([student_2.id, student_1.id, student_3.id])
     end
   end
-
```
jwright: :+1:

spec/performers/grade_export_performer_spec.rb
```diff
@@ -92,94 +92,4 @@
       end
     end
   end
-
```
jwright: :+1:

spec/performers/gradebook_export_performer_spec.rb
```diff
@@ -6,6 +6,7 @@
   let(:user) { create(:user) }
   let(:attrs) {{ user_id: user[:id], course_id: course[:id] }}
   let(:performer) { GradebookExportPerformer.new(attrs) }
+  let(:exporter) { GradebookExporter.new.gradebook(course[:id]) }
   subject { performer }
 
   describe "public methods" do
```
jwright: Can we remove this describe as we should not be testing anything other than "public methods"?

app/exporters/multiplied_gradebook_exporter.rb
chcholman: There are instances where these grades exist and may have feedback or a predicted grade, but the raw score and score are actually nil.

app/exporters/course_grade_exporter.rb
chcholman: @jwright Would it be better to call it FinalGrade? We don't have a distinct construct for it - it's the student's total score for the course, and the final grade or level that they achieve at the course level.

PR #1556, opened 2015-12-31
1530 submission history

### LINKED ISSUE #1530: Submission History
As a student or a staff member
I should be able to view the history of submission changes when the submission is updated
So that I can view my progression

jwright:
This adds the ability to track and report changes that occur to a `Submission`. All changes for a submission are tracked using [PaperTrail](https://github.com/airblade/paper_trail). 

### Database changes

PaperTrail stores changes for the models in a `versions` table which was created as part of this PR.

### Changesets

The PaperTrail [changesets](https://github.com/airblade/paper_trail#diffing-versions) are then modified a bit on the object as to add a bit of metadata that is needed for the display of the changes. This property (`#history`) was added to a model concern and it can be attached to any object that needs this behavior.

### Displaying human readable changesets

The history of submissions are displayed for the staff via the `Submission#show` route (e.g. `/assignments/:assignment_id/submissions/:id`) and the history of the submissions are displayed for the student via the `Assignments#show` route (e.g. `/assignments/:assignment_id`). The history is displayed in a tabbed interface. Here is an example for the student:
 
screen shot 2015-12-31 at 11 07 34 am

The way that the application takes changesets and turns them into a "human readable form" is through the use of a small library I created called `HumanHistory`. It uses a Tokenizer to create tokens for a sentence.
It converts the following changeset (example of a created submission):

```
{
  "id" => [nil, 145],
  "created_at" => [nil, "2015-12-31 14:35:45"],
  "event" => "create",
  "object" => "Submission",
  "actor_id" => 12
}
```

Into the following sentence:

`Jimmy Page created the submission on December 31st, 2015 at 2:35 PM`

It also handles changes:

```
{
  "updated_at" => ["2015-12-31 14:35:45", "2015-12-31 16:22:21"],
  "link" => [nil, "http://example.org"],
  "event" => "update",
  "object" => "Submission",
  "actor_id" => 12
}
```

Into:

`You updated the link from "" to "http://example.org" on December 31st, 2015 at 4:22 PM`

### A few questions

* The display of the changes do not have any styles applied to them. They are simple divs and it can easily be changed if we wish.

* It does not currently handle large strings. So if a user changes the `text_comment`, the content will be displayed in it's entirety. I can change this to only display the first few words with a link to hover over and see more but I wanted to get this PR out first and have you let me know if it needs to change.

* The date/time are displayed in UTC time. I assume we want them changed to the current user's timezone but let me know if that is not desired.

Closes #1530


spec/models/submission_file_spec.rb
```diff
@@ -18,20 +18,29 @@
...
+  describe "#assignment" do
+    it "returns the assignment associated with the submission" do
       expect(subject.assignment).to eq(assignment)
     end
   end
 
-  describe "#student" do 
-    it "returns the course associated with the submission" do 
+  describe "#student" do
+    it "returns the course associated with the submission" do
```
chcholman: I think that description should probably be "returns the student associated with the submission"?

chcholman:
I've run bundle install and the database migrations. When I run foreman start, I get the following error in the log before the server shuts down: 
ActiveSupport::Concern::MultipleIncludedBlocks: Cannot define multiple 'included' blocks for a Concern

jwright:
@chcholman I had issues with `require_dependency` when running a single test file that required those classes. I could not run the file without including all of ActiveSupport. I will look for work arounds.

jwright:
@chcholman Welp, it helps when things are spelled right. Rails was not autoloading the `multiple_file_attributes` because the module name was `MultipleFileAttibutes` and Rails will not auto-load files that do not match module names :hankey:

chcholman:
@jwright I'm still unable to start the server after pulling down these changes. Same error as before: 

ActiveSupport::Concern::MultipleIncludedBlocks: Cannot define multiple 'included' blocks for a Concern

I've deleted the branch and fetched the updates from origin just to be sure, but it persists.

jwright:
@chcholman Ok. Sorry about that. Let me see what sort of Ruby hacking voodoo that Rails is doing now.

jwright:
@chcholman Ok. I fixed the issue. I added the `require_relative` back for the `Historical` concern back on `Submission` without testing it. I apologize.

I added it back because when the `submissions#show` was rendered in the browser, I was getting a `history not found on submission` error (which is added from the `Historical` concern). The issue was that this was a naming conflict with the `Historical` library I created. I fixed this by renaming the `Historical` library to `HumanHistory` 😎

It's ready for review again.

PR #1587, opened 2016-01-24
1313 external submission links

### LINKED ISSUE #1313: Links in assignment submissions or descriptions that go outside of the gradecraft domain should open in a new tab

jwright:
Redirects all links for `Submission` links and links within the text comments.

For links, this PR contains a new helper method called `external_link_to` which adds a `target=_blank` attribute to the link if the link is to an external source. 

In order to determine whether a link is external, another helper method called `external?` determines if the link is external by parsing it via the [`uri`](http://ruby-doc.org/stdlib-1.9.3/libdoc/uri/rdoc/URI.html) module and checking the `host`.

In order to sanitize the links within the text_comment, another helper method was created called `sanitize_external_links` which uses the [`Rails::Html::Sanitizer`](https://github.com/rails/rails-html-sanitizer) with a custom scrubber to add the `target=_blank` attribute for external links. The scrubber was modeled after the [`NoFollow`](https://github.com/flavorjones/loofah/blob/master/lib/loofah/scrubbers.rb#L188) scrubber which adds a `nofollow` attribute.

NOTE: I could not actually test the sanitizing of the `text_comment` as submissions get stripped of any links when they are saved via this: https://github.com/UM-USElab/gradecraft-development/blob/master/app/models/submission.rb#L18

Closes #1313


app/helpers/link_helper.rb
```diff
@@ -0,0 +1,68 @@
...
+
+  def external_link?(uri)
+    UriInspector.new(uri).external?
+  end
+
+  def sanitize_external_links(content)
+    sanitize content, scrubber: InternalLinkScrubber.new
+  end
+
+  def omission_link_to(name = nil, options = nil, html_options = nil, &block)
```
zazaian: Can we break this out into multiple sub-concerns in order to reduce method complexity here?

jwright: @zazaian I took a look at this and I am having a hard time breaking this out. Do you have any ideas on how to do this?

The big problem is that Rails' API for `link_to` is absolutely horrid. Each parameter is different based on if a block was passed or not.

zazaian: @jwright just because there's so much going on here I'm wondering if this warrants splitting out into a separate OmissionLinkBuilder class. Feels like a lot of logic to stuff into six consecutive local variables. That said I made a quick attempt at it and it didn't feel a lot better as a separate class:

```
class OmissionLinkBuilder
  include ActionView::Helpers::UrlHelper

  def initialize(name=nil, options=nil, html_options=nil, &block)
    @name = name
    @options = options
    @html_options = html_options
    @block = &block
    @block_given = block_given?
  end
  attr_reader :name, :options, :html_options, :block, :block_given

  def build_options
    omission_options = (block_given ? options : html_options) || {}
    @limit = omission_options.delete(:limit) { 50 }
    @indicator = omission_options.delete(:indicator) { "..." }
  end

  def build_content
    content = block_given ? capture(block) : name
    @original_content = content.html_safe
    @final_content = "#{content[0..(limit - indicator.length)]}#{indicator}" if content.length > limit
  end

  def build_block_options
    if block_given
      @block = lambda { @final_content }
      @options = { "title" => @original_content }.merge(options || {})
    else
      @name = content
      @html_options = { "title" => @original_content }.merge(html_options || {})
    end
  end

  def render
    build_options; build_content; build_block_options
    link_to @name, @options, @html_options, @block
  end
end
```

jwright:
@zazaian @chcholman I agree that this does not clear this up much. It also adds the question about where do these files go? They support helpers which cannot be placed in the `app/helpers` directory.

I think this problem stems from readability of the existing `link_to` helpers, which Mike points out as having a lot of functionality in 5 variables. The fact that this has 5 parameters is an issue.

I vote for committing this but I am open to extracting it out if someone thinks it valuable.

zazaian:
@jwright I think it's fine to put this in for now, but let's be mindful of other areas in which we're running into similarly complex link-building scenarios. Maybe the patterns will give us some clues on how to properly extract them.

jwright:
:+1:

Issue #1533, opened 2015-12-24
Grading Submission Revisions


jwright:
As a professor
When I assign a grade to a submission
Then that grade should be for that specific submission revision
So that I can grade other revisions later

jwright:
Related to #875

jwright:
@chcholman How do we want to show this in the UI? Should we just show it on the submission history? Should we implement a UI like we talked about in Slack (e.g. Basecamp history)?

chcholman:
@jwright The tree history is quite cool, but it can even be a little bit simpler: 

Assignment 1
Submitted January 20, 2016, 10:15PM
* file attachment.pdf
Score: 80
Feedback:  [ text feedback ]
Rubric Grade: [grid]
--
Resubmitted Feb 1, 2016, 8:42AM
* revised file-attachment.pdf
Score: 95
Feedback: [feedback]
Rubric grade if present

I am wondering about 1) showing it in reverse order (but not convinced) and 2) highlighting difference ("You earned 15 more points than your first submission")

jwright:
@chcholman Is this partial going to replace what is currently in the submission history tab when showing a submission?

chcholman:
@jwright That would be ideal - to show them in a unified manner.

jwright:
- [ ] Only show grade event if raw score or feedback were updated (raw score, state, or feedback)
- [ ] Make sure grade event is not shown if displaying to students and the grade is not visible
- [ ] Change sentence for when the field changes from nil
- [ ] Do not put quotes around integers in sentence
- [ ] Remove date from content
- [ ] Students do not need to see state changes on grades
- [ ] Fix css so that it only affects the timeline

PR #1593, opened 2016-01-27
1533 hiding grade changes

jwright:
Displays only changes that are safe for a student to see within the submission history. A student should only see grade changes when it is safe to view the grade changes. This includes grade changes that have not been released or previous raw score changes.

In order to know if a grade should be displayed, it needs to be `reified` at a specific version. The `HistoryFilter` needs to get both the `changeset` and the `version`.

In order to not display the previous raw scores for grade changes, the concept of a change description formatter was introduced. This formats the change description for different changesets. I added a `RawScoreChangeDescriptionFormatter` that gets used when the attribute is a raw score.

Ref #1533


lib/human_history/raw_score_change_description_formatter.rb
```diff
@@ -0,0 +1,14 @@
+require_relative "default_change_description_formatter"
+
+module HumanHistory
+  class RawScoreChangeDescriptionFormatter < DefaultChangeDescriptionFormatter
+    def formattable?
+      attribute == "raw_score" && type.classify == "Grade"
```
acousticrobot: This is the first time that you have introduced GC specific logic into the lib folder, which up to now was abstracted away from our app-specific logic. I wonder as this develops, if more GC rules would be added to the lip/human_history folder, or if we could abstract the definitions of Object.type rules at a higher level.

jwright: @acousticrobot I thought about this very thing while I was creating the spec. This file does not need to be in this directory, it can be moved into app. Do you think that would be a good change?

jwright:
@acousticrobot Great idea. I can create a wiki page for it.

app/helpers/history_helper.rb
```diff
@@ -83,6 +83,9 @@ def build_sentence(changeset)
 
   def tokenize_sentence(changeset, options={})
     tokenizer = HumanHistory::HistoryTokenizer.new(changeset)
-    HumanHistory::HistoryTokenParser.new(tokenizer).parse({ current_user: current_user }.merge(options))
+    HumanHistory::HistoryTokenParser.new(tokenizer)
+      .parse({ current_user: current_user,
+               change_description_formatters: [HumanHistory::RawScoreChangeDescriptionFormatter, HumanHistory::DefaultChangeDescriptionFormatter] }
```
zazaian: Can these go into a #change_description_formatters method rather than being crammed into the method call here?

app/models/concerns/historical.rb
```diff
@@ -13,11 +23,12 @@ def has_history?
 
   def history
     self.versions.reverse.map do |version|
-      changeset = version.changeset.dup
-      changeset.merge!("object" => self.class.name)
-      changeset.merge!("event" => version.event)
-      changeset.merge!("actor_id" => version.whodunnit)
-      changeset.merge!("recorded_at" => version.created_at)
+      history = HistoryItem.new(version)
+      history.changeset.merge!("object" => self.class.name)
```
zazaian: Can we merge all of these once as a separate hash rather than performing iterative merges for each attribute?

spec/human_history/default_change_description_formatter_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+describe HumanHistory::DefaultChangeDescriptionFormatter do
+  let(:attribute) { "attribute" }
+  let(:changes) { ["previous", "current"] }
+  let(:type) { "Object" }
+
+  subject { described_class.new attribute, changes, type }
+
+  describe "#initialize" do
+    it "is initialized attribute, changes, and type" do
+      expect(subject.attribute).to eq attribute
```
zazaian: Can we try to observe the "one expectation per it block" convention here?

spec/models/history_filter_spec.rb
```diff
@@ -1,74 +1,81 @@
...
+      expect(described_class.new(history).empty_changeset?(history.first.changeset.first)).to eq true
     end
   end
 
   describe "#exclude" do
     it "filters a changeset by event type" do
-      changeset = [{ "event" => "blah", "change" => ["before", "after"] },
-                   { "event" => "bleh", "change" => ["before", "after"] }, {}]
-      result = described_class.new(changeset).exclude("event" => "blah").changeset
+      history = [OpenStruct.new(changeset: { "event" => "blah", "change" => ["before", "after"] }),
```
zazaian: Is there a more elegant way to design some of these variable conventions that would be more expressive to a reader?

spec/toolkits/historical_toolkit.rb
```diff
@@ -33,32 +33,32 @@
       model.save
     end
 
-    it "returns the changesets for the created #{fixture}" do
+    it "returns the history for the created #{fixture}" do
```
zazaian: Again, multiple expectations per example. would be cool to clean this up a bit.

lib/human_history/raw_score_change_description_formatter.rb
acousticrobot: :+1:

jwright: @acousticrobot I tried to pull this out and move this to `app/helpers/` but Rails autoloads all files in this directory and treats it as a module. I spent too much time on it so I feel like we should just leave it as it is.

spec/models/history_filter_spec.rb
jwright: @zazaian I do not know of one. I feel like this is really readable and placing it in a helper method may reduce it's readability without providing a good reason for the abstraction.

lib/human_history/raw_score_change_description_formatter.rb
acousticrobot: I hear ya. It's definitely a 'nice to have' and not a necessary change.

PR #1595, opened 2016-01-27
1581 new course membership

### LINKED ISSUE #1581: Create new course membership if user already exists
As an instructor
When I add student to my class that exists in another class I do not have access to
Then I should be able to create a new course membership for them for my course
And return a success notification

jwright:
Updates the user information and adds them to the current course when a professor tries to create a new user with an email that already exists.

In order to do this, a new service was created to "create or update a user". This service just checks to see if the user already exists and then calls the appropriate service to create or update a user.

The `CreatesNewUser` service stays untouched and a new `UpdatesUser` service updates the user attributes and then creates a `CourseMembership` for the new user and the current course.

Closes #1581

### DIFF
--- a/app/services/creates_or_updates_user/creates_or_updates_user.rb
+++ b/app/services/creates_or_updates_user/creates_or_updates_user.rb
@@ -1,4 +1,5 @@
 require_relative "../creates_new_user"
+require_relative "../updates_user"
 
 module Services
   module Actions
@@ -9,9 +10,20 @@ module Services
 
       executed do |context|
         attributes = context[:attributes]
-        course = attributes[:course]
+        email = attributes[:email]
 
-        Services::CreatesNewUser.create attributes
+        if email.blank?
+          context.fail! "Email can't be blank", 422
+          next context
+        end
+
+        if User.email_exists? email
+          course = attributes[:course]
+          context.add_to_context Services::UpdatesUser.update(attributes, course)
+        else
+          send_welcome_email = attributes[:send_welcome_email]
+          context.add_to_context Services::CreatesNewUser.create(attributes, send_welcome_email)
+        end
       end
     end
   end
--- a/app/services/creates_or_updates_user/creates_or_updates_user.rb
+++ b/app/services/creates_or_updates_user/creates_or_updates_user.rb
@@ -1,7 +1,9 @@
 module Services
   module Actions
-    class CreateOrUpdateUser
+    class CreatesOrUpdatesUser
       extend LightService::Action
+
+      expects :attributes, :course
 
       executed do |context|
       end
--- a/app/services/creates_or_updates_user/creates_or_updates_user.rb
+++ b/app/services/creates_or_updates_user/creates_or_updates_user.rb
@@ -1,11 +1,17 @@
+require_relative "../creates_new_user"
+
 module Services
   module Actions
     class CreatesOrUpdatesUser
       extend LightService::Action
 
-      expects :attributes, :course
+      expects :attributes, :course, :send_welcome_email
 
       executed do |context|
+        attributes = context[:attributes]
+        course = attributes[:course]
+
+        Services::CreatesNewUser.create attributes
       end
     end
   end
--- a/app/services/creates_or_updates_user/creates_or_updates_user.rb
+++ b/app/services/creates_or_updates_user/creates_or_updates_user.rb
@@ -18,7 +18,7 @@ module Services
         end
 
         if User.email_exists? email
-          course = attributes[:course]
+          course = context[:course]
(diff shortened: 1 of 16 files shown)


app/services/updates_user/updates_user.rb
```diff
@@ -0,0 +1,21 @@
...
+
+      executed do |context|
+        attributes = context[:attributes]
+        email = attributes[:email]
+
+        user = User.find_by_insensitive_email email
+        context.fail_with_rollback!("User could not be found", error_code: 404) if user.nil?
+
+        context.fail_with_rollback!("The user is invalid and cannot be saved", error_code: 422) \
+          unless user.update_attributes attributes
```
acousticrobot: Are we covered for all use cases here with both permissions and checks? for instance 
screen shot 2016-01-27 at 3 59 35 pm

app/services/creates_or_updates_user/creates_or_updates_user.rb
```diff
@@ -0,0 +1,30 @@
...
+          next context
+        end
+
+        if User.email_exists? email
+          course = context[:course]
+          context.add_to_context Services::UpdatesUser.update(attributes, course)
+        else
+          send_welcome_email = attributes[:send_welcome_email]
+          context.add_to_context Services::CreatesNewUser.create(attributes, send_welcome_email)
+        end
```
acousticrobot: :+1: I really like how you've directed further service actions here.

app/services/updates_user/updates_user.rb
jwright: @acousticrobot  Hmm. I see what you are saying here.

I think that this abstraction should be within the controller or the something beyond the service. If we put that restriction on the service, then the service will only be geared for that specific controller/action instance and making it less useful.

Since we are not using strong parameters in the app, the model is still the thing that tells what attributes can be updated. We control access to what is updated via controller authentication rules.

What are your thoughts on this? Do you agree that the controller/view should be responsible for what attributes are updated?

spec/features/creating_a_new_user_spec.rb
```diff
@@ -10,6 +10,24 @@
...
 
+    context "for an existing GradeCraft student" do
+      scenario "successfully" do
+        user = create(:user)
+        within(".pageContent #tab1") do
+          NewUserPage.new(user)
+            .submit(courses: [course_membership.course])
+        end
+
+        expect(current_path).to eq students_path
```
zazaian: Can these be broken out in to multiple spec examples rather than a very large one?

app/services/updates_user/updates_user.rb
acousticrobot: I definitely agree the controller should be responsible for managing permissions. My concern is more: 
1) If a new developer down the line uses this from within a student accessible route and doesn't take this into account (I'm thinking to the similar `update_attributes` that I found recently in self logged grades where we weren't checking it the grade passed in was self logged -- I was able as a student to update all my grades for the course using a REST client.) 
2) Since new services will be added and could be used in combinations we haven't foreseen, what if well intentioned `context[:attributes]` meant for another service also matched attributes on the user model.

I also agree it would be ugly to explicitly update from a hash of parameters, and as soon as someone adds, say :password to the list it's no more secure than before. Would it make sense to add `user_attributes: {...}` or something within the broader `attributes` passed to the service chain?

spec/features/creating_a_new_user_spec.rb
jwright: This would cause a performance hit since these expects are in integration specs. We would have to drive the web page 4 times instead of once. I don't think we should do that in these cases.

app/services/updates_user/updates_user.rb
jwright: @acousticrobot I don't believe I am following this. The `attributes` are essentially user attributes. Are you suggesting we rename this to solve an issue?

I understand the concern here but I really believe this is a Rails issue and not the service issue. Rails `attr_accessible` method is horrible and the strong parameters are much better since they can provide attribute authorization on a controller/action level which is where this belongs I think. The service does not change the fact that a developer can add a view parameter and the model will be updated as long as the attribute is `attr_accessible`. Maybe the service hides this fact where updating the model in the controller would not? This is possible but I think the good name of the service will prevent this along with the tests. 

I probably do not understand the suggested solution here.

@chcholman If you want to merge this in and we can address this concern when @acousticrobot is back, I am fine with that. We can also wait, it's up to the priority.

chcholman: @jwright @acousticrobot I am going to merge this and add a note that this needs to be addressed next.

PR #1596, opened 2016-01-27
hotfix - store attachement uploads for groups

### LINKED ISSUE #1592: Store attachment uploads for groups
If a student is uploading an attachment that maps to a group submission, they're erroring out here: https://github.com/UM-USElab/gradecraft-development/blob/master/app/uploaders/attachment_uploader.rb#L15 

If possible, store the submission file by group name instead of student name. If not possible, store it according to the student responsible for the actual uploading.

acousticrobot:
add group name to group submission uploads
    bugfix submissions#show and it's presenter
    fix sample data for group submissions
    closes #1592

### DIFF
--- a/app/presenters/show_submission_presenter.rb
+++ b/app/presenters/show_submission_presenter.rb
@@ -10,7 +10,11 @@ class ShowSubmissionPresenter < SubmissionPresenter
   end
 
   def grade
-    assignment.grades.where(student_id: student.id).first
+    if assignment.is_individual?
+      assignment.grades.where(student_id: student.id).first
+    else
+      assignment.grades.where(group_id: group.id).first
+    end
   end
 
   def submission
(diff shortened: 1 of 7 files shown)


jwright:
@acousticrobot This is great. I would however override `group_id` as well on the `ShowSubmissionPresenter`. This is because if a client ever chose to use `group_id` instead of `group`, then it would get consistent results.

app/presenters/show_submission_presenter.rb
```diff
@@ -10,7 +10,11 @@ def id
   end
 
   def grade
-    assignment.grades.where(student_id: student.id).first
+    if assignment.is_individual?
```
jwright: Is there a test for this branch in code? If not, there should be so it is covered.

app/presenters/show_submission_presenter.rb
```diff
@@ -10,7 +10,11 @@ def id
...
+      return assignment.grades.where(student_id: student.id).first
```
jwright: This is a style thing but I would vote for removing `return` in line 14 and line 16 since they are not needed.

jwright:
@acousticrobot I caught a few more things

app/presenters/show_submission_presenter.rb
acousticrobot: The specs for this presenter are very slim, and in the interests of getting the bugfix in I decided to not write them myself.

PR #1598, opened 2016-01-28
1504 add item count to instructor nav bar

### LINKED ISSUE #1504: Add item count number to Instructor nav bar
The grading status, resubmissions, and ungraded links in the left hand nav should all be bolded and display a number of items to manage that reflect how many elements exist on each page to address

jwright:
Adds 3 "unread" indicators for the staff sidebar which indicate the number of resubmissions that need to be graded, the number of resubmissions, and a grading status number which is a sum of unreleased grades, ungraded submissions, and in progress grades.
  
469c49d8-c4f4-11e5-9a8e-dbd75fc0e63c

This uses helpers to achieve these counts. The counts are cached using Rails caching since they are on the sidebar and we do not want to hit the database everytime we perform a request with a staff member. The caching was tested manually to make sure the numbers update when a grade updates. This works because a grade update touches the `Course` which is what the cache key is updated with.

These counts also replaced the instance variables on the respective controller/views.

Closes #1504

### DIFF
--- a/app/helpers/grades_helper.rb
+++ b/app/helpers/grades_helper.rb
@@ -14,6 +14,7 @@ module GradesHelper
   end
 
   def in_progress_grades_count_cache_key(course)
+    # This cache key is busted when a grade is updated as a Grade touches a Course
     "#{course.cache_key}/in_progress_grades_count"
   end
 
@@ -24,6 +25,7 @@ module GradesHelper
   end
 
   def unreleased_grades_count_cache_key(course)
+    # This cache key is busted when a grade is updated as a Grade touches a Course
     "#{course.cache_key}/unreleased_grades_count"
   end
 end
--- a/app/helpers/grades_helper.rb
+++ b/app/helpers/grades_helper.rb
@@ -1,0 +1,29 @@
+module GradesHelper
+  extend SubmissionsHelper
+
+  def grading_status_count_for(course)
+    unreleased_grades_count_for(course) +
+      in_progress_grades_count_for(course) +
+      ungraded_submissions_count_for(course)
+  end
+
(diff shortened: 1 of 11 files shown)


app/helpers/grades_helper.rb
```diff
@@ -0,0 +1,31 @@
+module GradesHelper
+  extend SubmissionsHelper
```
zazaian: I like how this is put together. Nice work.

spec/helpers/grades_helper_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+  describe "#in_progress_grades_count_for" do
+    let!(:grade) { create :grade, status: "In Progress", course: course }
+
+    it "returns the number of grades that are in progress" do
+      expect(in_progress_grades_count_for(course)).to eq 1
+    end
+
+    it "caches the result" do
+      expect(course).to receive(:grades).and_call_original.at_most(:twice)
+      in_progress_grades_count_for(course)
```
zazaian: Why are we tolerating two calls instead of just one here?

spec/helpers/submissions_helper_spec.rb
```diff
@@ -0,0 +1,40 @@
+require "rails_spec_helper"
+require "./app/helpers/submissions_helper"
```
zazaian: This all looks good to me. Except for the question about double-hit caching instead of single hit, this should be good to go.

spec/helpers/grades_helper_spec.rb
jwright: The calls on lines 16 and 17 is one. This test then clears the cache key and allows for one more. This test verifies that the correct cache key is used.

PR #1602, opened 2016-01-29
1597 dont strip links

### LINKED ISSUE #1597: Don't strip out links in description fields
When instructors and students add description content we strip links from it, and should not.

jwright:
This removes the ability for submission text comments to have links stripped from their descriptions. This was because it was using the `RESTRICTED` config method from [Santitize](https://github.com/rgrove/sanitize).

In order to reduce the duplication of code, the cleaning of the html was moved into a Sanitize concern and used for the following:

- `Assignment#description`
- `Grade#feedback`
- `Group#text_proposal`
- `Submission#text_comment`

In addition, since images nor tables are allowed for the methods, the froala editor options were updated to take away those buttons.

Closes #1597

### DIFF
--- a/app/assets/javascripts/utilities.js
+++ b/app/assets/javascripts/utilities.js
@@ -23,10 +23,17 @@ $( "#tabs" ).tabs({
   }
 });
 
- $('.froala').editable(
-    {inlineMode: false,
-      minHeight: 280
-    })
+ $('.froala').editable({
+  inlineMode: false,
+  minHeight: 280,
+  buttons: [
+    'fullscreen', 'bold', 'italic', 'underline', 'strikeThrough', 'subscript',
+    'superscript', 'fontFamily', 'fontSize', 'sep', 'inlineStyle', 'blockStyle',
+    'sep', 'formatBlock', 'align', 'insertOrderedList', 'insertUnorderedList',
+    'outdent', 'indent', 'insertHorizontalRule', 'createLink', 'undo', 'redo',
+    'removeFormat', 'selectAll'
+  ]
+})
 
 // handle 'select all' buttons, used on release grade forms
 $(".select-all").click(function(e){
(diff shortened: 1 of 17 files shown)


app/models/submission.rb
```diff
@@ -124,10 +126,6 @@ def permissions_check(user)
     end
   end
 
-  def clean_html
-    self.text_comment = Sanitize.clean(text_comment, Sanitize::Config::RESTRICTED)
-  end
-
   def submit_something
     link.present? || text_comment.present? || submission_files.present?
   end
```
acousticrobot: I know the method exited before this PR, but maybe there's a more descriptive name than `submit_something`?

app/assets/javascripts/utilities.js
```diff
@@ -23,10 +23,17 @@ $( "#tabs" ).tabs({
...
-    {inlineMode: false,
-      minHeight: 280
-    })
+ $('.froala').editable({
+  inlineMode: false,
+  minHeight: 280,
+  buttons: [
+    'fullscreen', 'bold', 'italic', 'underline', 'strikeThrough', 'subscript',
+    'superscript', 'fontFamily', 'fontSize', 'sep', 'inlineStyle', 'blockStyle',
+    'sep', 'formatBlock', 'align', 'insertOrderedList', 'insertUnorderedList',
```
acousticrobot: 'sep' is in there twice.

jwright: It is because `sep` is a separator and so the toolbar has a visible vertical separator.

These are basically the defaults without the file and image uploaders.

acousticrobot: OK, cool. Tomorrow we may want to revisit these button definitions too then:
angular/controllers/GradeCtrl.js.coffee:163
This one was stripped down on purpose: angular/controllers/GradeRubricCtrl.js.coffee:245

jwright: @acousticrobot Yes. Let's make sure these are standardized.

app/models/submission.rb
jwright: @acousticrobot My latest commit, I changed this to a proper validator.

PR #1603, opened 2016-01-31
Fix bug where individually awarded badges do not get set to student visible

chcholman:
This fixes a bug where instructors awarding badges one at a time would award a badge, but the student would not be able to see it, and any points earned would not be added to the student's score.


jwright:
@chcholman I think this is great! No comments.

PR #1615, opened 2016-02-03
1509 refactor leaderboard

### LINKED ISSUE #1509: Refactor leaderboard to move private methods out of controller
Per this comment: https://github.com/UM-USElab/gradecraft-development/blob/master/app/controllers/students_controller.rb#L141

jwright:
This PR simplifies the `StudentsController#leaderboard` action and view by removing private methods and refactoring them so they are in a class (presenter).

Some of the logic for finding a `GradeSchemeElement` based on the score was moved into the `GradeSchemeElement` class.

Closes #1509


app/presenters/student_leaderboard_presenter.rb
```diff
@@ -0,0 +1,105 @@
...
+  def earned_badges
+    @earned_badges ||=
+      EarnedBadge.for_course(course)
+        .where(student_id: student_ids)
+        .includes(:badge)
+  end
+
+  def grade_scheme_elements
+    @grade_scheme_elements ||=
+      GradeSchemeElement.unscoped.for_course(course).order_by_low_range
```
acousticrobot: We added default scopes for some models when they were being used all over the place in the views (for instance, Users ordered by last name). Do you feel like you are jumping through hoops to get around the default order by high range? We could look into changing/removing it. Or is there another reason to call `unscoped` here and in other places in your code?

app/models/grade_scheme_element.rb
```diff
@@ -9,8 +9,27 @@ class GradeSchemeElement < ActiveRecord::Base
...
 
+  def self.default
+    GradeSchemeElement.new(level: "Not yet on board")
+  end
+
+  def self.for_course_and_score(course, score)
+    for_score score, unscoped.for_course(course.id).order_by_low_range.to_a
+  end
+
+  def self.for_score(score, elements)
```
acousticrobot: I understand after looking at the presenter and the specs what's going on here, but it took me a second -- I thought if `#for_course_and_score` took `(course, score)` then why doesn't `#for_score` take `(score)`. I don't like `for_score_and_elements` either though.

app/models/grade_scheme_element.rb
```diff
@@ -9,8 +9,27 @@ class GradeSchemeElement < ActiveRecord::Base
...
+    element = elements.sort_by!(&:low_range)
+      .find { |element| element.within_range?(score) }
+    unless elements.empty?
```
acousticrobot: This method will return nil if `elements.empty?`. No spec for this, so wanted to confirm this is expected behavior. I don't think we handle courses without GSEs anyway so this is probably a moot point.

chcholman: (we don't handle them well, but they do occasionally happen - we need to do more to prevent that)

app/presenters/student_leaderboard_presenter.rb
jwright: @acousticrobot We have to call `unscoped` here so that we can order it differently with `order_by_low_range`.

I did a quick look and it looks like this is affecting everytime we say `current_course.grade_scheme_elements`, then it is using that default scope. I am not opposed to putting that scope on and removing default scopes (I am not a fan) but I think we should do that in a separate PR and create an issue around it. Thoughts?

app/models/grade_scheme_element.rb
jwright: @acousticrobot This was needed for performance reasons. For out use case, we want to get all of the `GradeSchemeElement` for the current course before we loop through each student. This is why the method of `for_score` exists.

We could change it to a class method on the collection.

```
class Course
  has_many :grade_scheme_elements do
    def for_score(score)
      #....
    end
  end
end
```

What do you think about that?

app/models/grade_scheme_element.rb
jwright: @acousticrobot Great. I will add a spec for that. Great catch.

jwright:
@acousticrobot Just a few questions.

app/presenters/student_leaderboard_presenter.rb
acousticrobot: Definitely should be it's own PR.

To be clear, you're not a fan of default scopes? The reason why we added them was that almost every list had to be reordered. For example dozens of `students.alphabetized ` could be simplified to just `students` throughout the code by adding the default scope, and we would never care about students ordered by the `created_at` date.  But in this case having to constantly call `unscoped` and `order_by_low_range` defeats this advantage. The default scope only should be in there if it makes all uses easier.

app/models/grade_scheme_element.rb
acousticrobot: Would you have access to it from the `LeaderboardStudentDecorator` that doesn't have the course?

app/presenters/student_leaderboard_presenter.rb
jwright: I am not usually a fan because they usually have unintended consequences or they are a pain (using `unscoped`). I like being explicit as I believe it makes it more readable.

app/models/grade_scheme_element.rb
jwright: It would. The client would use `course.grade_scheme_elements.for_score(score)`.

acousticrobot: I like it.

app/presenters/student_leaderboard_presenter.rb
acousticrobot: Might be worth reexamining then on Tuesday.

PR #1618, opened 2016-02-03
1605 Upgrade File Upload Removal to AWS SDK v2

zazaian:
File removal for all classes of file attachments (GradeFile, AssignmentFile, SubmissionFile) etc were previously using AWS v1. Those have been upgraded to v2 and specs have been added for the entire workflow (there weren't any previously). Additionally I've added some spec examples in each of the corresponding model classes to test that the S3File library is included and therefore each KlassFile class can interact with the same S3 methods in the controller. 

@jwright @acousticrobot, can one of you guys review this for me? Danke.

### DIFF
--- a/app/controllers/uploads_controller.rb
+++ b/app/controllers/uploads_controller.rb
@@ -16,10 +16,10 @@ class UploadsController < ApplicationController
   protected
 
   def fetch_upload_with_model
-    @upload = upload_klass.find params[:upload_id].to_i
+    @upload = upload_class.find params[:upload_id].to_i
   end
 
-  def upload_klass
+  def upload_class
     params[:model].classify.constantize
   end
 
--- a/app/controllers/uploads_controller.rb
+++ b/app/controllers/uploads_controller.rb
@@ -1,16 +1,19 @@
 class UploadsController < ApplicationController
+  before_filter :fetch_upload, only: :remove
+
   def remove
-    upload = params[:model].classify.constantize.find params[:upload_id]
-    s3 = AWS::S3.new
-    bucket = s3.buckets["gradecraft-#{Rails.env}"]
-    if Rails.env == "staging" || Rails.env == "production"
-      if upload.filepath.present?
-        bucket.objects[CGI::unescape(upload.filepath)].delete
-      else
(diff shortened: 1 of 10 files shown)


zazaian:
@chcholman this is tested and working properly on staging. Just need a review here. @jwright or @acousticrobot? Mind taking a look?

app/controllers/uploads_controller.rb
```diff
@@ -1,16 +1,33 @@
...
     redirect_to :back
   end
+
+  protected
+
+  def fetch_upload_with_model
+    @upload = upload_klass.find params[:upload_id].to_i
+  end
+
+  def upload_klass
```
acousticrobot: What benefit is there to spelling it klass? I've seen this to avoid the reserved word 'class', but since `upload_class` isn't reserved, I'm not sure this is necessary. Maybe if you were directly working with a `klass` variable or calling Rails klass method... http://apidock.com/rails/ActiveRecord/Reflection/AssociationReflection/klass but you're calling `classify` here.

zazaian: I guess I'm kind of allergic to using 'class' in method names regardless of whether it's expressly substituting the keyword. Because replacing 'class' with 'klass' is such a pervasive convention in ruby it seems to me like this doesn't really effect readability at all. That said, if there's some agreement that we should only be substituting the class keyword for 'klass' and in no other circumstances I guess I'm fine with it. @chcholman @jwright, any thoughts here?

chcholman:
@zazaian @acousticrobot I'm okay with either, but I'd like us to agree on one and be consistent.

zazaian:
I'm really okay with this either way, and agree that consistency is really paramount here. The argument for #something_class instead of #something_klass is that it's an easier rule to follow overall, "only use #klass when you're explicitly replacing a method called #class", but it still feels weird to me. Maybe that's just habit. Any thoughts here @jwright?

zazaian:
I'm just going to change these to #upload_class for now so we can get this merged and in use. I don't don't think it warrants any further discussion. @chcholman this is done and ready for merge.

Issue #1534, opened 2015-12-24
Assignment copy feature should include attachment


chcholman:
As an instructor, when I copy an assignment it should copy the file attachments and media images over as well

jwright:
@chcholman Does this mean that the assignment files on S3 should be copied or should the assignments share the same file?

chcholman:
@jwright Because we store files directly in relationship to the course they're in, it seems like it makes sense to directly copy them on S3 into the appropriate folder. Open to an alternate solution though!

jwright:
@chcholman No, this is should be how it works, I just wanted to make sure.

chcholman:
:+1:

jwright:
@zazaian Oh, keeper of S3 code....I talked with @chcholman and it may be best to have you jump on this....the initial branch https://github.com/UM-USElab/gradecraft-development/tree/1534-assignment-copy-attachments is out there with a broken test that should be a good spot to pick up on.

I was heading down the path of using Carrierwave so it may not be the correct place to leave off on. Please let me know if you have any questions with `ModelCopier` needs.

zazaian:
@jwright sounds good to me, not a problem. I'll definitely hit you up when it comes to the ModelCopier stuff. My guess is that we're going to want to dig through the AWS API for batch copying on S3, or maybe even do it through `s3fs` so we can manage all of the actual S3 work in one or two interactions rather than hundreds or thousands.

Either way, I'll keep this in sight. Thanks for giving me some background on this earlier today.

chcholman:
Start here: https://github.com/UM-USElab/gradecraft-development/tree/1534-assignment-copy-attachments

PR #1622, opened 2016-02-04
1394 Refactor Course

### LINKED ISSUE #1625: Move SubmissionsExport to /app/models
Move SubmissionsExport to /app/models to comply with 'active_record_spec_helper' conventions.

jwright:
I have been tasked with refactoring the `Course` object. This object has some complex methods and one of those methods is the [`copy` method](https://codeclimate.com/github/UM-USElab/gradecraft-development/Course).

In order to fix this complexity, I am suggesting we use the [extract method to object](https://sourcemaking.com/refactoring/replace-method-with-method-object) refactoring.

I am proposing a `ModelCopier` object and I have stubbed out what I think the API to that method would look like for the `Course`. Please take a look [here](https://github.com/UM-USElab/gradecraft-development/commit/94e2f190f0b74526491895eb68e46a09be178fd7#diff-72aea198f50b1d4f2407a8f9f4c904f6R189)

My fear is that the api may be a little confusing with the copy method, especially the place where we have to copy the `#assignment_types` with a specific attribute value (the `Course` id).

What do you guys think? Can you leave comments on what you think about this API. Any suggested changes with code would be helpful!

----------------------------------------------------------------------------

This PR is just for the `ModelCopier` change. The object was created and modified to handle the following scenarios:

* Copying attributes
* Copying associations
* Prepending attribute values
* Passing attributes to child associations
* Do anything via a proc

I applied the `ModelCopier` to all the models that had a custom `#copy` method.

References #1394 
Closes #1625


chcholman:
@jwright I like this approach a lot. 

I should note that one area the current copy method is broken that it would be great to keep in mind as you address this is the management of file attachments - because they are stored in a course-specific folder, and we don't currently copy them, the links exist but the files are no longer accessible.

jwright:
Related issue #1534

zazaian:
Format looks good to me. Run with it.

app/models/course.rb
```diff
@@ -186,14 +186,12 @@ def user_term
...
-    copy = self.dup
-    copy.name.prepend "Copy of "
-    copy.save unless self.new_record?
-    copy.badges << self.badges.map(&:copy)
-    copy.assignment_types << self.assignment_types.map { |at| at.copy(course_id: copy.id) }
-    copy.challenges << self.challenges.map(&:copy)
-    copy.grade_scheme_elements << self.grade_scheme_elements.map(&:copy)
-    copy
+    ModelCopier.new(self).copy(associations: [
+      :badges,
```
acousticrobot: did you want to add `attributes: attributes,` here?

app/models/course.rb
```diff
@@ -186,14 +186,12 @@ def user_term
...
+      { assignment_types: { course_id: :id }},
+      :challenges,
+      :grade_scheme_elements
+    ], options: { prepend: { name: "Copy of " }})
```
acousticrobot: did you want to add `attributes: attributes,` here?

spec/models/model_copier_spec.rb
```diff
@@ -0,0 +1,88 @@
...
+    context "with prepend option" do
+      it "prepends the specified attribute with the specified text" do
+        name = model.name
+        subject = described_class.new(model).copy options: { prepend: { name: "Copy of " }}
+
+        expect(subject.name).to eq "Copy of #{name}"
+      end
+    end
+
+    context "with overrides option" do
```
acousticrobot: This is picky, but for documentation purposes, if I didn't see how this was used in criteria or assignment, I'd wonder why you would need to handle both overrides and attributes in order to change an attribute on the model.

jwright: @acousticrobot The `overrides` option is used to set additional things on the model that are not handled within the api.

The `critereon#copy` method uses it to set the `#add_default_options` to false for example:

https://github.com/UM-USElab/gradecraft-development/pull/1622/files#diff-0171c5465b73a9515c3b2d0160c6faf2R26

jwright:
@mkoon Would you mind documenting the "copying" of models using the `ModelCopier`. I can work together with you on that if you wish.

spec/models/model_copier_spec.rb
acousticrobot: It makes sense when you see it in use in `Criteria` and `Assignment` because it goes beyond a simple attribute override. I'm just saying in the specs it's only changing an attribute. If the specs were my entry point, I might be confused why it was in necessary. Like I said, I'm being picky.

PR #1623, opened 2016-02-04
Adds API namespace to routes and controllers

acousticrobot:
remove default orders criteria and levels
    standardize scope on "ordered" for these models
    API criterion index
    API badges intex
    use new api routes for rubric front end
    improves rubric service in Angular for rubric controllers


app/assets/javascripts/angular/controllers/GradeRubricCtrl.js.coffee
```diff
@@ -223,22 +218,6 @@
       $scope.criterionGrades[rg.criterion_id] = criterionGrade
     )
 
-  $scope.addCourseBadges = (courseBadges)->
```
jwright: :+1:

app/controllers/api/badges_controller.rb
```diff
@@ -0,0 +1,15 @@
+class API::BadgesController < ApplicationController
+
+  def index
+    @badges = current_course.badges.select(
```
jwright: Why do we just select certain fields here? Performance reasons?

I fear a higher cost of developing the API if we are just selecting certain fields but returning full objects to the views. Someone may infer that the object is fully populated. What are the tradeoffs here?

app/models/criterion.rb
```diff
@@ -17,7 +17,7 @@ class Criterion < ActiveRecord::Base
   validates :name, presence: true, length: { maximum: 30 }
   validates :order, presence: true
 
-  scope :ordered, lambda { order(:order) }
+  scope :ordered, -> { order(:order) }
```
jwright: :+1:

app/controllers/rubrics_controller.rb
```diff
@@ -28,32 +28,8 @@ def update
     respond_with @rubric, status: :not_found
   end
 
-  def existing_criteria
```
jwright: :+1:

spec/views/api/badges/index_spec.rb
```diff
@@ -0,0 +1,59 @@
+# encoding: utf-8
+require 'rails_spec_helper'
+include CourseTerms
+
+describe "api/badges/index" do
+  before(:all) do
+    @world = World.create
+                  .create_course({ badge_term: "baj" })
+                  .create_badge({ description: "..." })
+    @badge = @world.badge
```
jwright: This is a style thing but we should decide on this in the future. I prefer using `let` instead of creating instance variables. For one, they are lazy loaded and they do not pollute your specs with instance variables.

jwright:
@acousticrobot This is *really* nice. I love the amount of cleanup in this PR. Really well done.

I just have a few questions for you.

app/controllers/api/badges_controller.rb
acousticrobot: This pattern originally comes from the predictor routes and I used it for a couple reasons:
1) mostly performance but also:
2) cleaner views since we can just call `json.merge! badge.attributes`
3) cleaner json with only relevant fields
4) Removing sensitive information before it makes it to the view, no need for if/else in the jbuilder 
5) with a 1:1 relationship between the controller and the view, it didn't feel to hard to see the model that is passed to the view.

With the badges, it's wouldn't be much different either way, but in the predictor there were some models that were monstrous when fully passed through that made them hard to read, and full of sensitive information.

That being said, there are some benefits to sending full objects to the jbuilder and hand-picking the relevant info there. Some benefits would be:
1) I don't have any real evidence that select is that much more performative
2) some fields are selected for logic but not actually passed through, which means the select list doesn't pair with the final json
3) the view would be more explicit if it has one line for each attribute passed through, representing the final json better.

spec/views/api/badges/index_spec.rb
acousticrobot: Sounds good -- I'll submit this as-is but clean it up on my current branch.

app/controllers/api/badges_controller.rb
jwright: I think that #2 above (first part) solves the maintenance issue. I was unaware that Rails did this, I assumed they passed back all the attributes with nils. I like this pattern now. Thanks for the thoughtful response!

jwright:
@acousticrobot I am good with this!

PR #1635, opened 2016-02-08
S3 For Development // Cache New S3 Paths // Refactor Attachment Uploader

zazaian:
This pull request implements the following changes:

* Strip out all local file handling for models using AttachmentUploader and convert to fog/S3
* Implements a AWS_S3_DEVELOPER_TAG constant in .env to sandbox developer files on S3
* Caches the #store_dir for AttachmentUploader models to prevent breakage from path changes
* Refactors AttachmentUploader and adds a comprehensive spec for it
* Stores SubmissionsExports in gradecraft-development/developer-bucket-name
* Refactors AttachmentUploader file model specs that were testing filename formatting through the URL instead of directly
* Refactor /lib/S3File into S3Mananger::Carrierwave to better model its utility, clean up legacy methods


zazaian:
@jwright please take a look at this when you get a chance. Thanks!

app/exports/submissions_export.rb
```diff
@@ -25,8 +25,25 @@ class SubmissionsExport < ActiveRecord::Base
 
   before_create :set_s3_bucket_name
 
-  def build_s3_object_key(s3_object_filename)
-    "exports/courses/#{course_id}/assignments/#{assignment_id}/#{created_at_date}/#{created_at_in_microseconds}/#{s3_object_filename}"
+  def build_s3_object_key(object_filename)
+    if Rails.env.development?
```
jwright: Can we remove this check and just depend on the environment adjusting for each environment?

app/exports/submissions_export.rb
```diff
@@ -25,8 +25,25 @@ class SubmissionsExport < ActiveRecord::Base
...
+      [ ENV['AWS_S3_DEVELOPER_TAG'] ].concat(s3_object_key_pieces(object_filename)).join("/")
+    else
+      s3_object_key_pieces(object_filename).join("/")
+    end
+  end
+
+  def s3_object_key_pieces(object_filename)
```
jwright: This seems like it should be the responsibility of another class and not for the `SubmissionsExport` class. What do you think about moving the whole building of an s3 object key to another class?

app/models/grade.rb
```diff
@@ -84,7 +84,7 @@ def self.find_or_create_grades(assignment,students)
 
   def add_grade_files(*files)
```
jwright: This method does not seem to be used except in the specs. Can this method be removed?

app/models/submission_file.rb
```diff
@@ -1,5 +1,5 @@
 class SubmissionFile < ActiveRecord::Base
-  include S3File
+  include S3Manager::Carrierwave
```
jwright: I really like these changes are moved to the `S3Manager` :+1:

app/performers/submissions_export_performer.rb
```diff
@@ -34,7 +34,7 @@ def performer_steps
       :student_directories_created_successfully, # check whether the student directories were all created successfully
```
jwright: Added an issue to refactor this class out a bit (#1636).

app/uploaders/attachment_uploader.rb
```diff
@@ -8,12 +8,13 @@ class AttachmentUploader < CarrierWave::Uploader::Base
...
   # badge_file: uploads/<course-name_id>/badge_files/<timestamp_file-name.ext>
   # challenge_file: uploads/<course-name_id>/challenge_files/<timestamp_file-name.ext>
+  #
   def store_dir
-    course = "/#{model.course.courseno}-#{model.course.id}" if model.class.method_defined? :course
-    assignment =  "/assignments/#{model.assignment.name.gsub(/\s/, "_").downcase[0..20]}-#{model.assignment.id}" if model.class.method_defined? :assignment
-    file_type = "/#{model.class.to_s.underscore.pluralize}"
-    owner = "/#{model.owner_name}" if model.class.method_defined? :owner_name
-    "uploads#{course}#{assignment}#{file_type}#{owner}"
+    if Rails.env.development?
```
jwright: This seems like it should be the responsibility of another class and not for the `AttachmentUploader` class. What do you think about moving the whole building of an s3 object key to another class?

config/environments/development.rb
```diff
@@ -30,5 +30,5 @@
 end
 
 CarrierWave.configure do |config|
-  config.storage = :file
+  config.storage = :fog
```
jwright: Do we want to connect to S3 for development?

config/environments/test.rb
```diff
@@ -13,9 +13,8 @@
   config.session_store :cookie_store, key: '_gradecraft_session', :expire_after => 60.minutes
 end
 
-
 CarrierWave.configure do |config|
-  config.storage = :file
+  config.storage = :fog
```
jwright: I don't think we should connect to S3 for test unless we have the S3 API stubbed out.

spec/active_record_spec_helper.rb
```diff
@@ -70,6 +73,7 @@ def cache_dir
       File.join(File.dirname(__FILE__), "support/uploads/tmp")
     end
 
+    next if klass == AttachmentUploader
```
jwright: Why is this class special?

spec/models/assignment_file_spec.rb
```diff
@@ -1,9 +1,14 @@
-require "active_record_spec_helper"
+require "rails_spec_helper"
```
jwright: Why do we need to change this to `rails_spec_helper`?

spec/models/badge_file_spec.rb
```diff
@@ -1,9 +1,13 @@
-require "active_record_spec_helper"
+require "rails_spec_helper"
```
jwright: Why do we need to change this to `rails_spec_helper`?

spec/support/uni_mock/rails.rb
```diff
@@ -0,0 +1,9 @@
+module UniMock
```
jwright: What is `UniMock` mean?

spec/toolkits/models/shared/files_toolkit.rb
```diff
@@ -0,0 +1,14 @@
+module Toolkits
+  module Models
+    module Shared
+      module Files
+
+        def define_context
```
jwright: I don't believe that `define_context` describes what this method does. Can we think of a more self-documenting name?

spec/toolkits/uploaders/attachment_uploader.rb
```diff
@@ -0,0 +1,38 @@
+module Toolkits
+  module Uploaders
+    module AttachmentUploader
```
jwright: Since this is geared towards a single class/functionality, does it belong in "toolkits"?

spec/uploaders/attachment_uploader_spec.rb
```diff
@@ -0,0 +1,237 @@
+require 'rails_spec_helper'
+
+include Toolkits::Uploaders::AttachmentUploader
```
jwright: These includes should be included in the `RSpec.describe` block so they don't pollute the whole test suite.

app/exports/submissions_export.rb
zazaian: Are you suggesting that we just compact out the developer tag if it doesn't exist in other environments, or set a config variable for those other environments? My concern with the former is that if the developer tag is included in one of the other environments it'll break the S3 implementation, so it feels safer to expressly exclude the behavior here by environment. Perhaps I'm missing your suggestion, though.

app/exports/submissions_export.rb
zazaian: All that this method is doing is joining these pieces so we don't have to split them up and then concat the developer key, and re-join them. Is calling a #join on an array enough functionality to warrant a class? Probably not. Also this is only being used in this class.

Perhaps we could extract a module that manages s3 url building that could be included in models that have files being stored on S3, but I think this data would still want to live inside of the SubmissionsExport.

app/models/grade.rb
zazaian: Sure, I just amended this because it broke when caching was implemented for GradeFiles. This isn't really related to this pull request otherwise. This is really a separate issue.

app/performers/submissions_export_performer.rb
zazaian: Yeah, this is a priority, but probably not in the top 3 or 5 at the moment.

app/uploaders/attachment_uploader.rb
zazaian: I think I see what you're saying now. Like and S3Manager::ObjectPath object that takes pieces and assembles them? I still don't see how this gives us a lot of value considering how little is being done here. Also it seems like it wouldn't belong in lib since we're expressly declaring conditional behavior relative to environment. Do you have an example of what you're thinking here?

config/environments/development.rb
zazaian: For testing actual S3 behaviors it's very cumbersome to have to deploy to staging to test outcomes. We could probably split out an s3_development environment but that seems like it might be overblown. Either way S3 is integrated into enough of our file management workflow that replacing all of the s3 behaviors with local upload equivalents really isn't representative of actual behavior in a staging or production environment.

spec/active_record_spec_helper.rb
zazaian: This is a monkey patch for the aforementioned Carrierwave bug in regard to overwriting store dir. Now also the only uploader class that has specs of any kind.

spec/models/badge_file_spec.rb
zazaian: Currently `active_record_spec_helper` isn't implementing toolkits, and I didn't want to include all of them just to accommodate these couple of specs. That said, I think it's fine to include those, but I figured there was a reason that you'd expressly left them out.

spec/support/uni_mock/rails.rb
zazaian: This or some other module should serve as a space to condense app-wide stubbing behavior that isn't built into RSpec or any other stubbing libraries that we're using. I've added default behaviors for properly stubbing environment and time. Since these seem to be the 'correct' way to implement these, it makes more sense to keep them in one place to keep the specs DRY than to have to declare the same circumstances in dozens of spec files.

spec/toolkits/models/shared/files_toolkit.rb
zazaian: I agree. How about `open_toolkit` or `use_toolkit`?

spec/uploaders/attachment_uploader_spec.rb
zazaian: Agreed, this is an oversight.

spec/toolkits/uploaders/attachment_uploader.rb
zazaian: I'm not sure there's anything offensive about toolkits being used in only a single context. Where else would this go?

config/environments/test.rb
zazaian: I think this can be swapped back to :file without incident. There's a bug in carrierwave where it's automatically overwriting store_dir unless the config is monkeypatched by declaring the class that includes Carrierwave before declaring the configuration block. I'll test this with :file and see how it behaves.

Either way I think stubbing S3 for test is a priority.

zazaian:
This has been tested and confirmed working on staging for both caching of new asset paths as well as correct usage of older files using the :filepath attribute.

app/uploaders/attachment_uploader.rb
jwright: I thought this was on the `Attachment` class and not the `AttachmentUploader` class so I do think this is the responsibility of the uploader.

I was originally thinking something like this `AttachmentFile.new(self).path`.

config/environments/development.rb
jwright: I agree with this.

config/environments/test.rb
jwright: Awesome. Can we configure this back to `:file` or are you saying that should be done in another issue?

spec/active_record_spec_helper.rb
jwright: :+1:

spec/models/badge_file_spec.rb
jwright: I think we should include the toolkits in the specs that need them instead of changing this to use `rails_spec_helper` because it slows down the test tremendously.

spec/support/uni_mock/rails.rb
jwright: I agree. Nice™.

spec/toolkits/models/shared/files_toolkit.rb
jwright: How about `define_file_attributes`?

jwright:
@zazaian I just have a comment about the rails spec helper require, everything else looks good (minus the failing build :/).

app/performers/submissions_export_performer.rb
zazaian: I take that back, looks like this is #4 :) Thanks for adding the issue here :+1:

app/uploaders/attachment_uploader.rb
zazaian: Got it :+1:

config/environments/development.rb
zazaian: Cool :+1:

config/environments/test.rb
zazaian: I'll test this with :file and see if it breaks any of the S3 specs. If it doesn't play nice, let's include this in the 'stub s3 for testing' issue and leave it alone for now. Cool?

spec/models/badge_file_spec.rb
zazaian: Sounds good. You're definitely right here, this spec went from 1.3 seconds to 0.21 seconds with the active record spec helper. I didn't realize this was making such a drastic difference. :+1:

I'll expressly require the toolkit and mocking library being used here.

spec/toolkits/models/shared/files_toolkit.rb
zazaian: I don't like define_file_attributes because that's really only going to make sense in the context of this particular toolkit. There are a few different toolkits with this same behavior (context being defined from a module extension and then calling the encapsulating method post hoc) and it feels like the best approach to name them all similarly. Maybe define_attributes?

zazaian: Actually, in retrospect, if they all follow the `define_resourcename_attributes` structure for all of them then I think that makes sense. Thoughts?

zazaian:
@jwright fyi the failing spec on that build was from the HumanHistory::ActorHistoryToken spec. It looks like a local variable of 'subject' was being set and thus sporadically overwritten by the 'subject' method. I refactored the nomenclature on those examples with contexts and they seem to be working now.

```
Failures:

1) HumanHistory::ActorHistoryToken#parse returns `Someone` if the user is not found
Failure/Error: expect(subject.parse).to eq({ actor: "Someone" })

  expected: {:actor=>"Someone"}
       got: {:actor=>"Guiseppe Purdy"}

  (compared using ==)

  Diff:
  @@ -1,2 +1,2 @@
  -:actor => "Someone",
  +:actor => "Guiseppe Purdy",
  
# ./spec/human_history/actor_history_token_spec.rb:27:in `block (3 levels) in <top (required)>'
# ./spec/active_record_spec_helper.rb:96:in `block (3 levels) in <top (required)>'
# /home/rof/cache/bundler/ruby/2.2.0/gems/activerecord-4.2.3/lib/active_record/connection_adapters/abstract/database_statements.rb:213:in `block in transaction'
# /home/rof/cache/bundler/ruby/2.2.0/gems/activerecord-4.2.3/lib/active_record/connection_adapters/abstract/transaction.rb:184:in `within_new_transaction'
# /home/rof/cache/bundler/ruby/2.2.0/gems/activerecord-4.2.3/lib/active_record/connection_adapters/abstract/database_statements.rb:213:in `transaction'
# /home/rof/cache/bundler/ruby/2.2.0/gems/activerecord-4.2.3/lib/active_record/transactions.rb:220:in `transaction'
# ./spec/active_record_spec_helper.rb:95:in `block (2 levels) in <top (required)>'
# ./spec/rails_spec_helper.rb:71:in `block (3 levels) in <top (required)>'
# /home/rof/cache/bundler/ruby/2.2.0/gems/database_cleaner-1.5.1/lib/database_cleaner/generic/base.rb:16:in `cleaning'
# /home/rof/cache/bundler/ruby/2.2.0/gems/database_cleaner-1.5.1/lib/database_cleaner/base.rb:92:in `cleaning'
# /home/rof/cache/bundler/ruby/2.2.0/gems/database_cleaner-1.5.1/lib/database_cleaner/configuration.rb:86:in `block (2 levels) in cleaning'
# /home/rof/cache/bundler/ruby/2.2.0/gems/database_cleaner-1.5.1/lib/database_cleaner/configuration.rb:87:in `call'
# /home/rof/cache/bundler/ruby/2.2.0/gems/database_cleaner-1.5.1/lib/database_cleaner/configuration.rb:87:in `cleaning'
# ./spec/rails_spec_helper.rb:70:in `block (2 levels) in <top (required)>'

Finished in 10 minutes 55 seconds (files took 9.35 seconds to load)
3302 examples, 1 failure, 36 pending
```

config/environments/test.rb
zazaian: I swapped this back to :file and everything ran cleanly. All good here :+1:

jwright:
@zazaian I found a few things but it looks good. (Rebase needed :cry:)

zazaian:
@jwright sorry if I used let(:subject) there, that's obviously wrong. I was just patching it to get it working.

zazaian:
@jwright there are two different libraries being pulled in by support here, so I just required all of the individual files expressly.

Issue #1384, opened 2015-11-17
Refactor & test Course model


chcholman:

chcholman:
Currently: 72.7% Test Coverage, Grade: F

PR #1638, opened 2016-02-08
1384 refactor timeline

jwright:
Refactors the `Course` object by removing the complex method `#timeline_events` into a `Timeline` object.

The `Timeline` object gathers all of the events for a specifc `Course`.

References #1384

### DIFF
--- a/app/controllers/info_controller.rb
+++ b/app/controllers/info_controller.rb
@@ -21,7 +21,7 @@ class InfoController < ApplicationController
   end
 
   def timeline_events
-    @events = current_course.timeline_events
+    @events = Timeline.new(current_course).events
     render(:partial => 'info/timeline', :handlers => [:jbuilder], :formats => [:js])
   end
 
--- a/app/models/timeline.rb
+++ b/app/models/timeline.rb
@@ -4,4 +4,24 @@ class Timeline
   def initialize(course)
     @course = course
   end
+
+  def assignment_events
+    course.assignments.timelineable.with_dates
+  end
+
+  def event_events
+    course.events.with_dates
+  end
+
+  def challenge_events
+    course.challenges.with_dates if course.has_team_challenges?
+  end
+
+  def events
+    @events ||= [
+      assignment_events.to_a,
+      challenge_events.to_a,
+      event_events.to_a
+    ].flatten
+  end
 end
--- a/app/models/timeline.rb
+++ b/app/models/timeline.rb
@@ -1,0 +1,7 @@
+class Timeline
+  attr_reader :course
+
+  def initialize(course)
+    @course = course
+  end
+end
--- a/spec/models/timeline_spec.rb
+++ b/spec/models/timeline_spec.rb
@@ -11,29 +11,41 @@ describe Timeline do
   end
 
   describe "#events" do
-    it "includes assignments that have due dates or open dates" do
-      assignment = create(:assignment, course: course, due_at: Date.today)
-      assignment_no_date = create(:assignment, course: course, due_at: nil)
-      expect(subject.events).to eq [assignment]
+    context "with assignments for the course" do
+      let!(:assignment) { create :assignment, course: course, due_at: Date.today }
+      let!(:assignment_not_due) { create :assignment, course: course, due_at: nil }
+
+      it "includes the assignments" do
+        expect(subject.events).to eq [assignment]
+      end
     end
 
-    it "includes events that have due dates or open dates" do
-      event = create(:event, course: course, due_at: Date.today)
-      event_no_date = create(:event, course: course, due_at: nil)
-      expect(subject.events).to eq [event]
+    context "with events for the course" do
+      let!(:event) { create :event, course: course, due_at: Date.today }
+      let!(:event_not_due) { create :event, course: course, due_at: nil }
+
+      it "includes the events" do
+        expect(subject.events).to eq [event]
+      end
     end
 
-    it "includes challenge that have due dates or open dates if the course allows it" do
-      course.update_attributes team_challenges: true
-      challenge = create(:challenge, course: course, due_at: Date.today)
-      challenge_no_date = create(:challenge, course: course, due_at: nil)
-      expect(subject.events).to eq [challenge]
-    end
+    context "with challenges for the course" do
+      let!(:challenge) { create :challenge, course: course, due_at: Date.today }
+      let!(:challenge_not_due) { create :challenge, course: course, due_at: nil }
 
-    it "does not include challenge that have due dates or open dates if the course does not allows it" do
-      challenge = create(:challenge, course: course, due_at: Date.today)
-      challenge_no_date = create(:challenge, course: course, due_at: nil)
-      expect(subject.events).to eq []
+      context "that accepts team challenges" do
+        before { course.update_attributes team_challenges: true }
+
+        it "includes the challenges" do
+          expect(subject.events).to eq [challenge]
+        end
+      end
+
+      context "that does not accept team challenges" do
+        it "does not include the challenges" do
+          expect(subject.events).to eq []
+        end
+      end
     end
   end
 end
--- a/spec/models/timeline_spec.rb
+++ b/spec/models/timeline_spec.rb
@@ -2,16 +2,15 @@ require "active_record_spec_helper"
 
 describe Timeline do
   let(:course) { create :course }
+  subject { described_class.new(course) }
 
   describe "#initialize" do
     it "is initialized with a course" do
-      expect(described_class.new(course).course).to eq course
+      expect(subject.course).to eq course
     end
   end
 
   describe "#events" do
-    subject { described_class.new(course) }
-
     it "includes assignments that have due dates or open dates" do
       assignment = create(:assignment, course: course, due_at: Date.today)
       assignment_no_date = create(:assignment, course: course, due_at: nil)
--- a/spec/models/timeline_spec.rb
+++ b/spec/models/timeline_spec.rb
@@ -1,0 +1,11 @@
+require "active_record_spec_helper"
+
+describe Timeline do
+  let(:course) { create :course }
+
+  describe "#initialize" do
+    it "is initialized with a course" do
+      expect(described_class.new(course).course).to eq course
+    end
+  end
+end
--- a/spec/models/timeline_spec.rb
+++ b/spec/models/timeline_spec.rb
@@ -8,4 +8,33 @@ describe Timeline do
       expect(described_class.new(course).course).to eq course
     end
   end
+
+  describe "#events" do
+    subject { described_class.new(course) }
+
+    it "includes assignments that have due dates or open dates" do
+      assignment = create(:assignment, course: course, due_at: Date.today)
+      assignment_no_date = create(:assignment, course: course, due_at: nil)
+      expect(subject.events).to eq [assignment]
+    end
+
+    it "includes events that have due dates or open dates" do
+      event = create(:event, course: course, due_at: Date.today)
+      event_no_date = create(:event, course: course, due_at: nil)
+      expect(subject.events).to eq [event]
+    end
+
+    it "includes challenge that have due dates or open dates if the course allows it" do
+      course.update_attributes team_challenges: true
+      challenge = create(:challenge, course: course, due_at: Date.today)
+      challenge_no_date = create(:challenge, course: course, due_at: nil)
+      expect(subject.events).to eq [challenge]
+    end
+
+    it "does not include challenge that have due dates or open dates if the course does not allows it" do
+      challenge = create(:challenge, course: course, due_at: Date.today)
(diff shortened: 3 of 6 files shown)


app/controllers/info_controller.rb
```diff
@@ -21,7 +21,7 @@ def dashboard
   end
 
   def timeline_events
-    @events = current_course.timeline_events
+    @events = Timeline.new(current_course).events
```
zazaian: Why are we considering Timeline a model if it's not interacting with ActiveRecord outside of utilizing actual models? Can we find a better place for this?

app/models/timeline.rb
```diff
@@ -0,0 +1,27 @@
+class Timeline
```
zazaian: This looks better organized than #timeline_events on Course. Can we find a better place for it though?

spec/models/timeline_spec.rb
```diff
@@ -0,0 +1,40 @@
+require "active_record_spec_helper"
+
+describe Timeline do
+  let(:course) { create :course }
+
+  describe "#initialize" do
+    it "is initialized with a course" do
+      expect(described_class.new(course).course).to eq course
```
zazaian: Can we use Timeline instead of described_class since we're not nesting it inside of a #shared_examples block?

spec/models/timeline_spec.rb
```diff
@@ -0,0 +1,40 @@
+require "active_record_spec_helper"
+
+describe Timeline do
+  let(:course) { create :course }
+
```
zazaian: Should we create a helper for timeline since it's being used multiple times?
`let(:timeline) { Timeline.new(course) }`

spec/models/timeline_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+  describe "#initialize" do
+    it "is initialized with a course" do
+      expect(described_class.new(course).course).to eq course
+    end
+  end
+
+  describe "#events" do
+    subject { described_class.new(course) }
```
zazaian: Can we call #events on the timeline within the subject rather than on the subject since it's used in every example
`subject { timeline.events }`

spec/models/timeline_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+
+    it "includes assignments that have due dates or open dates" do
```
zazaian: Can we break these out into describe blocks to isolate what we're actually testing?
  ```  
  describe "an assignment" do
      let(:assignment) { create(:assignment, course: course, due_at: due_date) }
      before(:each) { assignment }

      # also can we break out the cases here into contexts to move context out of the example blocks?
      context "that has a due date" do
        let(:due_date) { Date.today }
        it "is included" do
          expect(subject).to include assignment
        end
      end

      context "that has no due date" do
        let(:due_date) { nil }
        it "isn't included" do
          expect(subject).to be_empty
        end
      end
    end
```

File: spec/models/timeline_spec.rb [hunk 7]
```diff
@@ -0,0 +1,40 @@
...
+      expect(subject.events).to eq [assignment]
+    end
+
+    it "includes events that have due dates or open dates" do
+      event = create(:event, course: course, due_at: Date.today)
+      event_no_date = create(:event, course: course, due_at: nil)
+      expect(subject.events).to eq [event]
+    end
+
+    it "includes challenge that have due dates or open dates if the course allows it" do
```
zazaian: Since these examples are the same perhaps we should isolate them in separate contexts to better clarify those differences?
```
describe "a challenge" do
  let(:challenge) { create(:challenge, course: course, due_at: due_date) }
  before(:each) { challenge }

  context "course allows team challenges" do
    before(:each) { course.update_attributes team_challenges: true }

    context "that has a due date" do
      let(:due_date) { Date.today }
      it "is included" do
        expect(subject).to include challenge
      end
    end

    context "that has no due date" do
      let(:due_date) { nil }
      it "isn't included" do
        expect(subject).to be_empty
      end
    end
  end

  context "course does not allow team challenges" do
    let(:due_date) { Date.today }
    it "isn't included" do
      expect(subject).to be_empty
    end
  end
end
```

app/controllers/info_controller.rb
jwright: I believe models do not need to be subclasses of ActiveRecord to make them models. This represents a domain class and it interacts with other domain classes so I think it is ok to place them in models. We have several non-AR backed models in the models directory as well.

app/models/timeline.rb
jwright: I am trying to over-polute models with simple, single responsibility classes. Rather then adding a method on an object that is responsible for storing the state of a course, I believe the gathering of timeline events should be it's own class. I think this is a much better place for it. Can you let me know why you feel it's better as a method on `course`?

spec/models/timeline_spec.rb
jwright: I think `described_class` is better for 2 reasons. 

1. Changing the name of the class will not cause errors if it is changed
2. It helps define what the SUT is

What are your thoughts around that?

spec/models/timeline_spec.rb
jwright: I moved the `subject` up so that all specs can use it. Good catch.

spec/models/timeline_spec.rb
jwright: The subject under test is `Timeline`. We are invoking a method on the subject. I think the subject should be the object, not the method.

I think we should have a specific act operation by calling the method on the subject under test.

spec/models/timeline_spec.rb
jwright: I like the context here but I believe the `describe` block is wrong. We are not verifying anything on "an assignment" but we are verifying that the subject (Timeline) contains an assignment or does not contain an assignment.

What do you think about this?

```
describe Timeline do
  describe "#events" do
    context "with an assignment that has a due date" do
      it "includes the assignment" do
      end
    end

    context "with an assignment that does not have a due date" do
      it "does not include the assignment" do
      end
    end
  end
end
```

spec/models/timeline_spec.rb
jwright: I like this except with the same comments above as having a "a challenge" describe block.

jwright:
@zazaian Comments provided.

app/controllers/info_controller.rb
zazaian: Technically in domain-driven design everything is a model, but Rails appropriates the term 'model' to mean 'ORM-adapting model'. If Rails had observed a purist approach to DDD /app would probably have a /models subdirectory, which would have directories named something to the effect of (deliberately overblown for demonstrative purposes) orm_models, controller_models, exporter_models etc. Because 'model' has such a strong association with 'orm model' in the Rails universe I don't think it's fair to say that it can also adopt the broad DDD meaning of 'model' since nothing else in Rails does. We should probably find a better space for non-orm models and accept that Rails handles orm models in an opinionated that deviates from DDD ethos in terms of nomenclature.

Fundamentally Timeline here is just gathering associations from orm models and putting them in one place. Maybe this would fit into /app/organizers? Or maybe even a presenter?

zazaian: Technically in domain-driven design everything is a model, but Rails appropriates the term 'model' to mean 'ORM-adapting model'. If Rails had observed a purist approach to DDD /app would probably have a /models subdirectory, which would have directories named something to the effect of (deliberately overblown for demonstrative purposes) orm_models, controller_models, exporter_models etc. Because 'model' has such a strong association with 'orm model' in the Rails universe I don't think it's fair to say that it can also adopt the broad DDD meaning of 'model' since nothing else in Rails does. We should probably find a better space for non-orm models and accept that Rails handles orm models in an opinionated that deviates from DDD ethos in terms of nomenclature.

Fundamentally Timeline here is just gathering associations from orm models and putting them in one place. Maybe this would fit into /app/organizers? Or maybe even a presenter?

app/models/timeline.rb
zazaian: I'm sorry if I wasn't clear here, I'm saying that I like that you broke the #timeline_events method out into the Timeline class.

spec/models/timeline_spec.rb
zazaian: I think that clarity here is more important than accommodating the hypothetical of "we might change this class name eventually". I think #described_class works best for included shared_example blocks that could be used in multiple contexts in which the described_class isn't clear. Since in this case this logic really only applies to Timeline, it might be best to just use the Timeline constant.

spec/models/timeline_spec.rb
zazaian: Cool, danke :+1:

spec/models/timeline_spec.rb
zazaian: Actually in this context the SUT is events. I suppose technically the broader SUT is Timeline, but it seems to make more sense to hone the subject in relative to the contextual blocks rather than relative to the entire spec. Thoughts?

spec/models/timeline_spec.rb
zazaian: So this is actually almost how i'd written it out initially, but I stripped it back because I was just trying to indicate the structure rather than the nomenclature. If we're going this far I think the nomenclature should actually be  `context "assignment has a due date"` since we're presuming that each of these tests is going to have an assignment.

I think the titles on the example blocks are perfect.

That said, just for reference, the half-baked "an assignment" nomenclature is definitely less expressive in the context of the spec itself, but has the benefit of producing a more succinct output:

```
Timeline
  #initialize
    is initialized with a course
  #events
    an assignment
      that has a due date
        is included
      that has no due date
        isn't included
    a challenge
      course does not allow team challenges
        isn't included
      course allows team challenges
        that has a due date
          is included
        that has no due date
          isn't included
    an event
      that has a due date
        is included
      that has no due date
        isn't included
```

So if you're reading the spec in order it says "an assignment that has a due date is included".

Either way, I think we're in agreement here. The more expressive context names are definitely preferred.

spec/models/timeline_spec.rb
zazaian: Cool, I think we're in agreement on that :+1:

app/controllers/info_controller.rb
jwright: In domain-driven design, not everything is a model. There are supporting classes that help with services and "external frameworks". AR is an external framework and our architecture should be about intent, not the framework itself.

Rails does not do anything special with the `/models` directory except autoload it. It does not care if the objects inside the directory are descendants of AR. It is simply a place for code organization.

What we decide here, we should make it a guideline for [Gradecraft Guides](https://github.com/UM-USElab/gradecraft-guides). We will need to find places for the following classes if we say "only put AR subclasses in models":

* assignment_type_weight.rb
* course_membership_builder.rb
* course_router.rb
* grade_scheme_element_scoring_extension.rb
* history_filter.rb
* model_copier.rb
* null_grade.rb
* null_student.rb
* resubmission.rb
* role.rb

I think we need another developer here to break this tie. @acousticrobot @chcholman thoughts?

app/models/timeline.rb
jwright: :+1: Gotcha. Thanks!

spec/models/timeline_spec.rb
jwright: Fixed

jwright:
@zazaian I fixed the spec organization issue. Much better, thanks.

I brought in reinforcements so we can decide as a team about the "no AR models in models dir" guideline and we can put a guideline in place for this PR and future code. :dancers:

spec/models/timeline_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+  subject { described_class.new(course) }
+
+  describe "#initialize" do
+    it "is initialized with a course" do
+      expect(subject.course).to eq course
+    end
+  end
+
+  describe "#events" do
+    context "with assignments for the course" do
```
zazaian: Slexy :+1:

spec/models/timeline_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+      let!(:assignment) { create :assignment, course: course, due_at: Date.today }
+      let!(:assignment_not_due) { create :assignment, course: course, due_at: nil }
+
+      it "includes the assignments" do
+        expect(subject.events).to eq [assignment]
+      end
+    end
+
+    context "with events for the course" do
+      let!(:event) { create :event, course: course, due_at: Date.today }
```
zazaian: Nice use of let! to trigger caching.

app/controllers/info_controller.rb
chcholman: @jwright @zazaian I'm okay with it, but that's partly because we've always had structs in this folder - that took me a while to get used to, but I expect it now. Perhaps worth (merging this work now but) having a #tuesdaychat about the patterns we see emerging here and where they should be stored? There's something consistent about how these are making sense of things.

PR #1639, opened 2016-02-08
Adds summary comments to rubric grades

acousticrobot:
* summary comments textarea on rubric grade form
  * better grade status handling
  * better specs for API routes
  * adds feedback to sample data

### DIFF
--- a/app/assets/javascripts/angular/controllers/GradeRubricCtrl.js.coffee
+++ b/app/assets/javascripts/angular/controllers/GradeRubricCtrl.js.coffee
@@ -1,33 +1,22 @@
 @gradecraft.controller 'GradeRubricCtrl', ['$scope', 'Restangular', 'Criterion', 'CourseBadge', 'CriterionGrade','RubricService', '$http', ($scope, Restangular, Criterion, CourseBadge, CriterionGrade, RubricService, $http) ->
 
   $scope.criterionGrades = {} # index in hash with criterion_id as key
-  $scope.gsiGradeStatuses = ["In Progress", "Graded"]
-  $scope.professorGradeStatuses = ["In Progress", "Graded", "Released"]
-  $scope.urlId = parseInt(window.location.pathname.split('/')[2])
+  $scope.noReleaseGradeStatuses = ["In Progress", "Graded"]
+  $scope.releaseNecessaryGradeStatuses = ["In Progress", "Graded", "Released"]
+  $scope.assignmentId = parseInt(window.location.pathname.split('/')[2])
+  $scope.studentId = parseInt(window.location.search.match(/student_id=(\d+)/)[1])
 
-  $scope.init = (rubricId, assignmentId, studentId, criterionGrades, gradeStatus, releaseNecessary, returnURL)->
-    $scope.rubricId = rubricId
-    $scope.assignmentId = assignmentId
-    $scope.studentId = studentId
-    if gradeStatus
-      $scope.gradeStatus = gradeStatus
-    else
-      $scope.gradeStatus = ""
-
-    # use 'Graded' for all assignments for which release is necessary, bypass ui
-    $scope.releaseNecessary = releaseNecessary
-    if not $scope.releaseNecessary
-      $scope.gradeStatus = "Graded"
-
+  $scope.init = (criterionGrades, returnURL)->
     $scope.returnURL = returnURL
-
     $scope.addCriterionGrades(criterionGrades)
 
   RubricService.getBadges()
-  RubricService.getCriteria($scope.urlId, $scope)
+  RubricService.getCriteria($scope.assignmentId, $scope)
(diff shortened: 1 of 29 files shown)


app/assets/javascripts/angular/controllers/GradeRubricCtrl.js.coffee
```diff
@@ -1,33 +1,22 @@
...
-    if gradeStatus
-      $scope.gradeStatus = gradeStatus
-    else
-      $scope.gradeStatus = ""
-
-    # use 'Graded' for all assignments for which release is necessary, bypass ui
-    $scope.releaseNecessary = releaseNecessary
-    if not $scope.releaseNecessary
-      $scope.gradeStatus = "Graded"
+  $scope.noReleaseGradeStatuses = ["In Progress", "Graded"]
```
jwright: This is essentially duplicated code with what we have on the server. I wonder if it's worth it to have this file go through the erb conversion and use a method here to retrieve the statuses for grades.

```
$scope.noReleaseGradeStatuses = [<%= Grade.STATUSES.join(",") %>]
```

app/assets/javascripts/angular/controllers/GradeRubricCtrl.js.coffee
```diff
@@ -221,7 +215,21 @@
...
+    buttons: [
+      "bold", "italic", "underline", "strikeThrough",
+      "insertOrderedList", "insertUnorderedList"
+    ]
+  }
+
+  $scope.froalaSummaryOptions = {
+    inlineMode: false,
+    minHeight: 100,
+    buttons: [
```
jwright: Can we not duplicate these options here? Maybe we can include a `FroalaOptions` function that all of these can use?

app/controllers/api/grades_controller.rb
```diff
@@ -0,0 +1,11 @@
+class API::GradesController < ApplicationController
+
+  before_filter :ensure_staff?
+
+  # GET api/assignments/:id/student/:student_id/grade
+  def show
+    @grade = Grade.find_or_create(params[:id],params[:student_id])
```
jwright: style: space after commas

app/services/creates_grade/builds_grade.rb
```diff
@@ -14,7 +14,8 @@ class BuildsGrade
 
         grade.raw_score = context[:attributes]["points_given"]
         grade.point_total = context[:assignment].point_total
-        grade.status =  context[:attributes]["grade_status"]
+        grade.status =  context[:attributes]["grade"]["status"]
```
jwright: Any need to worry if `["grade"]` attribute is not passed in?

app/views/api/grades/show.json.jbuilder
```diff
@@ -0,0 +1,14 @@
+json.data do
```
jwright: :+1:

spec/views/api/badges/index_spec.rb
```diff
@@ -1,12 +1,11 @@
 # encoding: utf-8
```
jwright: Why do we convert all of the double quotes to single quotes here?

I was going to propose a style that all quotes should be double. The reason is for consistency and you have to use double quotes for string interpolation so they might as well all be double quotes.

Thoughts?

jwright:
@acousticrobot I added a few comments

app/assets/javascripts/angular/controllers/GradeRubricCtrl.js.coffee
acousticrobot: What I'd like to do is abstract both the froala options and the textarea html into it's own component, but I've been holding off on this, since this entire file and the rubric_grade_form haml should be refactored into components.

app/services/creates_grade/builds_grade.rb
acousticrobot: No, status will always be present, and feedback will be blank in the params if these aren't filled in.

spec/views/api/badges/index_spec.rb
acousticrobot: I don't know why I thought you were using single quotes in your specs, but I believed I was the odd one out with double quotes and was trying to standardize. I'll go back to double since that's my preference too.

app/assets/javascripts/angular/controllers/GradeRubricCtrl.js.coffee
acousticrobot: I'm thinking it's not worth erb just for these two lines because I'm hoping they won't stick around. I didn't like keeping them either because I don't think Angular should be handling any of the grade logic, but I wanted to get the PR in, and then see if I could either put this right into the haml which is littered with Angular anyway, or get it out of both files and into it's own component.

app/controllers/api/grades_controller.rb
acousticrobot: :+1:

PR #1646, opened 2016-02-10
1641 track submission file changes

### LINKED ISSUE #1641: Submission history should track file uploads/additions
As a student if I upload a new file, it should be tracked by submission history.

jwright:
Adds submission files to the submission history. 

Since a `SubmissionFile` is an association, it has it's own version set in [PaperTrail](https://github.com/airblade/paper_trail). This PR merges that in with the submission history (submissions and grades) in order to display the history of the submission file.

Since "Submission file" is not the [preferred nomenclature dude](https://www.youtube.com/watch?v=OYOzUHnPJvU), this PR has the added ability to rename the object for display purposes. We use this to rename "Submission file" to "Attachment".

screen shot 2016-02-10 at 4 02 42 pm

NOTE: There is a better way to display this data as it should be displayed within the "Submission" history changes themselves. This is a future issue that will be created.

Closes #1641

### DIFF
--- a/app/models/history_filter.rb
+++ b/app/models/history_filter.rb
@@ -44,6 +44,17 @@ class HistoryFilter
     self
   end
 
+  def rename(options)
+    object_key = options.keys.first
+    object_value = options.values.first
+
+    history.map(&:changeset).each do |changeset|
+      object = changeset["object"]
+      changeset["object"] = object_value if object == object_key
+    end
+    self
+  end
+
   def empty_changeset?(set)
     set.values.none? { |value| value.is_a? Array }
   end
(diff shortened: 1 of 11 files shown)


app/models/history_filter.rb
```diff
@@ -44,6 +44,17 @@ def remove(options)
...
+    object_key = options.keys.first
+    object_value = options.values.first
+
+    history.map(&:changeset).each do |changeset|
+      object = changeset["object"]
+      changeset["object"] = object_value if object == object_key
+    end
+    self
+  end
+
```
acousticrobot: :+1: This is a very clean enhancement.

PR #1650, opened 2016-02-11
1384 refactor course

jwright:
Refactor the course model and reduce the complexity of it by extracting the updating of course memberships for instructors of record to an `InstructorOfRecord` class.

Created a new, non AR backed, model that interacts with instructors of record for `CourseMembership`. This includes the updating of the course memberships based on user ids and the retrieval of users that are marked as instructors of record. Doing so also eliminated a method on `User` that should not of been there.

Removed unused `.students_auditing` methods from `Course` and `User`.

NOTE: This introduces a new, non AR backed, class into the `app/models` directory. Depending on the answer to this open PR question: https://github.com/UM-USElab/gradecraft-development/pull/1638#discussion_r52484671, this may need to be moved.

References #1384

### DIFF
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -25,22 +25,7 @@ class Course < ActiveRecord::Base
 
   def instructors_of_record_ids=(value)
     user_ids = value.map(&:to_i)
-
-    # Remove instructors of record that are not in the array of ids
-    course_memberships.select do |membership|
-      membership.instructor_of_record && !user_ids.include?(membership.user_id)
-    end.each do |membership|
-      membership.instructor_of_record = false
-      membership.save
-    end
-
-    # Add instructors of record that are in the array of ids
-    course_memberships.select do |membership|
-      !membership.instructor_of_record && user_ids.include?(membership.user_id)
-    end.each do |membership|
-      membership.instructor_of_record = true
-      membership.save
-    end
+    InstructorOfRecord.new(self).update_course_memberships(user_ids)
   end
 
   # Staff returns all professors and GSI for the course.
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -23,7 +23,7 @@ class Course < ActiveRecord::Base
   end
 
   def instructors_of_record
-    InstructorOfRecord.new(self).users
+    InstructorOfRecord.for(self).users
   end
 
   def instructors_of_record_ids
@@ -32,7 +32,7 @@ class Course < ActiveRecord::Base
 
   def instructors_of_record_ids=(value)
     user_ids = value.map(&:to_i)
-    InstructorOfRecord.new(self).update_course_memberships(user_ids)
+    InstructorOfRecord.for(self).update_course_memberships(user_ids)
   end
 
   def students_being_graded
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -23,7 +23,7 @@ class Course < ActiveRecord::Base
   end
 
   def instructors_of_record
-    InstructorOfRecord.for(self).users
+    InstructorsOfRecord.for(self).users
   end
 
   def instructors_of_record_ids
@@ -32,7 +32,7 @@ class Course < ActiveRecord::Base
 
   def instructors_of_record_ids=(value)
     user_ids = value.map(&:to_i)
-    InstructorOfRecord.for(self).update_course_memberships(user_ids)
+    InstructorsOfRecord.for(self).update_course_memberships(user_ids)
   end
 
   def students_being_graded
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -15,8 +15,15 @@ class Course < ActiveRecord::Base
     end
   end
 
+  # Staff returns all professors and GSI for the course.
+  # Note that this is different from is_staff? which currently
+  # includes Admin users
+  def staff
+    User.with_role_in_course("staff", self)
+  end
+
   def instructors_of_record
-    User.instructors_of_record(self)
+    InstructorOfRecord.new(self).users
   end
 
   def instructors_of_record_ids
@@ -26,13 +33,6 @@ class Course < ActiveRecord::Base
   def instructors_of_record_ids=(value)
     user_ids = value.map(&:to_i)
     InstructorOfRecord.new(self).update_course_memberships(user_ids)
-  end
-
-  # Staff returns all professors and GSI for the course.
-  # Note that this is different from is_staff? which currently
-  # includes Admin users
-  def staff
-    User.with_role_in_course("staff", self)
   end
 
   def students_being_graded
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -43,14 +43,6 @@ class Course < ActiveRecord::Base
     User.students_being_graded(self,team)
   end
 
-  def students_auditing
-    User.students_auditing(self)
-  end
-
-  def students_auditing_by_team(team)
-    User.students_auditing(self,team)
-  end
-
   def students_by_team(team)
     User.students_by_team(self, team)
   end
--- a/app/models/instructor_of_record.rb
+++ b/app/models/instructor_of_record.rb
@@ -29,7 +29,7 @@ class InstructorOfRecord
   end
 
   def not_instructors_of_record(user_ids)
-    course.course_memberships.select do |membership|
+    course.course_memberships.select(&:staff?).select do |membership|
       user_ids.include?(membership.user_id) && !membership.instructor_of_record?
     end
   end
--- a/app/models/instructor_of_record.rb
+++ b/app/models/instructor_of_record.rb
@@ -16,11 +16,15 @@ class InstructorOfRecord
     end
   end
 
+  def users
+    course.course_memberships.instructors_of_record.map(&:user)
+  end
+
   private
 
   def current_instructors_of_record(user_ids)
-    course.course_memberships.select do |membership|
-      !user_ids.include?(membership.user_id) && membership.instructor_of_record?
+    course.course_memberships.instructors_of_record.select do |membership|
+      !user_ids.include?(membership.user_id)
     end
   end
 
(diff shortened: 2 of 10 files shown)


app/models/instructor_of_record.rb
```diff
@@ -0,0 +1,36 @@
...
+  end
+
+  def update_course_memberships(user_ids)
+    current_instructors_of_record(user_ids).each do |membership|
+      membership.update_attributes instructor_of_record: false
+    end
+
+    not_instructors_of_record(user_ids).map do |membership|
+      membership.update_attributes instructor_of_record: true
+      membership
```
zazaian: @jwright is the goal here to return an array of updated formerly-non-instructor-of-record memberships and return those, or is this incidental? It seems like we should also return an array of the instructors of record that were removed instead of ignoring those and only adding the ones that were added. Maybe a hash with something like ```
{ removed: [ <these instructors removed>], added: [ <these instructors added> ] }``` would better summarize the outcome of that action.

app/models/instructor_of_record.rb
```diff
@@ -0,0 +1,36 @@
...
+  private
+
+  def current_instructors_of_record(user_ids)
+    course.course_memberships.instructors_of_record.select do |membership|
+      !user_ids.include?(membership.user_id)
+    end
+  end
+
+  def not_instructors_of_record(user_ids)
+    course.course_memberships.select do |membership|
```
zazaian: Shouldn't we only be querying for staff here? It seems like there wouldn't be any circumstances in which we have students being reassigned as instructors of record. @chcholman? Thoughts on this workflow? Not a big deal but probably less overhead to iterate over 5-15 staff instead of 50-500 students.

app/models/instructor_of_record.rb
```diff
@@ -0,0 +1,36 @@
...
+  end
+
+  def users
+    course.course_memberships.instructors_of_record.map(&:user)
+  end
+
+  private
+
+  def current_instructors_of_record(user_ids)
+    course.course_memberships.instructors_of_record.select do |membership|
```
zazaian: Same here, can we presume that we only need to look for non-student course memberships?

spec/models/instructor_of_record_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+  end
+
+  describe "#update_course_memberships" do
+    let(:membership) { create :staff_course_membership, course: course }
+
+    context "for new instructors of record" do
+      it "adds instructors of record to the course membership" do
+        memberships = InstructorOfRecord.new(course).update_course_memberships([membership.user_id])
+
+        expect(course.instructors_of_record).to eq [membership.user]
```
zazaian: This seems to violate the generate one-expectation-per-example guideline in Rspec. Can we split these out to isolate the expectations?

spec/models/instructor_of_record_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+      expect(subject.users).to eq [membership.user]
+    end
+  end
+
+  describe "#update_course_memberships" do
+    let(:membership) { create :staff_course_membership, course: course }
+
+    context "for new instructors of record" do
+      it "adds instructors of record to the course membership" do
+        memberships = InstructorOfRecord.new(course).update_course_memberships([membership.user_id])
```
zazaian: This should probably be defined in a let block so it can be shared between the example blocks.

spec/models/instructor_of_record_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+        expect(memberships.map(&:id)).to eq [membership.id]
+      end
+    end
+
+    context "for existing instructors of record" do
+      it "removes instructors of record to the course membership if they are not included" do
+        membership.update_attributes instructor_of_record: true
+        memberships = InstructorOfRecord.new(course).update_course_memberships([])
+
+        expect(course.instructors_of_record).to be_empty
```
zazaian: Same as above. It seems cleaner to isolate the expectation blocks according to convention and break the memberships local variable method out into a let block shared within that context.

app/models/instructor_of_record.rb
zazaian: Oops, sorry, I missed the instructors_of_record scope here. This one is fine :+1:

app/models/course.rb
```diff
@@ -15,8 +15,15 @@ class Course < ActiveRecord::Base
...
+  # Staff returns all professors and GSI for the course.
+  # Note that this is different from is_staff? which currently
+  # includes Admin users
+  def staff
+    User.with_role_in_course("staff", self)
+  end
+
   def instructors_of_record
-    User.instructors_of_record(self)
+    InstructorOfRecord.new(self).users
```
zazaian: This is a much better syntax, nice.

app/models/instructor_of_record.rb
```diff
@@ -0,0 +1,36 @@
+class InstructorOfRecord
```
zazaian: This feels more to me like it fits in /app/models than Timeline does, but it still feels like a new pattern. Let's be sure to discuss this in Tuesday chats. As you said /app/models is only autoloading whatever is in that directory, so it doesn't seem like there's a particularly compelling case not to autoload it from a /model_support directory or something to that effect to leave the AR models space to run free and roam in the pastures :sheep: :pig2: :goat:

jwright: :+1: for :pig_nose:

app/models/instructor_of_record.rb
jwright: @zazaian Good catch here. In actuality, we probably want to return all the `CourseMembership`s that are marked as instructors of record. In reality, we are not really using those but if we just return `course.course_memberships.instructors_of_record` and the client does not use it, that's ok because of lazy loading.

I will make the change

app/models/instructor_of_record.rb
jwright: @zazaian You are right. There is a validation to not allow the boolean set on student memberships. I will make the change.

app/models/course.rb
```diff
@@ -25,29 +32,7 @@ def instructors_of_record_ids
...
-      membership.save
-    end
-  end
-
-  # Staff returns all professors and GSI for the course.
-  # Note that this is different from is_staff? which currently
-  # includes Admin users
-  def staff
-    User.with_role_in_course("staff", self)
+    InstructorOfRecord.new(self).update_course_memberships(user_ids)
```
zazaian: This feels like a curious pattern to me. Why does it make implicit sense that a new `InstructorOfRecord` object (which isn't actually representing a single row in the instructor_of_records db table) would implicitly take a course? Isn't `InstructorOfRecord.new` kind of a misnomer in this context? Shouldn't this be something more like `InstructorsOfRecord.for(course).some_method`? or `InstructorsOfRecord.in_course(course)`. Does InstructorOfRecord really describe what this class is responsible for?

spec/models/instructor_of_record_spec.rb
jwright: :+1:

app/models/instructor_of_record.rb
zazaian: Cool, it makes sense that it would be all new instructors of record, let's just be mindful that that's what we need on the far end of the equation.

app/models/instructor_of_record.rb
zazaian: Cool.

spec/models/instructor_of_record_spec.rb
jwright: While I agree with the general idea of `1 expectation per spec`, I would make a caveat in saying this. If the expectations are expecting the same behavior or outcome I think it is fine, especially in instances where the test time would increase. What do you think about that? 

If you agree on that, then I feel like these are testing the same behavior here. Do you agree on that?

app/models/course.rb
zazaian: @jwright: I probably added this after you started looking at the other notes. Any thoughts on nomenclature? I think we can both agree that the Chinaman is not the issue. Sort of in the vain of what you and @acousticrobot seem to be getting at with the active-verb-nomenclature of the service patterns i wonder if we can find a way to describe this workflow that better encapsulates what's happening here? I liked the class name here when I looked at it initially but it feels to me like we might be able to find a better set of descriptions for the overall workflow.

spec/models/instructor_of_record_spec.rb
zazaian: Ah, I see what you're saying. I can get behind that, but let's at least be sure to add comments to demonstrate how the expectations are related. I would say, though, that running a second example here likely wouldn't add more than one or two hundredths of a second to the run.

Two expectations in a block seems fine to me, let's just be sure to explain the context here with comments if we can't rely on an example heading or context heading to make that distinction.

app/models/course.rb
jwright: @zazaian Haha. I actually started writing this as `InstructorsOfRecord.for` but I thought, nah, that's what `#new` is for.

I totally agree and I would be willing to rewrite this if everyone is cool with basically aliasing new. I think it does bring context value.

spec/models/instructor_of_record_spec.rb
jwright: Got it. I like it.

In addition, this may be over-testing since it is testing the same object. I actually think one of these can come out and this is probably a by-product of TDD.

zazaian: Okay cool, I kind of noticed that as well, though I suppose I could see the value in making sure that you get the same outcome both ways. Either path is fine with me.

app/models/course.rb
zazaian: :+1:

acousticrobot: I'm fine with that, guys.

jwright: @zazaian We aren't talking about the people who built the railroads.

zazaian: :cn: :mountain_railway: :+1:

PR #1654, opened 2016-02-13
1624 group graded assignments bug

### LINKED ISSUE #1624: Group graded assignments appearing in ungraded submissions
As an instructor, once I've graded an assignment (either group or individual) it should not appear on the ungraded assignment list. Currently group assignments are still marked as ungraded despite having been graded.

jwright:
Fixes a bug where a group graded assignment would still appear in the ungraded assignment list.

The assignment did not come off of the ungraded submissions list because the `#submission_id` was not being assigned for group grades.

This PR fixes that and a feature integration was modified to check that a submission was assigned to the grade.

Closes #1624

### DIFF
--- a/app/views/groups/_proposal_section.haml
+++ b/app/views/groups/_proposal_section.haml
@@ -10,7 +10,7 @@
         %fieldset.proposal.panel
           %span.columns
             %h6= "Created: #{proposal.created_at}"
-            %p= proposal.proposal
+            %p= raw proposal.proposal
             - i += 1
             = f.hidden_field :id
     - else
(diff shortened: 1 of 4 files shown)


app/views/groups/_proposal_section.haml
```diff
@@ -10,7 +10,7 @@
         %fieldset.proposal.panel
           %span.columns
             %h6= "Created: #{proposal.created_at}"
-            %p= proposal.proposal
+            %p= raw proposal.proposal
             - i += 1
```
acousticrobot: :+1: good catch

Issue #1658, opened 2016-02-15
Add AWS_S3_DEVELOPER_TAG to .env.sample


zazaian:
Need to add AWS_S3_DEVELOPER_TAG to the .env.sample in the repo.

PR #1659, opened 2016-02-15
Add AWS_S3_DEVELOPER_TAG to .env.sample with comments on its usage

zazaian:
This PR simply adds the AWS_S3_DEVELOPER_TAG to .env.sample and comments on its usage.

References #1658.


zazaian:
@acousticrobot: awesome! Glad those are helpful, thanks for the feedback :+1:

zazaian:
@chcholman this was signed off on by @acousticrobot and should be mergeable.

Issue #1662, opened 2016-02-15
Disable grade submission if no status selected


chcholman:
As an instructor, when I submit grades I need to make a conscious choice about what status they have (In Progress, Graded, or Released), but I need to choose one.

PR #1664, opened 2016-02-16
Grades without statuses

chcholman:
As a grader, sometimes I submit grades without statuses, and this then means that things land in a sort of limbo: that a submission may be marked as 'graded' but the student can't see their grade, and it hasn't been added to their score, AND the instructor no longer has a nudge that the object needs to be graded. 

This PR creates a new (hopefully temporary, dependent on #1662 ) section for these in-limbo grades to be identified on the grading status page. It also breaks the student syllabus table into two partials depending on role, removing numerous back and forth permission checks and allows them to depend on the two different parameters in play here: students can only see their grades if they're student-visible. Instructors can see and modify grades if they've ever been instructor-modified.


app/models/grade.rb
```diff
@@ -55,8 +55,9 @@ class Grade < ActiveRecord::Base
 
   scope :completion, -> { where(order: "assignments.due_at ASC", :joins => :assignment) }
   scope :graded, -> { where('status = ?', 'Graded') }
-  scope :in_progress, -> { where('status = ?', 'In Progress') }
+  scope :in_progress, -> { where status: "In Progress" }
   scope :released, -> { joins(:assignment).where("status = 'Released' OR (status = 'Graded' AND NOT assignments.release_necessary)") }
+  scope :in_limbo, -> { where("instructor_modified = true AND status = ''") }
```
jwright: style issue: I think we should stay away from where clauses stored in strings as much as we can. These tend to get lost and they are hard to catch changes to. In addition, I would chain the already existing `instructor_modified` scope to this one.

```
scope: :in_limbo, -> { instructor_modified.where(status: "") }
```

app/models/grade.rb
```diff
@@ -55,8 +55,9 @@ class Grade < ActiveRecord::Base
...
   scope :graded_or_released, -> { where("status = 'Graded' OR status = 'Released'")}
   scope :not_released, -> { joins(:assignment).where("status = 'Graded' AND assignments.release_necessary")}
   scope :instructor_modified, -> { where('instructor_modified = ?', true) }
```
jwright: I know this is not part of your changes but this may be better as

```
scope :instructor_modified, -> { where(instructor_modified: true) }
```

app/models/grade.rb
```diff
@@ -55,8 +55,9 @@ class Grade < ActiveRecord::Base
 
   scope :completion, -> { where(order: "assignments.due_at ASC", :joins => :assignment) }
   scope :graded, -> { where('status = ?', 'Graded') }
-  scope :in_progress, -> { where('status = ?', 'In Progress') }
+  scope :in_progress, -> { where status: "In Progress" }
   scope :released, -> { joins(:assignment).where("status = 'Released' OR (status = 'Graded' AND NOT assignments.release_necessary)") }
+  scope :in_limbo, -> { where("instructor_modified = true AND status = ''") }
   scope :graded_or_released, -> { where("status = 'Graded' OR status = 'Released'")}
```
jwright: I know this is not part of your changes but this may be better as

```
scope :graded_or_released, -> { where(status: ['Graded', 'Released'])}
```

app/models/grade.rb
```diff
@@ -55,8 +55,9 @@ class Grade < ActiveRecord::Base
...
   scope :not_released, -> { joins(:assignment).where("status = 'Graded' AND assignments.release_necessary")}
```
jwright: I know this is not part of your changes but this may be better as

```
scope :not_released, -> { joins(:assignment).where(assignments: { release_necessary: true }).where(status: 'Graded')}
```

app/models/grade.rb
```diff
@@ -55,8 +55,9 @@ class Grade < ActiveRecord::Base
 
   scope :completion, -> { where(order: "assignments.due_at ASC", :joins => :assignment) }
   scope :graded, -> { where('status = ?', 'Graded') }
-  scope :in_progress, -> { where('status = ?', 'In Progress') }
+  scope :in_progress, -> { where status: "In Progress" }
```
jwright: :+1:

app/controllers/info_controller.rb
```diff
@@ -33,13 +33,15 @@ def awarded_badges
   # Displaying all ungraded, graded but unreleased, and in progress assignment submissions in the system
   def grading_status
     @title = "Grading Status"
-    grades = current_course.grades
-    unrealeased_grades = grades.not_released
+    grades = current_course.grades.instructor_modified
```
jwright: Are we sure we want to use only `instructor_modified` grades in all of the queries below?

app/models/grade.rb
chcholman: I didn't know you could do that!

app/controllers/info_controller.rb
```diff
@@ -33,13 +33,15 @@ def awarded_badges
...
+    unreleased_grades = grades.not_released
     in_progress_grades = grades.in_progress
+    in_limbo_grades = grades.in_limbo
     @ungraded_submissions = current_course.submissions.ungraded.includes(:assignment, :grade, :student, :group, :submission_files)
     @ungraded_submissions_by_assignment = @ungraded_submissions.group_by(&:assignment)
-    @unreleased_grades_by_assignment = unrealeased_grades.group_by(&:assignment)
+    @unreleased_grades_by_assignment = unreleased_grades.group_by(&:assignment)
     @in_progress_grades_by_assignment = in_progress_grades.group_by(&:assignment)
+    @in_limbo_grades_by_assignment = in_limbo_grades.group_by(&:assignment)
```
jwright: I know this may be a big change but I don't think `in_limbo` is a good description...or at least it needs some specification or what it is when it is used in the client. It may be better as chaining the scopes manually by saying `instructor_modified.no_status`.

app/models/user.rb
```diff
@@ -335,6 +335,12 @@ def grade_released_for_assignment?(assignment)
     end
   end
 
+  def grade_instructor_modified_for_assignment?(assignment)
```
jwright: Hmm. This does not seem like it should be on the `User`. We are looking up a grade for an assignment. This may not need to change now but this is another reason I feel we need a `GradeBook` object.

Since we are using this in a view, can we just retrieve the grade here (https://github.com/UM-USElab/gradecraft-development/pull/1664/files#diff-8cf6417a57ec8ae71c977ec2894e0449R13), can we just check `grade.instructor_modified?` instead of using the `grade_instructor_modified_for_assignment` variable in the view? That would eliminate the need for this method.

app/views/info/_in_limbo_grades_table.haml
```diff
@@ -0,0 +1,52 @@
...
+            %th{"data-dynatable-no-sort" => "true", :width => "120px" }
+              %button.button.select-all= "Check"
+              %button.button.select-none= "Uncheck"
+          %tbody
+            - grades.each do |g|
+              - student = g.student
+              - team = student.team_for_course(course)
+              %tr
+                - if assignment.is_individual?
+                  %td= link_to student.try(:first_name), student_path(student)
```
jwright: No need for `try` here as this view is already dependent on the fact that a student has to exist

app/views/info/_in_limbo_grades_table.haml
```diff
@@ -0,0 +1,52 @@
...
+                  %td= link_to student.try(:last_name), student_path(student)
+                  %td
+                    - if course.has_teams? && student.present? && team.present?
+                      = link_to team.try(:name), team
```
jwright: No need for `try` here as the view is already dependent on a team with `team.present?`

app/views/info/_in_limbo_grades_table.haml
```diff
@@ -0,0 +1,52 @@
...
+                  %td= points g.score
+                  %td= raw g.feedback
+                %td
+                  .right
+                    %ul.button-bar
+                      %li= link_to glyph(:eye) + "See Grade", assignment_grade_path(g.assignment, :student_id => g.student_id), :class => "button"
+                      %li= link_to glyph(:edit) + "Edit Grade", edit_assignment_grade_path(:assignment_id => assignment.id, :student_id => student.try(:id)), :class => "button"
```
jwright: Remove `try` here

app/views/students/syllabus/_assignments.haml
```diff
@@ -16,105 +16,7 @@
...
-                            %li= link_to glyph(:edit) + "Edit Grade", edit_assignment_grade_path(assignment.id, :student_id => @student.id), :class => 'button'
-                          - else
-                            %li= link_to glyph(:check) + "Grade", edit_assignment_grade_path(:assignment_id =>assignment.id, :student_id => @student.id), :class => 'button'
-                        - elsif assignment.has_groups?
-                          - if grade_released_for_assignment
-                            %li= link_to glyph(:edit) + "Edit Grade", group_grade_assignment_path(assignment, :group_id => group_for_assignment), :class => 'button'
-                          - else
-                            %li= link_to glyph(:check) + "Grade", group_grade_assignment_path(assignment, :group_id => group_for_assignment), :class => 'button'
\ No newline at end of file
+        - if current_user_is_staff?
```
jwright: :+1:

app/controllers/info_controller.rb
```diff
@@ -33,13 +33,15 @@ def awarded_badges
   # Displaying all ungraded, graded but unreleased, and in progress assignment submissions in the system
   def grading_status
```
jwright: There is a definite need here for a `Presenter`.

app/views/students/syllabus/_instructor_table.haml
```diff
@@ -0,0 +1,97 @@
...
+    %th
+    %th= "#{term_for :assignment}"
+    %th Points
+    %th Due
+    %th
+    %th
+%tbody
+- assignments = assignment_type.assignments.includes(:assignment_type, :unlock_conditions)
+- assignments.each do |assignment|
+  - grade_instructor_modified_for_assignment = @student.grade_instructor_modified_for_assignment?(assignment)
```
jwright: There is a definite need here for a presenter. Declaring variables in views should be avoided.

app/views/students/syllabus/_instructor_table.haml
```diff
@@ -0,0 +1,97 @@
...
+  - group = @student.group_for_assignment(assignment)
+  - submission = @student.submission_for_assignment(assignment)
+  %tr
+    %td= render "students/syllabus/assignment_icons", :assignment => assignment
+    %td
+      = link_to assignment.name, assignment_path(assignment)
+      - if ! assignment.name_visible_for_student?(@student)
+        = glyph("eye-slash")
+    %td
+      - if assignment.points_visible_for_student?(@student)
```
jwright: Another case for a presenter here as this if statement is pretty deep.

app/views/students/syllabus/_student_table.haml
```diff
@@ -0,0 +1,85 @@
...
+    %th
+    %th= "#{term_for :assignment}"
+    %th Points
+    %th Due
+    %th
+    %th
+%tbody
+- assignments = assignment_type.assignments.includes(:assignment_type, :unlock_conditions)
+- assignments.each do |assignment|
+  - grade_visible_for_assignment = @student.grade_released_for_assignment?(assignment)
```
jwright: There is a definite need here for a presenter. Declaring variables in views should be avoided.

app/models/grade.rb
chcholman: Much better. Fixed.

app/models/grade.rb
chcholman: Fixed.

app/models/grade.rb
chcholman: Fixed.

app/controllers/info_controller.rb
chcholman: This is how it should work. I will go through again and ensure that we're adding instructor modified as necessary.

app/controllers/info_controller.rb
chcholman: Much better. Fixed.

app/views/info/_in_limbo_grades_table.haml
chcholman: Fixed.

app/views/info/_in_limbo_grades_table.haml
chcholman: Fixed.

app/controllers/info_controller.rb
```diff
@@ -33,13 +33,15 @@ def awarded_badges
...
     @title = "Grading Status"
-    grades = current_course.grades
-    unrealeased_grades = grades.not_released
+    grades = current_course.grades.instructor_modified
+    unreleased_grades = grades.not_released
     in_progress_grades = grades.in_progress
+    no_status_grades = grades.no_status
```
jwright: :+1:

app/models/grade.rb
```diff
@@ -15,10 +15,10 @@ class Grade < ActiveRecord::Base
     :earned_badges_attributes, :feedback_read, :feedback_read_at,
     :feedback_reviewed, :feedback_reviewed_at, :is_custom_value, :graded_at
 
-  STATUSES= ["In Progress", "Graded", "Released"]
+  STATUSES = ['In Progress', 'Graded', 'Released']
```
jwright: This is something we should chat about for the styles. I have been leaning towards always using double quotes for everything as a style guide. This is because there is no perceived performance increase for using single quotes and double quotes are needed for string interpolation. For these reasons, I think an appropriate style is to always use double quotes.

I am not saying we should change this here but just something we need to discuss for the style guide.

app/models/grade.rb
```diff
@@ -54,12 +54,13 @@ class Grade < ActiveRecord::Base
   after_destroy :save_student_and_team_scores
 
   scope :completion, -> { where(order: "assignments.due_at ASC", :joins => :assignment) }
```
jwright: So much win here :100:

app/models/user.rb
jwright: What are your thoughts on this method? Do we need to keep it?

jwright:
@cait I just have one more question left :)

app/models/user.rb
chcholman: Wow, yeah, that was a much, much better approach. Fixed. Thank you!

PR #1684, opened 2016-02-22
Issue 1660: LoginEventLogger and EventLogger Extractions

zazaian:
This pull request implements the following changes, and has been tested both through a mongo console and a redis console to confirm queuing and analytics event creation behaviors:

* Adds `LoginEventLogger` to the `ApplicationController` for logging course login and user login behaviors. This PR also moves the definition of event-relevant attributes into the respective EventLogger classes rather than defining them in the controller. Instead, and `event_session` method is defined and passed into EventLogger classes to provide common context for events.

* Similarly cleans up the EventLogger implementation in the `AnalyticsEventsController` by extracting event attributes and implementing an `event_session_with_params` method for passing relevant event data into `PredictorEventLogger` and `PageviewEventLogger`. Add fallback for each of these so that the event is directly inserted into Mongo in the event that Resque is unavailable.

* Adds an `ApplicationEventLogger` for storing common application `base_attrs` that were being used by all EventLogger descendants but were being defined independently in various places across the app. Also implements a `params` method that handles attribute formatting for params passed in from controller actions.

* Adds a `LoginEventLogger` class for logging course login and user login behaviors.

* Extracts enqueuing behaviors from `PageviewEventLogger` and moves them into common EventLogger::Enqueue library in order to share them with other EventLogger classes. Adds logic for building `page` attribute inside of `PageviewEventLogger`.

* Adds methods for building attributes from `event_session_with_params` inside of `PredictorEventLogger`, adds EventLogger::Enqueue behaviors to `PredictorEventLogger` to add Mongo fallback for predictor events.

* Adds `Resque` job throttling for `login_event_logger` and `predictor_event_logger` queues.

* Moves `EventLogger` to `EventLogger::Base`. Cleans up #perform method.

* Adds `EventLogger::Enqueue` module for handling enqueuing behaviors for EventLogger instances. Move fallback from controllers into `#with_fallback` methods in `EventLogger::Enqueue`.

* Adds an `EventLogger::Params` module for dynamically defining filtering methods for values that need to be taken from controller params and eventually logged to an Analytics::Event class.

* Extracts @ivar subclass inheritance behaviors from EventLogger::Base and ResqueJob::Base into `InheritableIvars` module, and re-implement in those classes with the module.

* Extracts `Logglier` logging behaviors from EventLogger::Base and ResqueJob::Base into `LogglyResque` module, and re-implement in those classes with the module. Move the definition of the `Logglier` tag into `/config/initializers/loggly.rb` and re-define inside of `LOGGLY_CONFIG` constant.

* Moves the `Lull` module out of the `EventHelpers` and into `/lib`. Re-implements methods on `Lull` as class methods rather than instance methods intended for inheritance. Replace instances of `Lull` inclusion with direct usage of `Lull` in `ApplicationController` and `AnalyticsEventsController`.

* Adds a spec for the `AnalyticsEventsController` (there wasn't one previously).

* Refactor the spec for event logging behaviors in `ApplicationController` after moving most of the event logging behaviors back into the EventLogger module.

* Adds a simple spec for the `UserSessionsController` (there wasn't one previously).

* Adds specs for all `EventLogger` subclasses.

* Updates or adds specs for all `EventLogger` vendor library classes and modules, and extracts common examples into SharedExample toolkits, moves context into `Toolkit` definitions, and defines test classes in `/spec/support/test_classes` rather than in the spec itself.

* Removes `rails_spec_helper` and `active_record_spec_helper` where possible in the EventLogger workflow and replace with explicit requirement and inclusion of all necessary test support files.


zazaian:
FYI I'm still going through to address formatting issues that are coming up in Code Climate.

chcholman:
@zazaian Thanks Mike. Don't worry about the quotations ones - I've got a PR that we'll merge after this that will standardize everything on double quotes.

zazaian:
Cool cool, sounds good @chcholman.

app/controllers/analytics_events_controller.rb
```diff
@@ -3,34 +3,25 @@ class AnalyticsEventsController < ApplicationController
...
-                                score: params[:score].to_i,
-                                possible: params[:possible].to_i,
-                                created_at: Time.now
-                                )  
+    PredictorEventLogger.new(event_session_with_params).enqueue_with_fallback
+
     render :nothing => true, :status => :ok
   end
 
   def tab_select_event
```
jwright: I don't see where this is being called from the front end anywhere. I only see it defined in the routes and used in tests. Is this created dynamically somewhere?

app/controllers/analytics_events_controller.rb
```diff
@@ -3,34 +3,25 @@ class AnalyticsEventsController < ApplicationController
 
   def predictor_event
```
jwright: :+1:  Much easier to read and understand what is going on here!

app/controllers/application_controller.rb
```diff
@@ -118,48 +116,27 @@ def enforce_view_permission(resource)
...
 
-  require_relative '../event_loggers/pageview_event_logger'
-  module ResqueManager
-    extend EventsHelper::Lull
-  end
-
-  # TODO: add specs for enqueing
   # Tracking page view counts
   def increment_page_views
     if current_user and request.format.html?
```
jwright: :+1:

app/controllers/analytics_events_controller.rb
zazaian: Actually I'm not sure, I couldn't find this either. It looks like this was added by @JangoSteve way back in the day. @chcholman, @acousticrobot: any idea whether the `tab_select_event` action in `AnalyticsEventsController` is being used anywhere?

chcholman: @jwright @zazaian That version of it is now defunct (given that we removed Foundation) and needs to be fixed using the new tab system we have. I'll add a ticket.

zazaian: Awesome @chcholman, thanks :+1:

app/event_loggers/application_event_logger.rb
```diff
@@ -0,0 +1,29 @@
+class ApplicationEventLogger < EventLogger::Base
+  @queue = :application_event_logger
```
jwright: I am not a big fan of exposing instance variables in base classes and then manipulating them. I think it can lead to unintended consequences and allow super classes to do things "willy-nilly" (technical term).

I think a better approach may be creating class level attr accessors with proper scope. What do you think about this?

```
module EventLogger
  class Base
    @queue = :event_logger

    class << self
      protected

      attr_accessor :queue
    end
  end
end

class ApplicationEventLogger < EventLogger::Base
  self.queue = :application_event_logger
end
```

Thoughts?

app/event_loggers/application_event_logger.rb
```diff
@@ -0,0 +1,29 @@
...
+  # Used by enqueuing methods in EventLogger::Enqueue
+  def event_type
+    "application"
+  end
+
+  # this is the default attribute set for EventLogger classes
+  # should be extended in #attrs inside of child classes for better
+  # granularity when more specific attributes are needed
+  def base_attrs
+    @base_attrs ||= {
```
jwright: Since `@base_attrs` are cached, then it will not pick up any changes to the `event_session` hash. I don't think we should cache `@base_attrs` here because order of operations becomes important then.

app/event_loggers/login_event_logger.rb
```diff
@@ -0,0 +1,41 @@
...
+  # Used by enqueuing methods in EventLogger::Enqueue
+  def event_type
+    "login"
+  end
+
+  ## class methods, for use when being called directly by Resque
+
+  # perform block that is ultimately called by Resque
+  def self.perform(event_type, data={})
+    @data = data
```
jwright: Another class level instance variable I think we should not manipulate without accessors.

app/event_loggers/login_event_logger.rb
```diff
@@ -0,0 +1,41 @@
...
+    data.merge!(last_login_at: previous_last_login_at)
```
jwright: Do we want to update `@data` here?

app/event_loggers/login_event_logger.rb
```diff
@@ -0,0 +1,41 @@
...
+    super
+    update_last_login
```
jwright: Hmmm.....this logger has an unintended consequence of updating the `last_login` for a `CourseMembership` when it's purpose is to log an event. This seems like it doesn't belong here but I don't know what the solution is. Perhaps an explicit callback or something may be different. I am unsure.

app/event_loggers/application_event_logger.rb
jwright: In addition, should the `@queue` instance variable only be available in the `EventLogger::Enqueue` module?

app/event_loggers/pageview_event_logger.rb
```diff
@@ -1,25 +1,31 @@
...
 
-  # message that posts to the log when being queued
-  @start_message = "Starting PageviewEventLogger"
+  # queue to use for login event jobs
+  @queue = :pageview_event_logger
+  @event_name = "Pageview"
 
-  def initialize(attrs={})
-    @attrs = attrs
+  attr_writer :page
```
jwright: I think a better name for this is `path`. `page` seems like it is caching the page details (html, etc).

app/event_loggers/pageview_event_logger.rb
```diff
@@ -1,25 +1,31 @@
...
-  def enqueue_in(time_until_start)
-    Resque.enqueue_in(time_until_start, self.class, 'pageview', @attrs)
+  def event_attrs
+    @event_attrs ||= base_attrs.merge page: page
   end
 
-  def enqueue_at(scheduled_time)
-    Resque.enqueue_at(scheduled_time, self.class, 'pageview', @attrs)
+  def page
+    @page ||= event_session[:request].try(:original_fullpath)
```
jwright: Can we move this initialization into the constructor?

app/event_loggers/predictor_event_logger.rb
```diff
@@ -1,22 +1,29 @@
...
-    Resque.enqueue_in(time_until_start, self.class, 'predictor', @attrs)
+  def event_type
+    "predictor"
   end
 
-  def enqueue_at(scheduled_time)
-    Resque.enqueue_at(scheduled_time, self.class, 'predictor', @attrs)
+  # params method is defined in ApplicationEventLogger
+  def event_attrs
+    @event_attrs ||= params ? base_attrs.merge(param_attrs) : base_attrs
```
jwright: Why does this initializer care about params in this case?

config/initializers/event_logger.rb
```diff
@@ -0,0 +1,6 @@
+EventLogger.configure do |config|
+  config.backoff_strategy = [
```
jwright: This method name does not seem to convey the data. How about `backoff_retry_intervals`?

config/initializers/resque_job.rb
```diff
@@ -0,0 +1,3 @@
+ResqueJob.configure do |config|
+  config.backoff_strategy = EventLogger.configuration.backoff_strategy
```
jwright: Same comment with the name here as well

lib/event_logger/base.rb
```diff
@@ -0,0 +1,66 @@
...
+
+    def self.failure_message_with_data(data)
+      "#{@failure_message} with data #{data}"
+    end
+
+    def self.analytics_attrs(event_type, data={})
+      { event_type: event_type, created_at: Time.zone.now }.merge data
+    end
+
+    def self.inheritable_ivars
```
jwright: Along the same lines of passing instance variables, I think this should be refactored to class attr accessors

lib/event_logger/enqueue.rb
```diff
@@ -0,0 +1,60 @@
...
+
+    def enqueue_at(scheduled_time)
+      Resque.enqueue_at(scheduled_time, logger_class, event_type, event_attrs)
+    end
+
+    def enqueue
+      Resque.enqueue(logger_class, event_type, event_attrs)
+    end
+
+    def enqueue_with_fallback
```
jwright: Can we have one method `enqueue_with_fallback` that has a default `time_until_start` of nil? I think the subtle differences in these names are confusing.

I wish Ruby had pattern matching :)

lib/event_logger/enqueue.rb
```diff
@@ -0,0 +1,60 @@
...
+      begin
+        # schedule the event in the background if Resque is available
+        enqueue
+      rescue
```
jwright: Should we just rescue `Resque` errors here?

lib/event_logger/enqueue.rb
```diff
@@ -0,0 +1,60 @@
...
+    end
+
+    # this is the default attribute set for EventLogger classes
+    # should be extended in #event_attrs inside of child classes for better
+    # granularity when more specific attributes are needed.
+    #
+    # Ideally this will be reorganized data from #event_session into a hash that's
+    # persistable in Mongo or whatever the back-end event store is
+    def base_attrs
+      @base_attrs ||= { created_at: Time.zone.now }
```
jwright: Why do we want to use server timezone specific time here? I believe we should store everything in UTC and display to the user's timezone where needed, or if we don't have the user's timezone, then display in server timezone. This eliminates the need of conversion to/from the server.

lib/event_logger/enqueue.rb
```diff
@@ -0,0 +1,60 @@
...
+    end
+
+    # this is set in the event that the base attributes are all that's
+    # needed by the target class
+    def event_attrs
```
jwright: Why do we need both `event_attrs` and `base_attrs`?

lib/event_logger/params.rb
```diff
@@ -0,0 +1,40 @@
...
+    # param_schema here should be in the format of :input_value => :output_value,
+    # where :input_value is the actual key in the params hash, and :output_value is
+    # the key that will be given in the output for EventLogging
+
+    # conditionally define methods for each parameter
+    def numerical_params(*param_schema)
+      param_schema.each do |param_format|
+
+        # if it's a symbol, just define a method with the same name as the input param
+        if param_format.is_a?(Symbol)
```
jwright: In order to remove the branching of logic here, would it be better to create classes that do this work? For example, you could create these classes:

```
class SymbolNumericalParam
end

class HashNumericalParam
end
```

and then call them in this method with:

```
  get_const("#{param_format.class.name}NumericalParam").new(param_format)
```

lib/inheritable_ivars.rb
```diff
@@ -0,0 +1,18 @@
+# this module adds some logic for automatically setting instance variables from
+# superclasses on their descendents. As with all such behaviors in Ruby, if those
+# instance variables are overwritten on the subclass then the superclass ivar
+# values will be ignored
+
+module InheritableIvars
```
jwright: I think this should be used to define class level getters/setters.

lib/is_configurable.rb
```diff
@@ -0,0 +1,16 @@
+# this should be extended in the target class/module
+module IsConfigurable
+  attr_writer :configuration
+
+  def configuration
+    @configuration ||= const_get("Configuration").new
```
jwright: Why the use of `const_get` here? Seems a bit magical and less performant when we can say `Configuration.new`.

lib/lull.rb
```diff
@@ -0,0 +1,45 @@
+# helpers for performing various tasks relative to the nightly low-usage downtime lull
+module Lull
+  class << self
+    def time_until_next_lull
+      next_lull_start - Time.zone.now
```
jwright: I think we should be using UTC here instead of server timezones.

lib/lull.rb
```diff
@@ -0,0 +1,45 @@
...
+      Time.zone.now > todays_lull_start and
+      Time.zone.now < todays_lull_end
+    end
+
+    def next_lull_start
+      is_after_todays_lull? ? tomorrows_lull_start : todays_lull_start
+    end
+
+    def tomorrows_lull_start
+      Date.tomorrow.to_time.change(lull_start_params)
```
jwright: `Date.tomorrow.to_time.utc`

spec/controllers/analytics_events_controller_spec.rb
```diff
@@ -1,11 +1,114 @@
 require 'rails_spec_helper'
+require 'resque-scheduler'
+require 'resque_spec/scheduler'
 
-RSpec.describe AnalyticsEventsController, type: :controller do
-  describe "POST predictor_event" do
-    skip "implement"
+describe AnalyticsEventsController, type: :controller do
+  extend Toolkits::EventLoggers::EventSession
```
jwright: I think we should use `World` here to keep it consistent with other spec code.

spec/controllers/analytics_events_controller_spec.rb
```diff
@@ -1,11 +1,114 @@
...
+      end
+
+      it "should render nothing" do
+        subject
+        expect(response.body).to be_empty
+      end
+
+      it "should render an :ok response" do
+        subject
+        expect(response.status).to eq(200)
```
jwright: This may be more readable (and consistent with other code): `expect(response).to be_ok`.

spec/controllers/analytics_events_controller_spec.rb
```diff
@@ -1,11 +1,114 @@
...
+    subject { post :predictor_event }
+
+    before(:each) { allow(logger_class).to receive_messages(new: event_logger) }
+
+    let(:logger_class) { PredictorEventLogger }
+
+    context "a user is logged in and the request is for html" do
+      it "should create a new PredictorEventLogger" do
+        expect(logger_class).to receive(:new).with(event_session_with_params) { event_logger }
+        subject
```
jwright: I think `subject` should always be the subject under test and not the outcome from an "act" step. I think this does not benefit the readability. I think `post :predictor_event` is easier to follow.

spec/controllers/analytics_events_controller_spec.rb
```diff
@@ -1,11 +1,114 @@
...
+      it "should render an :ok response" do
+        subject
+        expect(response.status).to eq(200)
+      end
+    end
   end
 
-  describe "POST tab_select_event" do
-    skip "implement"
+  describe "#event_session_with_params" do
```
jwright: I don't think we should be testing private/protected members.

jwright: Excuse me?

spec/controllers/application_controller/event_logging_spec.rb
```diff
@@ -0,0 +1,112 @@
+require 'rails_spec_helper'
+require 'resque-scheduler'
+require 'resque_spec/scheduler'
+
+# define routes for mapping #html_page and #json_page to the test controller
+include Toolkits::Controllers::ApplicationControllerToolkit::Routes
```
jwright: This is specific to this test so I think using a "toolkit" or one named as `ApplicationControllerToolkit` is misleading.

spec/controllers/application_controller/event_logging_spec.rb
```diff
@@ -0,0 +1,112 @@
...
+include Toolkits::Controllers::ApplicationControllerToolkit::SharedExamples
```
jwright: Since this is only used in this spec, I think we should not pull this out into a `SharedExamples` module and just include it here. This will help track down issues as it won't have a crazy call stack.

jwright: I see where this is being used in multiple locations now. Nevermind!

spec/controllers/application_controller/event_logging_spec.rb
```diff
@@ -0,0 +1,112 @@
...
+
+      it "should not call #{described_class}" do
+        allow(controller.request.format).to receive_messages(format)
+        expect(PageviewEventLogger).not_to receive(:new).with event_session
+        subject
+      end
+    end
+  end
+
+  after do
```
jwright: I think `after` blocks should be below `before` blocks in a spec. These are kind of hidden down here and someone adding a new spec may miss this.

spec/controllers/user_sessions_controller.rb
```diff
@@ -0,0 +1,24 @@
+require 'rails_spec_helper'
+
+describe UserSessionsController, type: :controller do
```
jwright: We don't really need `type: controller` in these since we have `infer_spec_type_from_file_location!` turned on.

spec/event_loggers/application_event_logger_spec.rb
```diff
@@ -0,0 +1,33 @@
+require 'active_record_spec_helper'
+require 'resque_spec/scheduler'
+
+require_relative '../toolkits/event_loggers/shared_examples'
```
jwright: This seems like a crazy amount of setup/required classes for one spec. Do we need all of this? If so, it is probably a sign of a complicated class or we are testing from the wrong place.

spec/event_loggers/login_event_logger_spec.rb
```diff
@@ -0,0 +1,143 @@
...
+
+  let(:logger_attrs) { login_logger_attrs } # pulled in from Toolkits::EventLoggers::Attributes
+  let(:expected_base_attrs) { application_logger_base_attrs } # pulled in from Toolkits::EventLoggers::ApplicationEventLoggerToolkit
+
+  # shared examples for EventLogger subclasses
+  it_behaves_like "an EventLogger subclass", LoginEventLogger, "login"
+  it_behaves_like "EventLogger::Enqueue is included", LoginEventLogger, "login"
+
+  describe "class methods" do
+    describe "self#peform" do
```
jwright: We should use class syntax here (`.perform`) for documenting purposes.

spec/event_loggers/login_event_logger_spec.rb
```diff
@@ -0,0 +1,143 @@
...
+      subject { class_instance.perform('login', logger_attrs) }
```
jwright: I think the subject under test should not be the "act" step here.

spec/event_loggers/login_event_logger_spec.rb
```diff
@@ -0,0 +1,143 @@
...
+
+      before(:each) { course_membership }
+
+      it "merges the previous last_login_at value into the data hash" do
+        allow(class_instance).to receive(:previous_last_login_at) { last_login.to_i }
+        expect(logger_attrs).to receive(:merge!).with({ last_login_at: last_login.to_i })
+        subject
+      end
+
+      it "sets the data hash to @data" do
```
jwright: We are testing internals to the class here. We can clean this up if we accessors.

jwright:
@zazaian Hey Mike, overall, I think this is a big improvement over what we had! I made several comments which I think are small overall.

A few notables though. I think the specs are getting way too fancy and they are very hard to follow. There are specs in a lot of different files. Since our tests do not have tests, I think we should try to limit things like shared examples, toolkits, etc. It's hard to know what is where. I did stop looking at the specs after a while because they were too hard to follow. Additionally, shared examples have notoriously hard to follow call stacks. A lot of time, this can be over-testing as just testing a module once may be enough.

We also need to fix the build in order to merge this in.

app/event_loggers/application_event_logger.rb
zazaian: I think this is a fine pattern. I was retaining this because it's a `Resque` convention but in subclasses it seems fine to write this instead as a simple accessor.

Is there any reason that you added the `protected` call here? I think in this context it would break accessibility of the queue attribute from Resque when the class is ultimately called by Resque itself.

Also, no, putting @queue in `EventLogger::Enqueue` would be erroneous here because the enqueue library is only modifying instance behavior, whereas Resque only cares about attributes on the class.

app/event_loggers/application_event_logger.rb
zazaian: The `event_session` hash should only be passed in once when the EventLogger::Base inheritor is built, and really shouldn't be modified. It's simply taking the `current_user`, `current_course`, `current_student`, and `request` from the controller and using those to build the logged parameters for the event. Can you fathom a reason why we'd need to change them? I'd argue that those really shouldn't change once they're set, which is why there's only an `attr_reader` here for them.

app/event_loggers/login_event_logger.rb
zazaian: @data isn't being updated here, it's being cached. Data is only being passed in as a method argument and doesn't have any other purpose in this context besides to store the original created_at date in order to use it later to update the course_membership with the proper last_login_at time.

I think it's fine to change this to an accessor, something like `:cached_data` would be okay, but again this isn't modifying the data but rather storing it to be called by self#update_last_login after the event is logged.

app/event_loggers/application_event_logger.rb
zazaian: Also please note that the `Resque` context these basically behave as constants, and will not be modified in any context except when they're set in any given class. Also please note that Resque reads the instance variable and not the attribute, so these should probably only be `attr_writer` if adding a setter here feels valuable.

app/event_loggers/login_event_logger.rb
zazaian: So this was the original behavior of the event logging implementation, but it could definitely be reworked. It makes sense that the EventLogger would only log the event and not also be responsible for updating the course_membership.

The problem, though, is that this behavior seems to be a very compelling reason to perform this action in a Resque job rather than in some helper class or what we've been referring to as a "service". @chcholman's concern is that in having 1,000 students login at the same time these numbers could spike, meaning a 1,000 extra times in which every user's `CourseMembership` has to be fetched and updated but in a foregrounded application context rather than the background.

I suppose one solution would be to break this out into two jobs, namely an `UpdateLastLogin` job and a `LoginEventLogger` job. In this context the former would perform the CourseMembership update in the background, and then trigger a new `LoginEventLogger` job to log the actual event.

Then again, it seems like the CourseMembership is going to have to be fetched on login anyway, so there's probably very little overhead in handling this in a service as long as we're careful about caching the `CourseMembership` from the context of the user authentication workflow.

Thoughts?

app/event_loggers/pageview_event_logger.rb
zazaian: `page` is what this value is currently called across the entire `Analytics` suite and all analytics aggregates. We could call it `path` but it feels kind of odd to call it `path` on the frontend and then `page` on the backend. It also feels unwieldy to modify the entire Analytics suite to reflect those changes, but I suppose I'm open to that.

Also this probably really isn't a `path` so much as a `url` or a `viewed_page_url`.

app/event_loggers/pageview_event_logger.rb
zazaian: Yup, I think that looks like a good move :+1:

app/event_loggers/predictor_event_logger.rb
zazaian: The `param_attrs` are all being filtered from the `params` hash that's being passed into the EventLogger's `event_session` by the Controller. If no params hash is present, trying to call the non-existent params here would raise and error.

config/initializers/event_logger.rb
zazaian: So this is actually just an attribute configuration for the `ExponentialBackoff` Resque plugin, which is used by `resque-retry`. We could name it differently here in the configuration, but ultimately it has to be named `backoff_strategy` on the Resque job class itself. Seems like it would be confusing to give it two different names.

zazaian: [https://github.com/lantins/resque-retry/blob/master/lib/resque/plugins/exponential_backoff.rb](url)

config/initializers/resque_job.rb
zazaian: See previous comment.

lib/event_logger/enqueue.rb
zazaian: So again, these are all `Resque::Scheduler` conventions and it seems like it would be confusing to invent a different interface for these behaviors. Enqueue, enqueue_at, and enqueue_in is consistent with the behaviors defined by `Resque::Scheduler`. We're simply adding fallback here to handle instances in which our complex server architecture causes faults enqueuing faults to Redis.

lib/event_logger/enqueue.rb
zazaian: Yup, good catch. These should actually be `redis` connection errors, but that's correct. I must have missed those when I ported them over from the controller.

lib/event_logger/enqueue.rb
zazaian: Yup, I think this makes sense :+1:

lib/event_logger/enqueue.rb
zazaian: The architecture here is such that `base_attrs` will be defined in an ApplicationEventLogger or some comparable parent class, and that `event_attrs` modify the application-specific base_attrs. I guess these could be called `application_attrs` if that helps to clarify things. But ultimately this is DRYing up the redundancy of build the same set of application event attrs that we were previously building in six or more different contexts.

lib/is_configurable.rb
zazaian: We actually can't use `Configuration` because it will be looking for `IsConfigurable::Configuration` instead of `ExtendedClass::Configuration`. As this is extending the configuration behaviors of three different behaviors we need to metaprogram the relationship to the `Configuration` constant and can't hard-code it.

lib/event_logger/params.rb
zazaian: Yeah, I think that's actually not a bad pattern. This is a lot more sophisticated than is likely warranted in this simple parameter filtering issue. Perhaps rather than `param_attrs` hardcoding the expected params we have a `filtered_params` attribute on the `EventLogger` that takes all of the filtered attributes, and then builds a final `param_attrs` hash based on what's been put there. Does that sound better?

lib/lull.rb
zazaian: Using UTC here is problematic since this library is only interacting with Resque/redis, and passing unzoned times into Redis will cause events to run not at server time but at UTC time.

lib/lull.rb
zazaian: Using utc here would break the `resque-scheduler` implementation.

spec/controllers/analytics_events_controller_spec.rb
zazaian: I think in this case because they're basic `Course`, `User` and `Student` objects that using `World` is definitely warranted. :+1:

spec/controllers/analytics_events_controller_spec.rb
zazaian: :+1: yup, that's fine

spec/controllers/analytics_events_controller_spec.rb
zazaian: I'm not really a fan of using `subject` only to define a larger subject under test for the entire spec. I think it's prone to a lot of redundancy and code that really isn't very DRY. I think it's a lot more readable to define the subject a given context block and then call that as `subject` rather than to call subject.some_action.some_other_behavior four or five times in the same block. It just seems like a bunch of dead un-DRY weight to me.

spec/controllers/application_controller/event_logging_spec.rb
zazaian: Thoughts on where these routes should be defined instead?

spec/controllers/analytics_events_controller_spec.rb
zazaian: Is there a compelling reason not to test this?

spec/controllers/user_sessions_controller.rb
zazaian: :+1:

zazaian:
So I think that overall the use of shared of examples here really gives us a lot for a little. Especially in instances where we have vendor libraries being included or extended in or from application classes we have in some cases four or five classes that have identical sets of behaviors that still might be situationally sensitive based on the context. All of the shared example specs are running in one or two hundredths of a second, so there's really very little overhead aside from having to looking shared examples for spec behavior rather than in a given spec.

I suppose we could strip them down and only test the modules, but I've seen a few instances in which the behaviors have differed based on the implementation, and they'll break in one shared example case but not in another. Again, this seems like really cheap insurance to me, and though it might feel more straightforward to just test the modules, I think we're leaving ourselves unnecessarily prone to regressions.

Perhaps moving the shared examples into `/spec/support/shared_examples` would create a less dense workflow than using the Toolkits for this? As I started extracting all of this functionality into vendor libraries I noticed that there was definitely a shared_example pattern emerging in order to avoid duplicate spec code and that might be a good pattern from now on.

Again, I suppose I'd be alright with moving the shared examples back just into the vendor library specs, but they really look an awful lot like the specs for testing the same included behaviors in the corresponding application code and feels redundant and un-DRY to me.

app/event_loggers/application_event_logger.rb
jwright: @zazaian Awesome. I placed protected here since I thought they were only modified by the super classes. I am unsure if Ruby allows this if they are superclass instance variables. I would have to play with it in the console.

I think `attr_writer` is fine.

app/event_loggers/application_event_logger.rb
jwright: I only suggest this because it forces an ordered set of operations which are not set by the logger interface. The client would have to know not to modify the `event_session` for any reason. In addition, it really does not offer much performance or memory benefits.

What about using `Hash#freeze` here instead of using an instance variable? I am not sure if that's an option.

app/event_loggers/application_event_logger.rb
zazaian: :+1: I think `attr_writer` sounds good here.

app/event_loggers/login_event_logger.rb
jwright: Gotcha. What about passing in `data` as a parameter instead of using a instance variable to pass around that is only used in this chain of methods?

app/event_loggers/login_event_logger.rb
jwright: Ooh. I like the idea of breaking these out into 2 different jobs unless you think that coordinating these into 2 jobs that need to be coordinated together is a large hassle (one fails, etc). A service would solve that problem with a rollback mechanism.

app/event_loggers/pageview_event_logger.rb
jwright: I am ok with the name if it's a large change. I was unaware of that.

app/event_loggers/login_event_logger.rb
zazaian: @jwright I think that makes sense, I'll look into reworking this workflow with that adjustment. :+1:

app/event_loggers/application_event_logger.rb
zazaian: Cool, I think that makes sense. `Hash#freeze` is probably a good compromise here. Thanks for the suggestion.

app/event_loggers/predictor_event_logger.rb
jwright: I see. It was a little confusing to see both `params` and `param_attrs` and know that those are linked in another part of the code.

What about just having `param_attrs` return an empty hash if `params` are nil? That way a merge with an empty hash would be a no op and this code can change to `@event_attrs` ||= base_attrs.merge(param_attrs`.

config/initializers/event_logger.rb
jwright: Yeah...I saw that later in the PR. I think this is fine since `Resque` uses it, even though I still think it's a poor name.

app/event_loggers/login_event_logger.rb
zazaian: @jwright I think breaking these out into two separate jobs would actually be a really stable way of handling this as long as they were responsible for only performing some behavior based on the data that is passed in at the time of enqueue rather than fetching the course_membership as we're currently doing and updating the login data based on what's stored in the record at the time the `CourseMembership` is being run. I'll look into this further and let you know what I come up with.

lib/event_logger/enqueue.rb
jwright: Got it. Makes sense. Thank you.

app/event_loggers/predictor_event_logger.rb
zazaian: That's actually how I'd written this initially but I believe I modified it. I think that's a good solution here :+1:

lib/event_logger/enqueue.rb
jwright: I like `event_attrs` and `base_attrs` (or `default_attrs`). However, it is confusing as to what changes when without having to look at a comment.

zazaian: What about just adding better commenting here for these?

zazaian: Ha, NVM: didn't see "without having to look at a comment". How about just changing `base_attrs` to `default_attrs`? Does that add clarity?

spec/controllers/application_controller/event_logging_spec.rb
zazaian: :+1: I think this is a fine change.

lib/is_configurable.rb
jwright: :+1:

spec/event_loggers/login_event_logger_spec.rb
zazaian: :+1: cool thanks, missed that

spec/controllers/analytics_events_controller_spec.rb
jwright: I disagree because that is not what the `subject` was designed for in RSpec. It is supposed to be the "receiver".

In addition, I feel DRY is not always important in specs, readability is more important. DRY sometimes requires fanciness and fanciness in specs (code without specs) can be dangerous.

I think using the method of `post :predictor_event` is no more drying than saying `subject` here.

Specs should follow Arrange, Act, Assert and when using `subject` as your act step, it can get really weird, especially when it is not consistent.

I am going to add this as a topic for our Tuesday chats and we can move on from it here.

spec/event_loggers/login_event_logger_spec.rb
zazaian: Okay, you and I seem to have kind of a fundamental disagreement about what constitutes a better workflow for using `subject`. What if we used a named subject in these cases with a name like `:outcome` or something similar that would allow us to condense the logic being tested in a given block without having to impede on the the `subject` namespace?

spec/controllers/analytics_events_controller_spec.rb
zazaian: In this context I agree that `post :predictor_event` is fine, but there are plenty of specs in the application in which method chains are called against a subject three or four times in the same context and that just feels needless to me.

spec/controllers/analytics_events_controller_spec.rb
jwright: It breaks encapsulation by testing members that are not supposed to be exposed to the outside world. It is an implementation detail and since it should be used by public methods, testing those public methods should be good enough. Over testing with private methods will mean a higher maintenance cost. It is akin in my mind as testing a model that is used in a controller from a controller test.

If I need to test a private method using TDD, I will test it and throw away the test or move it to a class if the method needs testing because of complexity.

spec/controllers/application_controller/event_logging_spec.rb
jwright: I think defining them right within this spec would be better so you can view what is defined (or the fact that this creates new routes).

spec/controllers/analytics_events_controller_spec.rb
zazaian: So what if in this context somebody changed the `params` call to a different hash, but the controller tests were still passing because they were technically receiving some kind of hash?

```
  def event_session_with_params
    event_session.merge { :some_other_params => "that could quietly be logging incorrect information in the EventLoggers }
  end
```

That to me feels like a regression that should be protected against. I see your point to a degree, but I'm having a hard time seeing how this is even in the same realm as testing model behavior through a controller.

lib/event_logger/enqueue.rb
jwright: I do think it helps but now I am confused as to why a superclass would need this to be exposed to them.

zazaian: So they can ultimately be extended in `event_attrs`. `event_attrs` contain the logic that's specific to the given `EventLogger` class that will ultimately generate the data being logged by the Analytics module.

spec/event_loggers/login_event_logger_spec.rb
jwright: I think that is an awesome idea.

FWIW, I usually use `result` as the result of "act" step. I am fine with `outcome` but we should be consistent with that in this project. I will add it to Tuesday Chats™

zazaian: I think `result` sounds good. If we can standardize this across this app I think we can help to add some conciseness to a lot of the specs.

zazaian:
@jwright thanks for the feedback here. I'll start working tomorrow on implementing the changes that we discussed and will bounce those back to you when I'm done.

spec/event_loggers/login_event_logger_spec.rb
zazaian: As noted these are actually separate entities and no internals are being tested here.

spec/event_loggers/application_event_logger_spec.rb
zazaian: If you look at the class itself it's actually quite simple, it just shares context with other event loggers here and thus uses behavior that's been extracted into shared context modules.

spec/controllers/application_controller/event_logging_spec.rb
zazaian: As noted this is being used in multiple places and putting the route definitions directly in the spec (they were before) would be redundant.

app/event_loggers/pageview_event_logger.rb
zazaian: I was thinking about this a bit more and I like the idea of calling this `viewed_page_url` on the frontend even if the name is page on the backend. It's a little awkward that the front end and back end names don't match but I think in this context because the Analytics suite is so dense and is really its own beast that it should be fine. I also think that `viewed_page_url` is much more expressive than 'path', and thus justifies the modification.

chcholman: @zazaian @jwright Can we delay making this name change at the moment if it's a large one?

zazaian: @chcholman sorry if I was unclear here. I'm suggesting that we change the name here only in the `PageviewEventLogger` and not in the `Analytics` suite. The former is a small change while it's latter change to `Analytics` that would be much more substantial.

chcholman: @zazaian Perfect :+1:

spec/controllers/application_controller/event_logging_spec.rb
zazaian: Actually on second thought I disagree here. I think it's confusing to place an `after` block before the examples that it will run after. If before and let blocks run in order it seems to stand to reason that after blocks should be placed after the examples in order to cognitively follow that they fall afterward.

Also with the other arrangement that happens before a set of examples between `subject`, `before` and `let` calls gumming up this workflow with `after` calls seems needless.

Additionally this methodology seems supported by style precedent in multiple rspec style guides.

app/event_loggers/application_event_logger.rb
zazaian: Hey @jwright: so I looked at this in practice and `freeze`-ing the hash doesn't really work, since the per-`event_attrs` for each EventLogger are being merged into it. I looked at just leaving this naked and un-cached, but as with our current example in the `ApplicationEventLogger` calling this multiple times for whatever reasons would cause multiple database hits instead of just caching the initial call and using that data.

I think perhaps we should probably just change the method name to `default_attrs` here and call it a day.

jwright: :dancers:  I am cool with that. Thanks for looking into it.

app/event_loggers/pageview_event_logger.rb
jwright: :+1:

lib/event_logger/enqueue.rb
jwright: Got it. Thanks.

PR #1685, opened 2016-02-22
Adds rubric grading functionality to group assignments

acousticrobot:
* extends the current functionality for grading with a rubric to assignments with groups
  * adds a helper #grade_with_rubric? to assignment model
  * refactor Angular controller for rubric grading
    * reduce the dependencies on outer modules
    * remove call to Angular controller init from haml
    * DRY up grade status logic ["In Progress", "Graded", etc.]
    * push all API calls into service
  * adds rubric params to views for grade show and edit pages.
    * _NOTE:_ we may need to add rubric details to some additional grade view pages, but summary grade and comments should work everywhere.
  * moves rubric grading submission from grades controller into api/criterion_grades
  * small updates to rubocop config
  * adds group to spec world helper


app/assets/javascripts/angular/services/RubricService.js.coffee
```diff
@@ -26,9 +67,28 @@
...
+        # We filter to the first student's grade since all students grades are identical (for now)
+        angular.copy(_.find(res.data, { attributes: {'student_id' : res.meta.student_ids[0] }}).attributes, grade)
+        angular.copy(res.meta.grade_status_options, gradeStatusOptions)
+      )
+
+  putRubricGradeSubmission = (assignment, params, returnURL)->
+    scopeRoute = if assignment.scope.type == "STUDENT" then "students" else "groups"
+    $http.put("/api/assignments/#{assignment.id}/#{scopeRoute}/#{assignment.scope.id}/criterion_grades", params).success(
+      (data)->
+        console.log(data)
```
jwright: We have a rouge `console.log` statement here.

app/assets/javascripts/angular/services/RubricService.js.coffee
```diff
@@ -26,9 +67,28 @@
...
+        window.location = returnURL
+    ).error(
+      (data)->
+        console.log(data)
```
jwright: Same here. I am not sure if we are just console logging errors?

app/assets/javascripts/angular/services/RubricService.js.coffee
```diff
@@ -26,9 +67,28 @@
...
+      )
+    else if assignment.scope.type == "GROUP"
+      $http.get('/api/assignments/' + assignment.id + '/groups/' + assignment.scope.id + '/grades/').success((res)->
+        # We filter to the first student's grade since all students grades are identical (for now)
+        angular.copy(_.find(res.data, { attributes: {'student_id' : res.meta.student_ids[0] }}).attributes, grade)
+        angular.copy(res.meta.grade_status_options, gradeStatusOptions)
+      )
+
+  putRubricGradeSubmission = (assignment, params, returnURL)->
+    scopeRoute = if assignment.scope.type == "STUDENT" then "students" else "groups"
```
jwright: It seems like this could be a utility function and used in lines [53](https://github.com/UM-USElab/gradecraft-development/pull/1685/files#diff-72a7f13ddb50e0fd2c984ee09da67808R53), [57](https://github.com/UM-USElab/gradecraft-development/pull/1685/files#diff-72a7f13ddb50e0fd2c984ee09da67808R57), [72](https://github.com/UM-USElab/gradecraft-development/pull/1685/files#diff-72a7f13ddb50e0fd2c984ee09da67808R72), [77](https://github.com/UM-USElab/gradecraft-development/pull/1685/files#diff-72a7f13ddb50e0fd2c984ee09da67808R77).

jwright: It seems like this could be a utility function and used in lines [53](https://github.com/UM-USElab/gradecraft-development/pull/1685/files#diff-72a7f13ddb50e0fd2c984ee09da67808R53), [57](https://github.com/UM-USElab/gradecraft-development/pull/1685/files#diff-72a7f13ddb50e0fd2c984ee09da67808R57), [72](https://github.com/UM-USElab/gradecraft-development/pull/1685/files#diff-72a7f13ddb50e0fd2c984ee09da67808R72), [77](https://github.com/UM-USElab/gradecraft-development/pull/1685/files#diff-72a7f13ddb50e0fd2c984ee09da67808R77).

app/controllers/api/criterion_grades_controller.rb
```diff
@@ -0,0 +1,51 @@
...
+      student_id: params[:student_id],
+      assignment_id: params[:id]
+    ).select(
+      :id, :student_id, :criterion_id, :level_id, :comments
+    )
+  end
+
+  def group_index
+    if !Assignment.find(params[:id]).has_groups?
+      render json: { message: "not a group assignment", success: false }, status: 400
```
jwright: Do we want to make this more "JSONAPI-like" or is that out of scope here?

It would change this to be something like:

```
render json: { errors: [{ detail: "not a group assignment" }], success: false }
```

http://jsonapi.org/format/#errors

app/controllers/api/grades_controller.rb
```diff
@@ -2,9 +2,26 @@ class API::GradesController < ApplicationController
 
   before_filter :ensure_staff?
 
-  # GET api/assignments/:id/student/:student_id/grade
+  # GET api/assignments/:id/students/:student_id/grade
   def show
     @grade = Grade.find_or_create(params[:id], params[:student_id])
+    @grade_status_options = @grade.assignment.release_necessary? ?
+      ["In Progress", "Graded", "Released"] : ["In Progress", "Graded"]
```
jwright: I feel like these should be in the `Grade` model and not in the controller as constants. We already have `Grade::STATUSES` that can be used. We should perhaps have `Grade::UNRELEASED_STATUSES`.

app/controllers/api/grades_controller.rb
```diff
@@ -2,9 +2,26 @@ class API::GradesController < ApplicationController
...
+    assignment = Assignment.find(params[:id])
+    if !assignment.has_groups?
+      render json: { message: "not a group assignment", success: false }, status: 400
+    else
+      students = Group.find(params[:group_id]).students
+      @student_ids = students.pluck(:id)
+      #TODO: Change this method globally to accept student ids
+      @grades = Grade.find_or_create_grades(params[:id], students)
+      @grade_status_options = assignment.release_necessary? ?
+        ["In Progress", "Graded", "Released"] : ["In Progress", "Graded"]
```
jwright: Same comment as above. I believe this is even more important now so we don't repeat these.

app/controllers/api/grades_controller.rb
```diff
@@ -2,9 +2,26 @@ class API::GradesController < ApplicationController
...
+
+  # GET api/assignments/:id/groups/:group_id/grades
+  def group_index
+    assignment = Assignment.find(params[:id])
+    if !assignment.has_groups?
+      render json: { message: "not a group assignment", success: false }, status: 400
+    else
+      students = Group.find(params[:group_id]).students
+      @student_ids = students.pluck(:id)
+      #TODO: Change this method globally to accept student ids
```
jwright: Doing this will really increase the efficiency here. Do we have an issue for this so it does not get lost somewhere?

app/controllers/grades_controller.rb
```diff
@@ -15,8 +13,8 @@ def show
     if current_user_is_student?
       redirect_to @assignment and return
     end
-    # TODO: we need to add rubrics for group assignment
-    if @assignment.rubric.present? && @assignment.is_individual?
+
+    if @assignment.grade_with_rubric?
```
jwright: :+1: Good use of "tell, don't ask". Nice fix!

app/views/api/criterion_grades/group_index.json.jbuilder
```diff
@@ -0,0 +1,12 @@
...
+  json.type "criterion_grades"
+  json.id criterion_grade.id.to_s
+
+  json.attributes do
+    json.merge! criterion_grade.attributes
+  end
+end
+
+json.meta do
+  json.student_ids @student_ids
```
jwright: Do we use these in the front-end? It looks like the front end is always grabbing the first one. Why is that?

app/views/grades/_group_standard_edit.html.haml
```diff
@@ -0,0 +1,28 @@
...
+    - if @assignment.has_levels?
+      = f.select :raw_score, @assignment_score_levels.map { |l| [l.formatted_name,l.value] }, :selected => @group.students.first.grade_for_assignment(@assignment).try(:raw_score), :include_blank => true
+    - else
+      = f.text_field :raw_score, :label => "Score", :hint => "Total Assignment Points: #{ points @assignment.point_total }", data: {autonumeric: true, "m-dec" => "0"}, :value => @group.students.first.grade_for_assignment(@assignment).try(:raw_score)
+
+  = f.text_area :feedback, :value => @group.students.first.grade_for_assignment(@assignment).try(:feedback), :class => "froala"
+
+  - if @assignment.release_necessary?
+    - if current_user_is_gsi?
+      = f.input :status, as: :select, :collection => ["In Progress", "Graded"], :selected => @group.students.first.grade_for_assignment(@assignment).try(:status)
```
jwright: Let's use a constant here as well instead of duplicating the statuses.

app/views/grades/_group_standard_edit.html.haml
```diff
@@ -0,0 +1,28 @@
...
+  = f.hidden_field :submission_id, value: @submission_id
+
+  Score:
+  = label_tag do
+    - if @assignment.has_levels?
+      = f.select :raw_score, @assignment_score_levels.map { |l| [l.formatted_name,l.value] }, :selected => @group.students.first.grade_for_assignment(@assignment).try(:raw_score), :include_blank => true
+    - else
+      = f.text_field :raw_score, :label => "Score", :hint => "Total Assignment Points: #{ points @assignment.point_total }", data: {autonumeric: true, "m-dec" => "0"}, :value => @group.students.first.grade_for_assignment(@assignment).try(:raw_score)
+
+  = f.text_area :feedback, :value => @group.students.first.grade_for_assignment(@assignment).try(:feedback), :class => "froala"
```
jwright: We should probably save off the `grade` for the view above and use it without duplicating the retrieval of it. Better yet, it should go into the controller if possible.

config/routes.rb
```diff
@@ -345,6 +344,11 @@
   namespace :api do
```
jwright: We may want to introduce some resourceful routing here a bit, especially for the students and groups. In addition, the `:id` params for `assignments` are not consistent and should be. We should use `:assignment_id` instead of `:id`.

```
resources :assignments, only: [] do
  resources :students, only: [] do
    resources :criterion_grades, only: :index
  end

  resources :groups, only: [] do
    resources :criterion_grades, only: :index
  end
end
```

spec/controllers/api/criterion_grades_controller_spec.rb
```diff
@@ -0,0 +1,133 @@
...
+      before(:each) do
+        world.create_group
+        world.assignment.groups << world.group
+        world.group.students.each do |student|
+          world.create_criterion_grade(assignment_id: world.assignment.id, student_id: student.id)
+        end
+      end
+
+      it "returns 400 error code with individual assignment" do
+        allow_any_instance_of(Assignment).to receive(:grade_scope).and_return("Individual")
```
jwright: I would use `world.assignment.update_attributes grade_scope: "Individual"` here instead of stubbing since `world.assignment` is a model.

spec/controllers/api/criterion_grades_controller_spec.rb
```diff
@@ -0,0 +1,133 @@
...
+        get :group_index, id: world.assignment.id, group_id: world.group.id, format: :json
+        expect(response.status).to be(400)
+      end
+
+      it "returns criterion_grades and student ids for a group" do
+        allow_any_instance_of(Assignment).to receive(:grade_scope).and_return("Group")
```
jwright: I would use `world.assignment.update_attributes grade_scope: "Group"` here instead of stubbing since `world.assignment` is a model.

spec/controllers/api/grades_controller_spec.rb
```diff
@@ -14,11 +13,45 @@
         expect(assigns(:grade).id).to eq(world.grade.id)
         expect(response).to render_template(:show)
       end
+
+      it "assigns all options when release necessary" do
+        allow_any_instance_of(Assignment).to receive(:release_necessary?).and_return(true)
```
jwright: I would use world.assignment.update_attributes release_necessary: true here instead of stubbing since world.assignment is a model.

spec/performers/controller_calls/rubric_worker_calls_spec.rb
```diff
@@ -0,0 +1,53 @@
+require 'rails_spec_helper'
+require "set"
+
+include ResqueJobSharedExamplesToolkit
+
+RSpec.describe API::CriterionGradesController, type: :controller, background_job: true do
```
jwright: We don't need the `type: :controller` here since we have `infer_spec_type_from_file_location!` turned on.

spec/performers/controller_calls/rubric_worker_calls_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+  include InQueueHelper
+  let(:course) { create(:course_accepting_groups) }
+  let(:assignment) { create(:assignment, course_id: course.id) }
+  let(:job_attributes) {{ grade_id: grade.id }} # for GradeUpdaterJob calls
+  let(:grade_attributes) {{ course_id: course.id, student_id: student.id, assignment_id: assignment.id }}
+  let(:grade) { create(:grade, grade_attributes) }
+  let(:cache_everything) { course; assignment; grade; student; professor }
```
jwright: I would use `let!` instead of doing this `cache_everything`

spec/performers/controller_calls/rubric_worker_calls_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+
+  before(:each) { ResqueSpec.reset! }
+
+  describe "triggering jobs as a professor" do
+    before(:each) { enroll_and_login_professor }
+
+    describe "#submit_rubric" do
+      skip "moved to api call, does this have to be run in the context of the GradesController?"
+      let(:request_attrs) {{ student_id: student.id, assignment_id: assignment.id, criterion_grades: [], grade: { status: "", feedback: "" }, format: :json }}
+      subject { put :update, request_attrs }
```
jwright: I think subjects should be the thing under test, not the "act" step.

spec/performers/controller_calls/rubric_worker_calls_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+
+      context "grade is not student visible" do
+        before { allow(grade).to receive(:is_student_visible?) { false } }
+
+        it_behaves_like "a failed resque job", GradeUpdaterJob
+      end
+    end
+  end
+
+  def enroll_and_login_professor
```
jwright: Should this be a helper method that other specs can use?

jwright:
@acousticrobot Overall I think this is very good. I love the small methods and great naming. It was very easy to follow. I just had a few comments.

spec/performers/controller_calls/rubric_worker_calls_spec.rb
acousticrobot: This file is that cut and paste of the `grade_controller#rubric_submission` spec we decided to ditch this morning on the other (find_or_create refactor) branch. Are you OK with me deleting the whole file from this branch? I think we decided it was more trouble than it was worth.

config/routes.rb
acousticrobot: Thank you! I wanted to do this, but didn't know how to do `only: []` to avoid all top-level resources.

app/views/grades/_group_standard_edit.html.haml
acousticrobot: This was moved verbatim from `group_edit.html.haml` so I might save refactoring for someone more familiar with the original code.

app/views/api/criterion_grades/group_index.json.jbuilder
acousticrobot: Current scope is to give all students an identical grade, but there is a feature request to be able to override individual grades for a group assignment. I wired up the back end to support the enhancement now, but on the front end only grab the first instance to populate the Angular form used for all grades.

app/controllers/api/grades_controller.rb
acousticrobot: Yup, updating all of GS's `find_or_create` methods is my other PR.

app/controllers/api/grades_controller.rb
acousticrobot: Good call, it's time to finish refactoring this grade status logic for good.

app/controllers/api/criterion_grades_controller.rb
acousticrobot: I like it.

app/assets/javascripts/angular/services/RubricService.js.coffee
acousticrobot: :+1:

app/controllers/api/grades_controller.rb
acousticrobot: Doing a quick search through the app, I think this problem runs beyond this PR. I've given it it's own ticket: https://github.com/UM-USElab/gradecraft-development/issues/1695

spec/controllers/api/criterion_grades_controller_spec.rb
acousticrobot: done

spec/controllers/api/criterion_grades_controller_spec.rb
acousticrobot: done

spec/controllers/api/grades_controller_spec.rb
acousticrobot: done

app/assets/javascripts/angular/services/RubricService.js.coffee
acousticrobot: I like it, but looking at the processing that has to happen to the resources it the get methods, I still have to break the rest of the GET methods into STUDENT or GROUP logic.  This would be good if we ever do a bigger refactor to the student logic to return [student] and then always iterate on individuals or groups.

config/routes.rb
acousticrobot: I'm changing these now to standardize on :id for assignment.
Adding the Resourceful routing gives us:
```
GET    /api/assignments/:assignment_id/students/:student_id/criterion_grades(.:format) api/criterion_grades#index {:format=>:json}
GET    /api/assignments/:assignment_id/groups/:group_id/criterion_grades(.:format)     api/criterion_grades#index {:format=>:json}
```
Does it make sense to update both resources from the same route, and introduce an `if (students)... else (groups)...` into the api/criterion_grades#index ?

app/controllers/api/criterion_grades_controller.rb
acousticrobot: done

app/assets/javascripts/angular/services/RubricService.js.coffee
jwright: I agree with that but it would remove the need for duplicate routes

app/controllers/api/grades_controller.rb
jwright: :+1:

app/controllers/api/grades_controller.rb
jwright: I saw that after I made this comment :cry:

app/views/api/criterion_grades/group_index.json.jbuilder
jwright: :crown:

config/routes.rb
jwright: No, I don't think so. I think splitting these up at the route level is the best thing to do right now. I think eventually moving this to a `GradesController` and using a `Gradable` concern for `Group` and `User` would be the best solution.

config/routes.rb
```diff
@@ -345,8 +344,12 @@
...
     get 'assignments/:id/criteria', to: 'criteria#index', defaults: { format: :json }
     get 'assignments/:id/students/:student_id/grade', to: 'grades#show', defaults: { format: :json }
+    get 'assignments/:id/groups/:group_id/grades', to: 'grades#group_index', defaults: { format: :json }
```
jwright: Unfortunately, Rails uses the resource name + "id" for nested resourceful routes, so I actually think we should reverse this to use `assignment_id` instead of `id`. It makes for finding resources "up the chain" in the url more explicit.

jwright:
I added more comment regarding the routes. I think it's good to go after that.

app/assets/javascripts/angular/services/RubricService.js.coffee
acousticrobot: It would work well for `.../criterion_grades` but unfortunately it's `.../students/:id/grade` and `.../group/:id/grades` so it would only work for one of the two GET route methods. :(

PR #1692, opened 2016-02-23
1312 add earned badge display page for students

ksofa2:
Adds a view to display full feedback on earned badges


ksofa2:
Will re-open from a branch in UM-USElab/gradecraft-development

PR #1694, opened 2016-02-23
refactor all find_or_create custom methods

acousticrobot:
Refactors the find_or_create methods to only take ids, not full objects, and to use native find_or_create_by when possible.  Adds missing specs for find_or_create methods. Checks for NullStudent in the badge and challenge predictor methods to avoid calling find_or_create with a null student.
Refactors the group_update method in the grades controller to use find_or_create, but still pass group_id onto the resulting grade.

    handle null student for predicted challenges
    add group id in group_update
    handle null students for predictor
    refactor criterion grades find or create
    refactor grade find in group update
    refactor find or create unlock state
    add group to world spec helper
    refactor grade find or create methods
    refactor find or create prediction models


app/controllers/badges_controller.rb
```diff
@@ -135,11 +135,15 @@ def predictor_badge_data
       :updated_at,
       :icon)
     badges.each do |badge|
-      prediction = badge.find_or_create_predicted_earned_badge(@student)
-      if current_user.is_student?(current_course)
-        badge.prediction = {id: prediction.id, times_earned: prediction.times_earned_including_actual}
+      if @student.class == NullStudent
```
jwright: Instead of having this `NullStudent` check here, what about this.

```
prediction = badge.find_or_create_predicted_earned_badge(@student.id)
if current_user.is_student?(current_course)
  badge.prediction = {id: prediction.id, times_earned: prediction.times_earned_including_actual}
else
  badge.prediction = {id: prediction.id, times_earned: prediction.actual_times_earned}
end
```

```
class Badge
  def find_or_create_predicted_earned_badge(student_id)
    student_id.nil? ? NullEarnedBadge.new : # find or create the predicted earned badge
  end
end
```

app/controllers/challenges_controller.rb
```diff
@@ -123,12 +123,15 @@ def predictor_challenge_data
       @grades = team.challenge_grades
 
       challenges.each do |challenge|
-        prediction = challenge.find_or_create_predicted_earned_challenge(@student)
-
-        if current_user.is_student?(current_course)
-          challenge.prediction = { id: prediction.id, points_earned: prediction.points_earned }
+        if @student.class == NullStudent
```
jwright: You could do the same thing here with a `NullEarnedChallenge`

jwright:
@acousticrobot Great implementation. I just had one suggestion that would remove a controller branching logic.

app/controllers/badges_controller.rb
acousticrobot: I was worried about letting the NullStudent object bleed out of the controller and into other models, but since you thought of it as well, it probably makes the most sense.

Issue #1691, opened 2016-02-23
Remove some-dir folder and image created by carrierwave test after completion


chcholman:
Right now whenever the test suite runs the 'some -dir' folder is created with an image inside of it. These should be cleaned up once the test has passed so they're not being committed back to the repo.

acousticrobot:
Also I would suggest that the directory should be built within the spec directory if possible, not public, so if it does get committed again it's not accessible in production.

PR #1697, opened 2016-02-25
Issue 1691: Fix Errant some-dir Directory Issue in /public

zazaian:
This PR is responsible for fixing a bug that was reported in issue #1691 as leaving a `some-dir` directory with some fixture image files in the `/public` directory of the app. It performs that change plus makes the following adjustments to the spec for `S3Manager::Carrierwave`:

* Replaces `rails_spec_helper` with `active_record_spec_helper` in order to retain the factory behaviors for `SubmissionFile`. Expressly requires the `S3Manager` library for testing rather than relying on rails to load it. The test starts up much faster now.

* Replace all instances of `s3_file_cylon` with `submission_file`, as the `S3FileCylon` nomenclature was left over from when there was a test class being used to test the attachment behaviors with that name.

* Fixes all `rubocop` issues, predominantly matters of reducing 80+ character lines to 80 or less.


spec/lib/s3_manager/carrierwave_spec.rb
```diff
@@ -163,40 +188,53 @@
...
+      end
+
+      it "caches the store_dir before create" do
+        expect(submission_file[:store_dir]).to eq(store_dir)
+      end
     end
 
-    it "caches the store_dir before create" do
-      expect(create_submission_file[:store_dir]).to eq("some-dir")
+    after(:each) do
```
zazaian: This is the block that is expressly removing the temp dirs plus files after running. `secure` is used to protect against vulnerabilities noted in the Ruby API.

spec/lib/s3_manager/carrierwave_spec.rb
```diff
@@ -163,40 +188,53 @@
     end
   end
 
-  describe "#cache_store_dir" do
-    subject { s3_file_cylon.instance_eval { cache_store_dir }}
-    before { allow(s3_file_cylon).to receive(:file) { double(:file, store_dir: "some-dir") }}
+  describe "caching the #store_dir" do
+    let(:store_dir) { Dir.mktmpdir }
```
acousticrobot: :+1:  Awesome -- This is a much better location for testing, especially in combination with `FileUtils.rm_rf`

spec/lib/s3_manager/carrierwave_spec.rb
acousticrobot: Looks great Mike, and will clean up nicely even if the spec fails down the road. :trophy:

spec/lib/s3_manager/carrierwave_spec.rb
zazaian: Cool, thanks! I think it was a combination of using Carrierwave `:fog` plus testing the `store_dir` here, since `fog` uses `/public/uploads`. This wouldn't have come up in testing since the change was made during the PR review phase on the release in which this came in. This completely circumvents `:fog` and doesn't actually rely on on carrierwave to generate that directory, which is fine since we're just testing caching of the directory path prior to the `SubmissionFile` or comparable `AttachmentUploader` model being created.

zazaian:
Awesome, @acousticrobot, sounds good. Thanks for taking the time to review this.

And yes indeed, rubocop seems to think there are around 29,000 formatting issues across the app right now, which really aren't dire, but if we're trying to play the universal uniformity game the only way to get there is one file at a time. :turtle:

PR #1700, opened 2016-02-25
1647 incorporate submission files

### LINKED ISSUE #1647: Incorporate Submission File history into Submission
As a user, when I'm looking at submission histories, Submission files and Submissions are currently displayed as separate items. Since from the user's perspective they're functionally the same, they should be displayed as a single node - so if I've created a new Submission with a file attachment, that would produce one "Submission Created" node rather two nodes ("Submission Created' and "Submission File Created").

jwright:
Merges and transforms the history for `Submission`s and `SubmissionFile`s so that they are not treated as separate events.

Instead of this (separate submission file):

screen shot 2016-02-25 at 11 27 51 am

It now displays the `SubmissionFile` changes with the `Submission`:

screen shot 2016-02-25 at 11 28 08 am

In addition, in order to create a "past tense" version of a string ("create" -> "created", "upload" -> "uploaded"), I created a custom string class that allows this. No 🐒patching.

Closes #1647

### DIFF
--- a/app/models/history_filter.rb
+++ b/app/models/history_filter.rb
@@ -55,6 +55,11 @@ class HistoryFilter
     self
   end
 
+  def transform(&blk)
+    history.map(&blk)
+    self
+  end
+
   def empty_changeset?(set)
     set.values.none? { |value| value.is_a? Array }
   end
--- a/app/models/history_filter.rb
+++ b/app/models/history_filter.rb
@@ -18,7 +18,8 @@ class HistoryFilter
       end
       result &= !yield(history_item) if block_given?
       result
-    end.delete_if { |history_item| empty_changeset?(history_item.changeset) }
+    end
+    clear_empty_changesets
     self
   end
 
@@ -31,7 +32,8 @@ class HistoryFilter
       end
       result &= yield(history_item) if block_given?
       result
-    end.delete_if { |history_item| empty_changeset?(history_item.changeset) }
+    end
+    clear_empty_changesets
     self
   end
 
@@ -40,7 +42,8 @@ class HistoryFilter
 
     @history = history.select do |history_item|
       history_item.changeset.delete_if { |key, value| name == key } if name
-    end.delete_if { |history_item| empty_changeset?(history_item.changeset) }
+    end
+    clear_empty_changesets
     self
   end
 
@@ -48,7 +51,7 @@ class HistoryFilter
     object_key = options.keys.first
     object_value = options.values.first
 
-    history.map(&:changeset).each do |changeset|
+    changesets.each do |changeset|
       object = changeset["object"]
       changeset["object"] = object_value if object == object_key
     end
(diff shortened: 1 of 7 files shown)


app/models/history_filter.rb
```diff
@@ -48,14 +72,26 @@ def rename(options)
...
+    self
+  end
+
   def empty_changeset?(set)
     set.values.none? { |value| value.is_a? Array }
   end
+
+  private
+
+  def clear_empty_changesets
```
acousticrobot: nice refactor here :small_red_triangle:

lib/string.rb
```diff
@@ -0,0 +1,40 @@
+module Gradecraft
```
acousticrobot: :star2: Awesome -- I like how you made this a `Gradecraft::String`, not just mysteriously overriding String  :do_not_litter:, but it still has all of the functionality of a regular String object.

This might be a good starting place for the `Sanitize` functionality Mike has been talking about.

spec/models/history_filter_spec.rb
```diff
@@ -81,6 +81,26 @@
     end
   end
 
+  describe "#merge" do
```
acousticrobot:
```
  describe "#merge" do
    it "merges the changeset from one object to another" do
      history = [OpenStruct.new(version: OpenStruct.new(item_type: "FromObjectType"),
                                changeset: { "event" => "from_object",
                                             "created_at" => "from_object",
                                             "updated_at" => "from_object",
                                             "object" => "FromType",
                                             "from_string" => "from_object",
                                             "from_array" => ["from_object"],
                                             "common_array" => ["from_object"]}),
                 OpenStruct.new(version: OpenStruct.new(item_type: "ToObjectType"),
                                changeset: { "event" => "to_object",
                                             "created_at" => "to_object",
                                             "updated_at" => "to_object",
                                             "object" => "ToType",
                                             "to_string" => "to_string",
                                             "to_array" => ["to_object"],
                                             "common_array" => ["to_object"]})
      ]
      result = described_class.new(history).merge("FromObjectType" => "ToObjectType")
        .changesets

      expect(result).to eq [{ "event" => "to_object",
                              "created_at" => "to_object",
                              "updated_at" => "to_object",
                              "object" => "ToType",
                              "to_string"=>"to_string",
                              "to_array" => ["to_object"],
                              "common_array" => ["from_object"],
                              "from_array"=>["from_object"]}]
      expect(result.first.has_key? "from_string").to be_falsey
    end
  end
```

acousticrobot: This looks from the screen examples like it works great in the context of Submissions and SubmissionFiles. 

My only question is about the intended behavior of `HistoryFilter#merge` as it might extend to other object merges.  I've pasted a update to the spec with more generic `item_type`s that I think serves to cover and document all the possible outcomes that I can see from the logic path. I also don't know without getting into the weeds of it why this would be the desired outcome. Perhaps a few comments as to why we take some common values from one object or another, and why some are not persisted?

jwright: @acousticrobot I updated the spec to make it more generic but it differs from your example. Your `"from_string"` would not happen in the real world and the other things have been handled in the example. Thoughts from the new changes?

PR #1713, opened 2016-02-29
Issue 1712: Logger double was spilling into subsequent spec example

zazaian:
This PR resolves the issue where a double was  being passed into two separate examples rather than generating a new one, which was throwing an error depending on the order in which the specs were being called.

To resolve this I replaced the doubles with `Event::Logger.new` objects and stubbed out the validity behaviors that were being tested for messaging in the describe block for the `.notify_event_outcome` block. This should be working properly now.

Additionally I removed the `self` prefixes from the describe blocks here, as using a `.` instead of `#` when describing methods implicitly suggests that the method being tested is a class method rather than an instance method.

### DIFF
--- a/spec/lib/event_logger/base_spec.rb
+++ b/spec/lib/event_logger/base_spec.rb
@@ -13,14 +13,17 @@ describe EventLogger::Base, type: :vendor_library do
   include Toolkits::Lib::ResqueRetryToolkit::SharedExamples
 
   let(:event_type) { "waffle" }
-  let(:some_data) {{ some: "new weird data" }}
+  let(:some_data) { { some: "new weird data" } }
   let(:time_now) { Time.parse "Jun 20 1942" }
   let(:mongoid_event) { double(:mongoid_event) }
-  let(:analytics_attrs) {{ event_type: event_type, created_at: time_now }.merge(some_data) }
+  let(:analytics_attrs) do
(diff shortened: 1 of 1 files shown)


zazaian:
@acousticrobot: fyi this requires your review when you get a moment. Thanks!

zazaian:
@acousticrobot: haha, no worries. Just wanted to make sure you saw it. Take your time, thanks :+1:

zazaian:
Hey there @acousticrobot, thanks for taking the time to test this out for me. I think you're probably right, we're working with class methods here so the behaviors are a bit strange when working with doubles. This was only failing for me every third or fourth time when running the whole suite, so it was hard to troubleshoot.

I've replaced the Logger double with a `Logger.new(STDOUT)` so it's just using a standard Ruby `Logger` and outputs to STDOUT instead of Loggly, which is fine since we're only testing the output here and not the destination. This seems to have stabilized the spec both running it focused, as a spec file, and for the whole suite.

Please take a look at this when you get an opportunity but I think this has fixed it. Thanks again for the assistance here, it's much appreciated.

zazaian:
@acousticrobot indeed I saw that the `Logger` is now actually printing that text to `STDOUT` in the terminal when the spec is run but it seemed fine to me as well. Thanks for taking a look at this and for the feedback.

@chcholman is ready for merge.

PR #1715, opened 2016-02-29
Adds a has submission icon to Predictor

acousticrobot:
This addresses a bug where we were not summing predicted points for any closed assignments,
even if we had enabled the slider. This happened because we were not passing information
about the student's submissions to the front end.

This adds a file icon in the predictor icon bar if the student has made a submission for
the assignment, and only discounts predicted points for closed assignments without
a student submission.

Additionally, this updates the sample data to more explicitly test all areas of concern
in the predictor, and to better describe what is being tested within the assignment title
and description. It also fixes a number of bugs which prohibited predicted data from
accurately representing model states found in the real world.

screen shot 2016-02-29 at 2 56 33 pm

### To Test:

  1. Run sample data
  2. Sign in as a student and check all assignments in the Predictor "Predictor Settings" section.
  3. Check that the :information_source:  description matches both the exisiting and desired state (visible icons, slider type / points display)
  4. Check that the prediction sliders influence the top graph in the expected manner.

### DIFF
--- a/app/assets/javascripts/angular/controllers/PredictorCtrl.js.coffee
+++ b/app/assets/javascripts/angular/controllers/PredictorCtrl.js.coffee
@@ -120,7 +120,7 @@
       # use raw score to keep weighting calculation on assignment type level
       if assignment.grade.raw_score != null
         total += assignment.grade.raw_score
-      else if ! assignment.pass_fail && !$scope.closed_for_prediction(assignment) && includePredicted
+      else if ! assignment.pass_fail && ! assignment.closed_without_sumbission && includePredicted
         total += assignment.grade.predicted_score
     )
     total
--- a/app/assets/javascripts/angular/controllers/PredictorCtrl.js.coffee
+++ b/app/assets/javascripts/angular/controllers/PredictorCtrl.js.coffee
@@ -120,7 +120,7 @@
       # use raw score to keep weighting calculation on assignment type level
       if assignment.grade.raw_score != null
         total += assignment.grade.raw_score
-      else if ! assignment.pass_fail && ! assignment.has_closed && includePredicted
+      else if ! assignment.pass_fail && !$scope.closed_for_prediction(assignment) && includePredicted
         total += assignment.grade.predicted_score
     )
     total
@@ -232,6 +232,14 @@
 
   $scope.badgeCompleted = (badge)->
     if (badge.point_total == badge.total_earned_points && ! badge.can_earn_multiple_times)
+      return true
+    else
+      return false
+
+  # We keep the predictor open on closed assignments IF the student has a
+  # a submission
+  $scope.closed_for_prediction = (assignment)->
+    if assignment.has_closed && !assignment.has_submission
       return true
     else
       return false
--- a/app/assets/javascripts/angular/templates/ng_predictor_assignments.html.haml
+++ b/app/assets/javascripts/angular/templates/ng_predictor_assignments.html.haml
@@ -37,7 +37,7 @@
       %div{'ng-if' => 'assignment.predictor_display_type == "checkbox" && assignment.pass_fail == true'}
         .predictor-binary-switch{'target'=>'assignment','target-type'=>'assignment','off-value'=>'0','on-value'=>'1'}
 
-      %div{'ng-if' => 'assignment.predictor_display_type == "checkbox"'}
+      %div{'ng-if' => 'assignment.predictor_display_type == "checkbox" && !assignment.pass_fail == true'}
         .predictor-binary-switch{'target'=>'assignment','target-type'=>'assignment','off-value'=>'0','on-value'=>'assignment.point_total'}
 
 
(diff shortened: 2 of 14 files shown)


chcholman:
@acousticrobot Awesome! The sample data looks exactly as described. One piece of language it'd be great to discuss tomorrow is around the conflicting icon statements that it both accepts submissions and no longer accepts submissions.

app/assets/javascripts/angular/controllers/PredictorCtrl.js.coffee
```diff
@@ -236,6 +236,14 @@
     else
       return false
 
+  # We keep the predictor open on closed assignments IF the student has a
+  # a submission
+  $scope.closed_for_prediction = (assignment)->
```
jwright: Do we want to use `is_closed_for_prediction` here for predicate methods? I am not sure of the guidelines but it just caught my eye since we use `is_` or `has_` for most predicate methods in js.

app/assets/javascripts/angular/templates/ng_predictor_assignments.html.haml
```diff
@@ -37,7 +37,7 @@
       %div{'ng-if' => 'assignment.predictor_display_type == "checkbox" && assignment.pass_fail == true'}
         .predictor-binary-switch{'target'=>'assignment','target-type'=>'assignment','off-value'=>'0','on-value'=>'1'}
 
-      %div{'ng-if' => 'assignment.predictor_display_type == "checkbox"'}
+      %div{'ng-if' => 'assignment.predictor_display_type == "checkbox" && !assignment.pass_fail == true'}
```
jwright: Do we want to use triple equals (`===`) to check for type or does angular handle this?

app/serializers/predicted_grade_serializer.rb
```diff
@@ -30,7 +30,8 @@ def score
     grade.score if grade.is_student_visible?
   end
 
-  def initialize(grade, current_user)
+  def initialize(assignment, grade, current_user)
```
jwright: Do we need this extra parameter here or are always going to use `grade.assignment`? If `assignment` and `grade.assignment` will be the same, I would vote for reducing the parameter complexity here.

spec/models/null_grade_spec.rb
```diff
@@ -64,4 +64,8 @@
   it "handles queries for assignments accepting submissions" do
     expect(subject.assignment.accepts_submissions?).to be_falsey
   end
+
+  it "returns a null student for student" do
```
jwright: :+1:

spec/serializers/predicted_assignment_serializer_spec.rb
```diff
@@ -77,7 +77,7 @@
       expect(exposed_attributes & subject.attributes.keys).to eq(exposed_attributes)
     end
 
-    describe "additional boolean flags" do
+    describe "#boolean flags" do
```
jwright: The `#` usually denotes an instance method that is being tested but `#boolean flags` is not a valid instance method name. Should this be `#boolean_flags`?

spec/serializers/predicted_assignment_serializer_spec.rb
```diff
@@ -162,15 +207,21 @@
         end
       end
 
-      describe "has_closed" do
+      describe "closed_without_submission" do
```
jwright: Should be changed to `#closed_without_submission?`?

jwright:
@acousticrobot This is great! I just had a few comments.

app/serializers/predicted_grade_serializer.rb
acousticrobot: The problem was, when student is an instance of `NullStudent`, `student.grade` was passed in as a `NullGrade`, and `NullGrade` did not have a reference to the actual assignment. `show_zero_in_predictor` was always false, which meant that closed assignments were shown as open in the predictor. Adding the assignment in the params allows the serializer to always use a real assignment, even when using a bogus grade.

spec/serializers/predicted_assignment_serializer_spec.rb
acousticrobot: :+1:

spec/serializers/predicted_assignment_serializer_spec.rb
acousticrobot: :+1:

app/assets/javascripts/angular/templates/ng_predictor_assignments.html.haml
acousticrobot: Not a bad thought, but in some of these ng-if conditionals I think we have to handle with our rails models sometimes using `true` and `nil` because they don't have a default boolean value in the schema. We'd have to do some thorough testing before changing them throughout the app.

app/assets/javascripts/angular/controllers/PredictorCtrl.js.coffee
acousticrobot: I think it would be a good standard to follow. I assigned the boolean attributes on the models predicate names in the ruby  serializer: `assignment.has_closed`, `.has_submission`...
 
In this Angular file, all the other boolean methods are without them: `$scope.assignmentTypeAtMax`, `$scope.articleGraded`, `$scope.challengeGraded`, `$scope.articleNoPoints`... so this probably should be done throughout the file and then we would have to find the methods as they are being called from the haml :sob:

PR #1719, opened 2016-03-01
1563 mobile course switch

### LINKED ISSUE #1563: Switching courses from mobile currently breaks

jwright:
Change route to change courses to be a get request. The switching was broke on mobile because the rails js was no longer handling a `POST` request for submitting the form.

Fixes #1563


app/assets/javascripts/courses.js
```diff
@@ -27,16 +27,7 @@
...
-            form.setAttribute("method", "post");
-            form.setAttribute("action", "/current_course/change");
-            var field = document.createElement("input");
-            field.setAttribute("type", "hidden");
-            field.setAttribute("name", "course_id");
-            field.setAttribute("value", id);
-            form.appendChild(field);
-            document.body.appendChild(form);
-            form.submit();
+            window.location = "/current_course/change?course_id=" + id
```
zazaian: Much, much, much better way of doing this :+1:

app/views/courses/index.html.haml
```diff
@@ -27,7 +27,7 @@
     %tbody
       - @courses.each do |course|
         %tr
-          %td= link_to course.name, change_current_course_path(course_id: course.id), data: { method: "post" }
+          %td= link_to course.name, change_current_course_path(course_id: course.id)
```
zazaian: :+1: good stuff

config/routes.rb
```diff
@@ -193,7 +193,7 @@
   end
   resources :course_memberships
 
-  post "/current_course/change" => "current_courses#change", :as => :change_current_course
+  get "/current_course/change" => "current_courses#change", :as => :change_current_course
```
zazaian: Shouldn't this actually be a `put` since we're updating the `current_course_id` attribute on the `user`?
```
    # /app/models/course_router.rb

    def change!(user, course)
      user.current_course_id = course.id
      user.save
      course
    end
```

zazaian:
Except for the routing nomenclature on `current_courses#change` I think this all looks good :+1: Bouncing back to you @jwright :basketball:

config/routes.rb
jwright: @zazaian Technically, probably but we will run into the same issue since changing courses needs to be a `GET` request and we are using links to change it. We could submit a form instead of using  the Rails js stuff. Let me know if you think it's worth it and I will redo this.

zazaian:
Eh, I could go either way here @jwright. It's definitely less obnoxious to style a list of links than a list of forms but using a form is probably the correct way to do this since both a record and a session are being modified. I'll let you make the call here. I think it's good to go if you want to merge it, otherwise if you'd like to wrap these in forms and use puts rather than gets that's fine too. :+1:

jwright:
I cannot find the reason that jquery rails (`jquery_ujs`) is not loading on mobile (or not handling the `data-` attributes) so I am going to keep these as `GET` requests.

zazaian:
That's fine @jwright. Seems a little wonky but at least it works.

jwright:
It is wonky but there must be a script that is taking over the `data-` attribute bindings and I cannot find it. Yay, JavaScript!

zazaian:
Haha, no worries -- specifically yay MOBILE javascript!! :+1:  :tada:

PR #1721, opened 2016-03-02
Turn off codeclimate js & focus rubocop assessment

chcholman:
This applies several rounds of cleanup, and turns off a series of rubocop assessments so that we can address the sets of complaints with a little bit more focus. 

Currently: 165 cops on with 1187 reported offenses, 60 turned off.


.rubocop.yml
```diff
@@ -297,14 +302,14 @@ Style/HashSyntax:
                  Prefer Ruby 1.9 hash syntax { a: 1, b: 2 } over 1.8 syntax
                  { :a => 1, :b => 2 }.
   StyleGuide: 'https://github.com/bbatsov/ruby-style-guide#hash-literals'
-  Enabled: true
```
zazaian: @chcholman it seems like it might be best to leave this one on since it's such a small fix and is very easy to spot, and very conventional.

.rubocop.yml
```diff
@@ -338,17 +343,17 @@ Style/InfiniteLoop:
...
   Description: 'Use lambda.call(...) instead of lambda.(...).'
   StyleGuide: 'https://github.com/bbatsov/ruby-style-guide#proc-call'
-  Enabled: true
+  Enabled: false
 
 Style/LeadingCommentSpace:
   Description: 'Comments should start with a space.'
   StyleGuide: 'https://github.com/bbatsov/ruby-style-guide#hash-space'
-  Enabled: true
+  Enabled: false
```
zazaian: I'm fine with leaving this one on though I could see how there are probably thousands of instances of this across the app.

.rubocop.yml
chcholman: @zazaian Cool! I can turn this one back on. I turned anything off that had more than 100 offenses.

.rubocop.yml
```diff
@@ -451,7 +456,7 @@ Style/NumericLiterals:
                  Add underscores to large numeric literals to improve their
                  readability.
   StyleGuide: 'https://github.com/bbatsov/ruby-style-guide#underscores-in-numerics'
-  Enabled: true
+  Enabled: false
```
zazaian: Is this one happening a lot? I'd be okay with leaving this one on.

.rubocop.yml
```diff
@@ -531,17 +536,17 @@ Style/RedundantException:
...
   Description: "Don't use return where it's not required."
   StyleGuide: 'https://github.com/bbatsov/ruby-style-guide#no-explicit-return'
-  Enabled: true
+  Enabled: false
 
 Style/RedundantSelf:
   Description: "Don't use self where it's not needed."
   StyleGuide: 'https://github.com/bbatsov/ruby-style-guide#no-self-unless-required'
-  Enabled: true
+  Enabled: false
```
zazaian: This is probably going to be a fairly large patch unto itself to get all of this into a state of conformity. Very good idea to leave this off until we can fix it app-wide. :+1:

.rubocop.yml
```diff
@@ -577,13 +582,7 @@ Style/SingleLineBlockParams:
...
   Description: 'Avoid single-line methods.'
   StyleGuide: 'https://github.com/bbatsov/ruby-style-guide#no-single-line-methods'
-  Enabled: true
-
-# Style/SingleSpaceBeforeFirstArg:
-#   Description: >-
-#                  Checks that exactly one space is used between a method name
-#                  and the first argument for method calls without parentheses.
-#   Enabled: true
+  Enabled: false
```
zazaian: This is definitely sensible though there probably aren't a ton of instances of this.

.rubocop.yml
```diff
@@ -833,12 +832,12 @@ Metrics/CyclomaticComplexity:
 Metrics/LineLength:
   Description: 'Limit lines to 80 characters.'
```
zazaian: This is probably the most rigorous offender to date. I'm fine with the 80 line limit though it will probably effect code readability overall. We should discuss this threshold before re-activating this. I've gotten used to it but there are several instances in which condensing methods down to 80 lines needlessly impacts readability with little gain besides conforming to this somewhat arbitrary threshold. 80-char limits were imposed once upon a time by IBM punch card systems and that carried over to some modern terminals. Since there really aren't any scenarios in which we'd need to access application code in-place directly on the production or staging servers, and even if there were we'd still be able to use fluid-width ANSI terminals, we should probably come up with the limit that makes for the most readable code (maybe 100? 110?) and iimpose that system-wide if we can rather than using 80 for the sake of it.

http://programmers.stackexchange.com/questions/148677/why-is-80-characters-the-standard-limit-for-code-width

zazaian:
Wow, this must have been a funnnnnnnnnnn PR to build out :grin: This all looks good, and if the goal is to just get as much passing as possible without having to dig through the rest of the app to bring further conformity in large waves right now then I think this is a reasonable compromise in terms of where to start addressing the rubocop issues. Seems good to go to me :+1:

.rubocop.yml
acousticrobot: :+1: +1 I'm all for keeping this one on too.

.rubocop.yml
```diff
@@ -12,7 +17,7 @@ Style/AccessModifierIndentation:
 
 Style/AccessorMethodName:
   Description: Check the naming of accessor methods for get_/set_.
-  Enabled: true
+  Enabled: false
```
jwright: Do we have a lot of offenders of this? I hope we don't and if we do, I think they are small. May be one of the top ones we turn on.

.rubocop.yml
```diff
@@ -89,11 +94,11 @@ Style/BlockDelimiters:
                 always ugly).
                 Prefer {...} over do...end for single-line blocks.
   StyleGuide: 'https://github.com/bbatsov/ruby-style-guide#single-line-blocks'
-  Enabled: true
+  Enabled: false
```
jwright: I think this should stay false

.rubocop.yml
chcholman: Good call - only 9. Turning it back on!

.rubocop.yml
```diff
@@ -189,7 +194,7 @@ Style/DoubleNegation:
 
 Style/EachWithObject:
   Description: 'Prefer `each_with_object` over `inject` or `reduce`.'
-  Enabled: true
+  Enabled: false
```
jwright: Interesting. I did not even know such method existed. :)

.rubocop.yml
```diff
@@ -214,15 +219,15 @@ Style/EmptyLinesAroundAccessModifier:
 
 Style/EmptyLinesAroundBlockBody:
   Description: "Keeps track of empty lines around block bodies."
-  Enabled: true
+  Enabled: false
```
jwright: I would prefer this to stay false.

.rubocop.yml
```diff
@@ -214,15 +219,15 @@ Style/EmptyLinesAroundAccessModifier:
...
 
 Style/EmptyLinesAroundClassBody:
   Description: "Keeps track of empty lines around class bodies."
-  Enabled: true
+  Enabled: false
```
jwright: I would prefer this to stay false.

.rubocop.yml
```diff
@@ -214,15 +219,15 @@ Style/EmptyLinesAroundAccessModifier:
...
 
 Style/EmptyLinesAroundModuleBody:
   Description: "Keeps track of empty lines around module bodies."
-  Enabled: true
+  Enabled: false
```
jwright: I would prefer this to stay false.

.rubocop.yml
jwright: I personally prefer 80 lines, because, well vim.

For times when it would make the line unreadabe, we can use `# rubocop:disable LineLength`

jwright:
I stopped commenting on some of these because I think it's too early, so ignore those :)

Overall, I think this approach is great. I think the following approach would be good for this:

1. Team discuss what rubocop style we prefer for a given topic
1. Add it to the style guidelines
1. Turn it on and fix the regressions
1. Drink

chcholman:
@jwright :+1:

Issue #1720, opened 2016-03-01
#last_login_at being called on nil CourseMembership


zazaian:
We're getting instances in which `#last_login_at` is being called on `nil`in the `LoginEventPerformer` in the event that no `CourseMembership` is found for the login. I'll have to look further into where these records would be enqueued in the first place, but for now it's probably best just to fix the bug and only call `#last_login_at` if a `CourseMembership` is found.

zazaian:
https://rpm.newrelic.com/accounts/367828/applications/2116987/traced_errors/e33450-3ba5fa5f-dff6-11e5-a1e3-b82a72d22a14

zazaian:
@chcholman I have a patch for this immediate issue with `nil` being called on `CourseMembership`, but I want to look into how logins could occur in the system without a `course_id` or `user_id` being saved in the session in the first place.

I'll take a closer look at this tomorrow morning. In the meantime you can see the patch I've started on branch ` 1720-nil_course_membership_error `.

chcholman:
@zazaian thanks

PR #1722, opened 2016-03-02
Resolve nil#last_login_at Error in LoginEventLogger, Speed Up Workflow (Issue 1720)

zazaian:
The crux of this PR is that there are circumstances in which `nil` value for `session[:user_id]` or `session[:course_id]` are being passed into a LoginEventLogger object and breaking the job because it presumes that these have to be present if a login event is triggered. The problem is that this isn't always true, but because the job is breaking early in the perform block we also don't have good logging data to troubleshoot the circumstances under which this is happening. So for now this PR resolves the bug by checking `CourseMembership` before using it, and also by adding logging for these events prior to any other logic being performed. Here's the breakdown:

* Adds a `logger.info` call to the beginning of the `.perform` block with the event `data` being used. This will help us troubleshoot more complex circumstances surrounding these bugs.

* Adds checking for `course_membership_attrs_present?` in order to make sure that there is a `course_id` and `user_id` coming into the `session` before we bother searching the system for the `CourseMembership` that we'd like to work with.

* As per my discussion with @jwright re: the PR in which this code came in, I've replaced all instances in which `subject` was being used to test outcomes with a `let(:result) { some_behavior }` call instead so that the `subject` method can properly define the subject as is conventional in RSpec testing.

* I've also moved the class methods into a `class << self` block on `LoginEventLogger` since it seems more readable than calling `self.` when defining every class method.

* I've also renamed `@data` to `@cached_data` to avoid any confusion over whether the `data` method is an accessor rather than a method argument, and better clarify that workflow.

Again, I don't believe this will resolve the root of the problem, but it will at least move us closer to being able to identify Login scenarios in which either `course_id` or `user_id` aren't present in the `session` for whatever reason, and cleans up the errors that we're receiving in the meantime.

This PR addresses issues further detailed in Issue #1720.

### DIFF
--- a/app/event_loggers/login_event_logger.rb
+++ b/app/event_loggers/login_event_logger.rb
@@ -51,11 +51,13 @@ class LoginEventLogger < ApplicationEventLogger
     end
 
     def update_last_login
-      course_membership.update_attributes(last_login_at: @cached_data[:created_at])
+      course_membership
+        .update_attributes(last_login_at: @cached_data[:created_at])
     end
 
     def course_membership
-      @course_membership ||= CourseMembership.where(course_membership_attrs).first
+      @course_membership ||= CourseMembership.where(course_membership_attrs)
+        .first
     end
 
     # let's make sure that we've got the necessary attributes present in order
--- a/app/event_loggers/login_event_logger.rb
+++ b/app/event_loggers/login_event_logger.rb
@@ -56,7 +56,8 @@ class LoginEventLogger < ApplicationEventLogger
     end
 
     def course_membership
-      @course_membership ||= CourseMembership.where(course_membership_attrs)
+      @course_membership ||= CourseMembership
+        .where(course_membership_attrs)
         .first
     end
 
--- a/app/event_loggers/login_event_logger.rb
+++ b/app/event_loggers/login_event_logger.rb
@@ -56,9 +56,8 @@ class LoginEventLogger < ApplicationEventLogger
     end
 
     def course_membership
-      @course_membership ||= CourseMembership
-        .where(course_membership_attrs)
-        .first
+      @course_membership ||= \
+        CourseMembership.where(course_membership_attrs).first
     end
 
     # let's make sure that we've got the necessary attributes present in order
--- a/app/event_loggers/login_event_logger.rb
+++ b/app/event_loggers/login_event_logger.rb
@@ -38,7 +38,9 @@ class LoginEventLogger < ApplicationEventLogger
   end
 
   def self.previous_last_login_at
-    course_membership.last_login_at.to_i if course_membership.last_login_at
+    if course_membership && course_membership.last_login_at
+      course_membership.last_login_at.to_i
+    end
   end
 
   def self.update_last_login
--- a/app/event_loggers/login_event_logger.rb
+++ b/app/event_loggers/login_event_logger.rb
@@ -31,10 +31,11 @@ class LoginEventLogger < ApplicationEventLogger
 
   # perform block that is ultimately called by Resque
   def self.perform(event_type, data = {})
-    @data = data
+    logger.info "Starting #{@queue} with data #{data}"
+    @cached_data = data
     data[:last_login_at] = previous_last_login_at
     super
-    update_last_login
+    update_last_login if course_membership
   end
 
   def self.previous_last_login_at
@@ -44,7 +45,7 @@ class LoginEventLogger < ApplicationEventLogger
   end
 
   def self.update_last_login
-    course_membership.update_attributes(last_login_at: @data[:created_at])
+    course_membership.update_attributes(last_login_at: @cached_data[:created_at])
   end
 
   def self.course_membership
@@ -52,6 +53,6 @@ class LoginEventLogger < ApplicationEventLogger
   end
 
   def self.course_membership_attrs
-    { course_id: @data[:course_id], user_id: @data[:user_id] }
+    { course_id: @cached_data[:course_id], user_id: @cached_data[:user_id] }
   end
 end
--- a/app/event_loggers/login_event_logger.rb
+++ b/app/event_loggers/login_event_logger.rb
@@ -28,31 +28,51 @@ class LoginEventLogger < ApplicationEventLogger
   end
 
   ## class methods, for use when being called directly by Resque
-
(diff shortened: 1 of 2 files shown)


app/event_loggers/login_event_logger.rb
```diff
@@ -28,28 +28,53 @@ def event_type
...
 
-  # perform block that is ultimately called by Resque
-  def self.perform(event_type, data = {})
-    @data = data
-    data[:last_login_at] = previous_last_login_at
-    super
-    update_last_login
-  end
+    def previous_last_login_at
+      # Let's be sure to check here whether last_login_at exists before
```
jwright: Very good comment here :100:

app/event_loggers/login_event_logger.rb
```diff
@@ -28,28 +28,53 @@ def event_type
   end
 
   ## class methods, for use when being called directly by Resque
+  class << self
+    # perform block that is ultimately called by Resque
+    def perform(event_type, data = {})
+      logger.info "Starting #{@queue} with data #{data}"
+      @cached_data = data
```
jwright: Great rename

app/event_loggers/login_event_logger.rb
```diff
@@ -28,28 +28,53 @@ def event_type
...
+    # to find the CourseMembership before we attempt to find it. There's no good
+    # reason to perform a search for the CourseMembership if we know that it
+    # can't possibly exist.
+    #
+    def course_membership_present?
+      course_membership_attrs_present? && course_membership
+    end
+
+    def course_membership_attrs_present?
+      @cached_data[:course_id].present? && @cached_data[:user_id].present?
```
jwright: Because we are duplicating the attrs here and on line 77, that coupling may result in a future bug. What about changing this line to something like:

```
course_membership_attrs.values.none?(&:nil?) # not sure if that is exact....using GH here :)
```

spec/event_loggers/login_event_logger_spec.rb
```diff
@@ -14,13 +14,10 @@
...
 
-  # this needs to be declared since Resque interacts with class-level instance
-  # variables, and using mulitple class instances could misrepresent class-level
-  # instance variable circumstances
-  let(:class_instance) { LoginEventLogger }
+  subject { LoginEventLogger }
 
   # build this off of the class instance for consistent behavior
-  let(:new_logger) { class_instance.new(event_session) }
+  let(:new_logger) { described_class.new(event_session) }
```
jwright: :+1:

spec/event_loggers/login_event_logger_spec.rb
```diff
@@ -37,107 +34,194 @@
...
-        result
+      context "a course membership is present" do
+        it "updates the last login" do
+          expect(subject).to receive(:update_last_login)
+          result
+        end
+      end
+
+      context "no course membership is present" do
+        it "updates the last login" do
```
jwright: I think this should be `"does not update the last login"`

jwright:
@zazaian Love this whole thing :heart_eyes: 

I just had a few small comments and it's ready to go.

app/event_loggers/login_event_logger.rb
zazaian: Cool, I changed this based on your comment from the last PR :+1:

app/event_loggers/login_event_logger.rb
zazaian: Cool, I think this is a better solution and was originally going to implement something very similar to this but didn't want to get too fancy. I think this is a better solution because it's future-proof and doesn't reiterate the `course_membership_attrs` values. I'll happily make this change.

spec/event_loggers/login_event_logger_spec.rb
zazaian: Yup, sorry -- missed this one.

zazaian:
@jwright awesome, thanks a ton for the feedback here. I'll make those two changes and bounce it back to you :basketball: :athletic_shoe:

app/event_loggers/login_event_logger.rb
zazaian: Also nice use of `none?` here, fits the circumstance perfectly.

app/event_loggers/login_event_logger.rb
```diff
@@ -28,28 +28,53 @@ def event_type
...
+    # to find the CourseMembership before we attempt to find it. There's no good
+    # reason to perform a search for the CourseMembership if we know that it
+    # can't possibly exist.
+    #
+    def course_membership_present?
+      course_membership_attrs_present? && course_membership
+    end
+
+    def course_membership_attrs_present?
+      course_membership_attrs.values.all?(&:present?)
```
zazaian: @jwright note that I just used `all?(&:present?)` instead of `none?(&:nil?)` to avoid the unnecessary double-negative.

spec/event_loggers/login_event_logger_spec.rb
```diff
@@ -14,130 +14,230 @@
...
-        result
+      context "a course membership is present" do
+        it "updates the last login" do
+          expect(subject).to receive(:update_last_login)
+          result
+        end
+      end
+
+      context "no course membership is present" do
+        it "does not update the last login" do
```
zazaian: @jwright this was updated also as you noted.

zazaian:
@jwright all of those changes have been made here and I even knocked a couple of the 80+ character lines down to <= 80 for good measure. Please re-review when you get a chance. Thanks :bow:

app/event_loggers/login_event_logger.rb
jwright: Did I just compliment myself? :cold_sweat:

app/event_loggers/login_event_logger.rb
jwright: :+1: Much better.

jwright:
@zazaian I think this is great and ready to go!

zazaian:
@jwright awesome! Thanks for the feedback :mega: Much appreciated.

zazaian:

app/event_loggers/login_event_logger.rb
zazaian: Haha, no -- you were just confused by what was happening here so I changed the variable name to clarify :wrench:

PR #1723, opened 2016-03-02
1670 Grade Proctor

### LINKED ISSUE #1670: Grade Proctor

jwright:
### Background

The idea of a `Proctor` is to ensure that given a resource (in this case, `Grade`), that it can be viewable or updatable within a given context (in this case, a specific `User` and a specific `Course`). This would be the single place where permissions for viewing or updating the resource would be analyzed. 

### Why is the current situation a problem?

The issue is that we have multiple ways to determine if a user can either edit a `Grade` or if a user can view a `Grade`. These live in views, controllers, and models and they are all different. Here are a few places where permissions are checked:

```
AssignmentPresenter#grades_available_for?(user)
```

```
AssignmentPresenter#scores_for(user)
```

```
GradesController#update
```

```
Grade#is_student_visible?
```

```
ChallengeGrade#is_student_visible?
```

```
GradesController#before_filter
```

```
Grade.student_visible scope
```

```
Views -> Grades -> mass_edit
```

### What needs to be handled

* We need to be able to ask the proctor if a grade is visible
* We need to be able to ask the proctor if a grade is editable
* We need to have one place to use the Grade statuses
* We need to be able to integrate with Canable (delayed until we move to CanCan (#1748)
* We need to be able to access this logic from within Models, Views, and Controllers
* This needs to work with regular grades
* This needs to work for predicted grades as well

### Implementation

The `GradeProctor` lives in `lib` since it does not require any real knowledge of application. It only requires access to specific messages.

Only `#viewable?` and `#updatable?` are implemented in this situation. I have not seen much use for a `#destroyable?` or `#creatable?` modules since these are usually performed by the system.

The `#viewable?` and `#updatable?` methods are implemented in separate modules to keep small, focused files.

The `GradeProctor` works with both a `Grade` and a `NullGrade`. `NullGrade` had to be updated in some ways to respond to additional messages.

#### #viewable?

This is used in the following ways:

```
GradeProctor.new(grade).viewable?

GradeProctor.new(grade).viewable? current_user

GradeProctor.new(grade).viewable? current_user, current_course
```

The `viewable?` method does not require a user nor a course. When they are not specified, it uses the `#student` and the `#course` methods on the grade.

A `Grade` is viewable when:

* the `grade` is not nil
* the `grade` is for the `course`
* the `user` is a staff member within the `course`
* the `user` is a student in the `course` and it has been released or graded and a release is not necessary

#### #updatable?

This is used in the following ways:

```
GradeProctor.new(grade).updatable?

GradeProctor.new(grade).updatable? current_user

GradeProctor.new(grade).updatable? current_user, current_course
```

The `updatable?` method does not require a user nor a course. When they are not specified, it uses the `#student` and the `#course` methods on the grade.

A `Grade` is updatable when:

* the `grade` is not nil
* the `grade` is for the `course`
* the `user` is a staff member within the `course`
* the `user` is a student in the `course` and the `grade.assignment` is student logged
* the `user` is a student in the `course` and the `grade` is predicted

#### additional changes

I started going down the path of also using the `GradeProctor` for `ChallengeGrade`s. In order to do this, I needed to combine some of the functionality of `Grade` and `ChallengeGrade` that were duplicated across the two objects. I combined this functionality into a `GradeStatus` concern.

Once I found that `ChallengeGrade` is geared towards a `Team` instead a student, then that made it impossible to use a `GradeProctor` for a `ChallengeGrade`. This, however, does not invalidate the concern as it removes duplicate functionality.

### Caveats

This is a first pass and I am sure logic where permissions and status is checked for a `Grade` was not replaced. The focus of this initial pass was to remove the `#is_student_visible?` method on `Grade`. This was because that method was incomplete and additional things were not checked or checked outside of that method.

In addition, the [move to `CanCan`](https://github.com/UM-USElab/gradecraft-development/issues/1748) will allow us to remove some of the custom `before_filter` and allow us to wrap the `GradeProctor` in a CanCan `Ability`.

Closes #1670


jwright:
I think `GradeProctor` is better than `Proctor::Grade` and even if we have multiple "proctors", we can use a base `Proctor` object/module/behavior.

jwright:
@acousticrobot Thanks! I have updated the PR to reflect that as a requirement.

chcholman:
@jwright This is looking awesome! 

Even though they're named similarly, I would separate ChallengeGrade out into a different Proctor, as they apply at the team level rather than the student level. (The counter would be that if we're including ChallengeGrade in this, then we should also include EarnedBadge, because they're a closer parallel, directly connecting one Badge to one Student.)

jwright:
@chcholman Awesome! Great suggestions. I agree, I think they should be separated.

app/models/concerns/grade_status.rb
```diff
@@ -0,0 +1,59 @@
+# Model concern for behaviors around a grade status for an ActiveRecord class
+# that represents a grade with a given status
+module GradeStatus
+  extend ActiveSupport::Concern
+
```
acousticrobot: Do we want to move these lines out of the Grade model and put them here?

```
  STATUSES = ["In Progress", "Graded", "Released"]
  UNRELEASED_STATUSES = ["In Progress", "Graded"]
```

app/models/concerns/grade_status.rb
```diff
@@ -0,0 +1,59 @@
...
+  class_methods do
+    # Used in class definitions to specify which belongs_to relationship responds to a release_necessary flag
+    # This is used in the scopes to determine what relationship to join with
+    # USAGE:
+    # class Grade
+    #   releasable_through :assignment
+    # end
+    def releasable_through(relationship=nil)
+      @releasable_relationship = relationship
+    end
```
acousticrobot: This is a very cool abstraction.

app/models/grade.rb
```diff
@@ -96,12 +92,8 @@ def feedback_reviewed!
...
   end
 
-  def is_graded?
-    self.status == "Graded"
-  end
-
-  def in_progress?
-    self.status == "In Progress"
+  def predicted?
+    self.predicted_score != 0 && !self.predicted_score.nil?
```
acousticrobot: I know we were talking about this before, but to reiterate:

 I don't know how we can add a `predicted?` method unless we create a `student_modified` field that works similarly to `instructor_modified` (and this would then probably be shared with self grading assignments).

predicted_score has a `default 0 null: false` database constraint, so `!self.predicted_score.nil?` will always be true.

the problem with `self.predicted_score != 0` is that a student can willfully predict a zero for the assignments they don't plan on completing, and this is very common since GC philosophy encourages courses with multiple paths to completion that include more assignments than one student would complete during the course.

app/models/unlock_condition.rb
```diff
@@ -177,7 +177,7 @@ def check_grade_earned_condition(student)
       check_if_grade_earned_meets_condition_value(grade)
     elsif condition_date?
       check_if_grade_earned_met_condition_date(grade)
-    elsif grade.present? && grade.is_student_visible?
+    elsif grade && GradeProctor.new(grade).viewable?(student, grade.course)
```
acousticrobot: I think you can remove `grade.course` here now.

acousticrobot: Also, below you changes maintained `grade.present?`; I would either use `present?` here or remove it on lines 188, 197

app/models/user.rb
```diff
@@ -330,9 +330,8 @@ def points_to_next_level(course)
 
   # Checking specifically if there is a released grade for an assignment
   def grade_released_for_assignment?(assignment)
-    if grade_for_assignment(assignment).present?
-      grade_for_assignment(assignment).is_student_visible?
-    end
+    grade = grade_for_assignment(assignment)
```
acousticrobot: I'm noticing the `grade_for_assignment` method for the first time, which current is:

`grades.where(assignment_id: assignment.id).first || grades.new(assignment: assignment)`

This will never save the or fully create the grade object. I think if we change this to

`Grade.find_or_create(assignment.id,self.id)` we will have an actual grade object with the correct course id and can change lines 334 and 517 to `GradeProctor.new(grade).viewable? self`

app/models/null_grade.rb
```diff
@@ -9,10 +9,27 @@ def assignment
...
   def is_student_visible?
     true
   end
 
+  def is_released?
+    false
+  end
+
+  def is_graded?
+    true
```
acousticrobot: It looks like this still works with the current generic predictor, but we'll just have to watch this one and make sure it doesn't cause the predictor widgets to show up as final grade widgets (this is a fragile logic path that involves both ruby and js and often changes as new grade functionality is added).

app/models/grade.rb
chcholman: Chiming in on the `predicted?` method - this method produces a number that we actually show back to the instructors to help them gauge how many students are planning on _doing_ an assignment, thus the filtering out of 0 predictions. That means the method should definitely be renamed to be accurate, but kept :)

app/models/concerns/grade_status.rb
jwright: I thought about that but constants should not be defined in a module because the module can be included in multiple classes, therefore causing a "already defined" warning. I vote for keeping them defined in Grade for now or breaking them out into another class.

jwright: Actually, I am wrong about that. They get the module scope so they will be used as `GradeStatus::STATUSES` and `GradeStatus::UNRELEASED_STATUSES`.

I will make this change.

app/models/grade.rb
acousticrobot: We could go with
```
effort_predicted?
  self.predicted_score > 0
end
```

jwright: @acousticrobot Ah, yes. I like the idea of a `student_modified` field. I can add that (in a different PR that will be merged in first).

Is there another way that may be an easier way to implement this? I am still confused on how we differentiate predicted grades from normal grades.

app/models/concerns/grade_status.rb
acousticrobot: :+1

app/models/unlock_condition.rb
jwright: :+1:

app/models/user.rb
jwright: I will have to see what the downstream issues are by using a create here. I am not sure if we want to have the Grade created here or if it is needed downstream.

app/models/null_grade.rb
jwright: Would the addition of a `#predicted?` help here if we add that (with `student_modified`) to a `Grade`?

jwright:
@acousticrobot The question though is how do we determine if a Grade is updatable for a predicted grade? We may not need to implement this yet since we just use it on the `#mass_edit` view. Thoughts?

app/models/user.rb
jwright: @acousticrobot I took a look at where `User#grade_for_assignment` is used and it is used a lot. In all of these instances, I don't believe we want to create a grade in these situations. Especially because it is used in views a lot and I don't believe checking this value should have the side effect of creating a new `Grade` object. I left this code the way it. Please let me know if you have additional thoughts.

jwright:
@acousticrobot Thanks for the info!

I removed the `Grade#predicted?` method for this PR since the `GradeProctor` is not being used in the predictor.

Instead of adding another method `Grade#prediction_updatable?` (which is the problem we are trying to avoid), I believe we should wait until the predicted scores are extracted to a seperate model/table like you suggested. That way, we can make use of a PredictedGradeProctor instead.

app/models/user.rb
acousticrobot: I think the method may need to be refactored or at least revisited, but I agree that it is beyond the scope of this PR.

PR #1725, opened 2016-03-03
1520 exclude grade

chcholman:
This PR: 

- Removes the multiple updater job and sends everything through the single grade updater job
- Removes the _improved_cached_course_score_ method and sends everything through the _cached_course_score_
- Adds a scope that filters out grades Instructors have marked as to be excluded
- Creates a button on the student syllabus page that allows a staff member to mark a grade to be excluded


app/controllers/grades_controller.rb
```diff
@@ -1,8 +1,11 @@
 class GradesController < ApplicationController
   respond_to :html, :json
-  before_filter :set_assignment, only: [:show, :edit, :update, :destroy, :submit_rubric]
-  before_filter :ensure_staff?, except: [:feedback_read, :self_log, :show, :predict_score, :async_update] # TODO: probably need to add submit_rubric here
-  before_filter :ensure_student?, only: [:feedback_read, :predict_score, :self_log]
+  before_filter :set_assignment, only: [:show, :edit, :update, :destroy,
```
acousticrobot: This is a personal style preference, but I actually find the original before filters, at ~100 characters, easier to read on a single line.

app/controllers/grades_controller.rb
```diff
@@ -64,8 +68,8 @@ def update
 
       # @mz TODO: ADD SPECS
       if @grade.is_released? || (@grade.is_graded? && ! @assignment.release_necessary)
-        @grade_updater_job = GradeUpdaterJob.new(grade_id: @grade.id)
-        @grade_updater_job.enqueue
+        grade_updater_job = GradeUpdaterJob.new(grade_id: @grade.id)
+        grade_updater_job.enqueue
```
acousticrobot: :+1:

app/controllers/grades_controller.rb
```diff
@@ -439,6 +481,18 @@ def mass_update_grade_ids
...
+                           course_id: current_course.id).enqueue
+  end
+
+  def enqueue_multiple_grade_update_jobs(grade_ids)
+    grade_ids.each do |grade_id|
+      grade_updater_job = GradeUpdaterJob.new(grade_id: grade_id)
+      grade_updater_job.enqueue
+    end
+  end
+
```
acousticrobot: I think this is a good refactor, even though it's more code in the controller, considering that `multiple_grade_update_performer` was 95% identical to `grade_update_performer` :clap:

spec/controllers/grades_controller_spec.rb
```diff
@@ -326,6 +326,61 @@
...
+          feedback_reviewed: true,
+          feedback_reviewed_at: Time.now,
+          instructor_modified: true,
+          graded_at: DateTime.now
+        )
+        post :exclude, {id: @grade.id}
+
+        @grade.reload
+        expect(@grade.excluded_from_course_score).to eq(true)
+        expect(@grade.raw_score).to eq(500)
```
acousticrobot: We may want to add `expect(@grade.score).to eq(500)` here as well. It assumes the calculation of the score, which we aren't testing here, but it will also assure that the exclusion isn't calculated into the grade's score, but only on the assignment type level and above.

spec/controllers/grades_controller_spec.rb
```diff
@@ -326,6 +326,61 @@
...
+      end
+
```
acousticrobot: I would also add

```
      it "adds exclusion metadata" do
        current_time = DateTime.now
        post :exclude, {id: @grade.id}

        @grade.reload
        expect(@grade.excluded_date).to be > current_time
        expect(@grade.excluded_by).to eq(@professor.id)
      end
```

spec/controllers/grades_controller_spec.rb
```diff
@@ -326,6 +326,61 @@
...
+        @grade.update(
+          raw_score: 500,
+          status: "Graded",
+        )
+        post :include, {id: @grade.id}
+
+        @grade.reload
+        expect(@grade.excluded_from_course_score).to eq(false)
+        expect(@grade.raw_score).to eq(500)
+      end
```
acousticrobot: I would add the "excluded details" to the expectations:

```
      it "marks the Grade as included, and clears the excluded details" do
        @grade.update(
          raw_score: 500,
          status: "Graded",
          excluded_from_course_score: true,
          excluded_by: 2,
          excluded_date: Time.now
        )
        post :include, {id: @grade.id}

        @grade.reload
        expect(@grade.excluded_from_course_score).to eq(false)
        expect(@grade.raw_score).to eq(500)
        expect(@grade.excluded_date).to be nil
        expect(@grade.excluded_by).to be nil
      end
```

db/schema.rb
```diff
@@ -437,6 +437,9 @@
     t.datetime "graded_at"
     t.integer  "adjustment_points",                      default: 0,     null: false
     t.text     "adjustment_points_feedback"
+    t.boolean  "excluded_from_course_score",             default: false
+    t.datetime "excluded_date"
+    t.integer  "excluded_by"
```
acousticrobot: In order to match existing fields (`updated_at`, `created_at`,`graded_at`, and  `graded_by_id`) maybe these should change to `excluded_at`, and `excluded_by_id`

app/controllers/grades_controller.rb
chcholman: I agree with you. Switching it back.

db/schema.rb
chcholman: :+1: Excellent catch

app/controllers/grades_controller.rb
chcholman: thanks!

spec/controllers/grades_controller_spec.rb
chcholman: Nice, adding these.

spec/controllers/grades_controller_spec.rb
chcholman: Added.

spec/controllers/grades_controller_spec.rb
chcholman: Added.

PR #1728, opened 2016-03-07
Feature/1704 override rubric total points

acousticrobot:
This feature adds the ability to adjust the total points assigned on a rubric graded assignment. It adds an `adjustment_points` and `adjustment_points_feedback` field on the grade, and uses the existing `final_score` to store the adjusted score.

### Grading Form:
screen shot 2016-03-09 at 12 56 14 pm

### Student Grade View:
screen shot 2016-03-09 at 2 02 20 pm

### Ruby Concerns:
A temporary method `final_points` has been added to the Grade model, to handle older grades that have a null `final_score` by passing through the `raw_score`.

### Javascript Concerns:
The Smart Number module has been updated to handle negative numbers.

### DIFF
--- a/app/assets/javascripts/angular/controllers/GradeRubricCtrl.js.coffee
+++ b/app/assets/javascripts/angular/controllers/GradeRubricCtrl.js.coffee
@@ -90,7 +90,7 @@
     )
     points
 
-  $scope.pointsGivenAfterAdjustment = ()->
+  $scope.finalPoints = ()->
     $scope.pointsGiven() + $scope.adjustmentPoints()
 
   $scope.gradedCriteria = ()->
--- a/app/assets/javascripts/angular/controllers/GradeRubricCtrl.js.coffee
+++ b/app/assets/javascripts/angular/controllers/GradeRubricCtrl.js.coffee
@@ -185,7 +185,7 @@
 
   $scope.gradeParams = ()->
     {
-      raw_score: $scope.pointsGivenAfterAdjustment(),
+      raw_score: $scope.pointsGiven(),
       feedback: $scope.grade.feedback,
       status:   $scope.grade.status,
       points_adjustment: $scope.grade.points_adjustment,
--- a/app/assets/javascripts/angular/controllers/GradeRubricCtrl.js.coffee
+++ b/app/assets/javascripts/angular/controllers/GradeRubricCtrl.js.coffee
@@ -31,11 +31,11 @@
     )
     points or 0
 
-  $scope.pointsAdjustment = ()->
-    parseInt($scope.grade.points_adjustment) || 0
+  $scope.adjustmentPoints = ()->
+    parseInt($scope.grade.adjustment_points) || 0
 
   $scope.pointsAdjusted = ()->
-    $scope.pointsAdjustment() != 0
+    $scope.adjustmentPoints() != 0
 
   $scope.pointsDifference = ()->
     $scope.pointsPossible() - $scope.pointsGiven()
@@ -91,7 +91,7 @@
     points
 
   $scope.pointsGivenAfterAdjustment = ()->
-    $scope.pointsGiven() + $scope.pointsAdjustment()
+    $scope.pointsGiven() + $scope.adjustmentPoints()
 
   $scope.gradedCriteria = ()->
     criteria = []
@@ -188,8 +188,8 @@
       raw_score: $scope.pointsGiven(),
       feedback: $scope.grade.feedback,
       status:   $scope.grade.status,
-      points_adjustment: $scope.grade.points_adjustment,
-      points_adjustment_feedback: $scope.grade.points_adjustment_feedback
+      adjustment_points: $scope.grade.adjustment_points,
+      adjustment_points_feedback: $scope.grade.adjustment_points_feedback
     }
 
   # Document any updates to this format in the specs: /spec/support/api_calls/rubric_grade_put.rb
--- a/app/assets/javascripts/angular/controllers/GradeRubricCtrl.js.coffee
+++ b/app/assets/javascripts/angular/controllers/GradeRubricCtrl.js.coffee
@@ -31,6 +31,12 @@
     )
     points or 0
 
+  $scope.pointsAdjustment = ()->
+    parseInt($scope.grade.points_adjustment) || 0
+
+  $scope.pointsAdjusted = ()->
+    $scope.pointsAdjustment() != 0
+
   $scope.pointsDifference = ()->
     $scope.pointsPossible() - $scope.pointsGiven()
 
@@ -83,6 +89,9 @@
         points += criterion.selectedLevel.points
     )
     points
+
+  $scope.pointsGivenAfterAdjustment = ()->
+    $scope.pointsGiven() + $scope.pointsAdjustment()
 
   $scope.gradedCriteria = ()->
     criteria = []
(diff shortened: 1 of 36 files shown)


app/assets/javascripts/angular/controllers/GradeRubricCtrl.js.coffee
```diff
@@ -31,6 +31,12 @@
     )
     points or 0
 
+  $scope.adjustmentPoints = ()->
+    parseInt($scope.grade.adjustment_points) || 0
+
+  $scope.pointsAdjusted = ()->
```
jwright: Is this our convention in JS for predicate methods? Would a rename to `arePointsAdjusted` be better so that we know it returns a boolean? I am unsure if we use `is` and `are` for predicate methods elsewhere.

app/assets/javascripts/angular/controllers/GradeRubricCtrl.js.coffee
```diff
@@ -84,6 +90,9 @@
     )
     points
 
+  $scope.pointsGivenAfterAdjustment = ()->
```
jwright: Rename to `finalPoints` to match backend?

app/models/grade.rb
```diff
@@ -36,7 +37,7 @@ class Grade < ActiveRecord::Base
   accepts_nested_attributes_for :earned_badges, reject_if: proc { |a| (a["score"].blank?) }, allow_destroy: true
 
   before_validation :cache_associations
-  before_save :cache_point_total
+  before_save :calculate_points
```
jwright: :+1:

app/models/grade.rb
```diff
@@ -108,33 +109,41 @@ def in_progress?
   # Handle raw score attributes with commas (ex "300,000")
   def raw_score=(rs)
     if rs.class == String
-      rs.gsub!(",","").to_i
+      rs.delete!(",").to_i
```
jwright: :+1:

app/models/grade.rb
```diff
@@ -108,33 +109,41 @@ def in_progress?
...
+  # is not available.
+  #
+  # TODO between semesters:
+  #  * run rake task to calculate update all grades with final scores
+  #  * change codebase and add migrations to standardize points/score nomenclature
+  #    -  point_total -> full_points
+  #    -  raw_score -> raw_points
+  #    -  final_score -> final_points
+  #  * remove this method, but keep calls to grade.final_points in the views
+  #
```
jwright: :100:

app/models/grade.rb
```diff
@@ -108,33 +109,41 @@ def in_progress?
...
+  def final_points
+    final_score || raw_score
   end
 
   def predicted_score
     self[:predicted_score] || 0
   end
 
-  def point_total
```
jwright: Was this method not used?

app/models/grade.rb
```diff
@@ -185,14 +194,38 @@ def check_unlockables
...
   private
 
+  # full points (with student's weighting)
+  def calculate_point_total
+    assignment.point_total_for_student(student)
+  end
+
+  # totaled points (adds adjustment, without weighting)
+  def calculate_final_score
+    return nil unless raw_score.present?
```
jwright: Do we want to return `nil` here or 0? It seems like we would want to return 0 so we don't have to deal with magic nils that may be interpreted to have meaning.

We also need to guard against a `nil` for `adjustment_points`

```
  def calculate_final_score
    (raw_score || 0) + (adjustment_points || 0)
  end
```

app/models/grade.rb
```diff
@@ -185,14 +194,38 @@ def check_unlockables
...
+    raw_score + adjustment_points
+  end
+
+  # points with student's weighting
+  def calculate_score
+    return nil unless raw_score.present?
```
jwright: Same question as above.

In addition, we probably need to protect against a `nil` `weighting` value

app/models/grade.rb
```diff
@@ -185,14 +194,38 @@ def check_unlockables
...
+
   def self.student_visible_sql
     ["status = 'Released' OR (status = 'Graded' AND assignments.release_necessary = ?)", false]
   end
 
   def save_student
-    if self.raw_score_changed? || self.status_changed?
-      student.save
-    end
+    return unless self.raw_score_changed? || self.status_changed?
```
jwright: Do we want to return `false` here since `save` returns a boolean? The consumer may expect a boolean from this method.

app/models/null_grade.rb
```diff
@@ -41,6 +41,10 @@ def final_score
     nil
   end
 
+  def final_points
+    nil
```
jwright: Again. Not sure if using `nil` here is the right thing to do.

spec/models/assignment_type_spec.rb
```diff
@@ -191,7 +191,7 @@
 
     it "returns the total score a student has earned for an assignment type and has a reduced final score" do
       assignment = create(:assignment, course: world.course, assignment_type: assignment_type, release_necessary: true)
-      grade = create(:grade, student: student, raw_score: 100, final_score: 75, assignment: assignment, status: "Released")
+      grade = create(:grade, student: student, raw_score: 100, adjustment_points: -25, assignment: assignment, status: "Released")
```
jwright: We can probably use `build` here which will cut down on processing since it does not have to hit the database.

spec/models/grade_spec.rb
```diff
@@ -61,6 +61,37 @@
     end
   end
 
+  describe "calculation of final_score" do
+    it "is nil when raw_score is nil" do
```
jwright: These will need to change if we decide to not use `nil`.

jwright:
@acousticrobot This is really thorough and well tested. Great job! 

I just had one large overall question, comment. Should we store/return `nil` in point values? My inclination is no, because then dealing with nils is a pain (and a major source of bugs in Ruby). If we are using `nil`s to signify something, then that is another story (and something we should probably not be doing either). I'd love to get your overall thoughts.

app/assets/javascripts/angular/controllers/GradeRubricCtrl.js.coffee
acousticrobot: Good call, now that we know what it's going to be called!

app/assets/javascripts/angular/controllers/GradeRubricCtrl.js.coffee
acousticrobot: We don't, but I wouldn't mind if we went through all the Angular and renamed all the methods that return booleans.

app/models/grade.rb
acousticrobot: I moved it to private and renamed it `calculate_point_total`

app/models/grade.rb
acousticrobot: Grades that exists but haven't been graded have a raw_score and score of nil, so the student doesn't think they got a 0.
Adjustment points though can be zero by default to I made them default to zero in the schema.

app/models/grade.rb
acousticrobot: I'm not sure about how nil weights are handled internally, but I built this from the existing logic: `final_score || ((raw_score * assignment_weight).round if raw_score.present?)  || nil`

app/models/grade.rb
acousticrobot: I don't want to change this too much because I don't know what this is handling. This was clean up from rubocop.

spec/models/assignment_type_spec.rb
acousticrobot: It has to hit the `before_save` method to calculate score.

PR #1729, opened 2016-03-07
Insert new criterion levels between min and max

### LINKED ISSUE #1703: UX around reordering of criteria levels
As an instructor, when I add criteria levels, they should appear inside of the lowest and highest levels.

acousticrobot:
This is a UX enhancement which inserts new levels in between the min and max
levels.
screen shot 2016-03-07 at 4 05 20 pm

closes #1703


jwright:
Beutismus!

PR #1731, opened 2016-03-08
use grade status constants

### LINKED ISSUE #1695: Grade status constants
Throughout the app we have grade status arrays: `["In Progress", "Graded", "Released"] , ["In Progress", "Graded"]...` 

As @jwright mentioned on https://github.com/UM-USElab/gradecraft-development/pull/1685/files#r53848420
>>I feel like these should be in the Grade model and not in the controller as constants. We already have Grade::STATUSES that can be used. We should perhaps have Grade::UNRELEASED_STATUSES.

Throughout the app, we shoud refactor these out of the controllers and views and into constants on the Grade model.

acousticrobot:
As per suggestion from @jwright, this refactors all grade status arrays to reference constants on the Grade model. (*Note:* one reference remains on `app/assets/javascripts/angular/controllers/GradeCtrl.js.coffee`; this file is due for a more thorough refactoring, so I have left it for now.)

closes #1695

### DIFF
--- a/app/models/grade.rb
+++ b/app/models/grade.rb
@@ -16,9 +16,7 @@ class Grade < ActiveRecord::Base
     :feedback_reviewed, :feedback_reviewed_at, :is_custom_value, :graded_at
 
   STATUSES = ["In Progress", "Graded", "Released"]
-
-  # Note Pass and Fail use term_for in the views
-  PASS_FAIL_STATUS = ["Pass", "Fail"]
+  UNRELEASED_STATUSES = ["In Progress", "Graded"]
 
   belongs_to :course, touch: true
   belongs_to :assignment, touch: true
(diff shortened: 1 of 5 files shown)


app/models/grade.rb
```diff
@@ -16,9 +16,10 @@ class Grade < ActiveRecord::Base
     :feedback_reviewed, :feedback_reviewed_at, :is_custom_value, :graded_at
 
   STATUSES = ["In Progress", "Graded", "Released"]
+  UNRELEASED_STATUSES = ["In Progress", "Graded"]
 
   # Note Pass and Fail use term_for in the views
-  PASS_FAIL_STATUS = ["Pass", "Fail"]
+  PASS_FAIL_STATUSES = ["Pass", "Fail"]
```
jwright: This was changed but it doesn't seem to be used anywhere. Can we remove this?

jwright:
I think this is good. I would just suggest a test for `Grade::STATUSES` and `Grade::UNRELEASED_STATUSES` for regression. It would be a simple test:

```
describe ".STATUSES" do
  it "returns an array of all the status values" do
    expect(described_class.STATUSES).to eq ["In Progress", "Graded", "Released"]
  end
end

describe ".UNRELEASED_STATUSES" do
  #...
end
```

PR #1736, opened 2016-03-08
Reorganize badge views and controllers

ksofa2:
This is a work in progress, not ready to merge. I'm looking for feedback in general on how I've reorganized things (I've only moved code around, there's much refactoring to do still).


ksofa2:
I can't see the details of the Codeship build failure. Can I get access to see more?

app/views/notification_mailer/earned_badge_awarded.html.haml
```diff
@@ -21,5 +21,5 @@
 %p{:align => "center"}
   Check out what other #{ @course.badge_term.pluralize.downcase } you can earn in #{ @course.name }
   = succeed "." do
-    %a{:href => "https://www.gradecraft.com/my_badges"} here
-%hr/
\ No newline at end of file
+    %a{:href => "https://www.gradecraft.com/badges"} here
```
ksofa2: I assume I can use badges_path here, but I'm not sure how to test this change.

config/routes.rb
```diff
@@ -307,7 +307,7 @@
   get "timeline" => "students#timeline"
   get "syllabus" => "students#syllabus"
   get "course_progress" => "students#course_progress"
-  get "my_badges" => "students#badges"
+  get "my_badges", to: redirect("/badges")
```
ksofa2: I don't think this needs to be a absolute path.

chcholman:
@ksofa2 Awesome! Great to see this. 

I think @jwright has reached out about setting you up with CodeShip access.

Email templates can be viewed at http://localhost:1080

I think we'd probably want to take the my_badges route out all together?

ksofa2:
The only reason I'm redirecting the /my_badges URL to /badges in order to not break any bookmarks and incoming links users might have created that reference the old URL.

(If we keep the redirect, the entry in routes.rb should get a comment explaining the history.)

PR #1741, opened 2016-03-09
1740 Nil Changeset bug

### LINKED ISSUE #1740: Nil changeset bug
https://rollbar.com/gradecraft/gradecraft-development/items/1482/?person_page=0&#traceback

jwright:
A bug occurred on production for https://rollbar.com/gradecraft/gradecraft-development/items/1482/?person_page=0&#traceback.

This is a result of merging histories where the destination changeset does not exist.

This adds a check to the `#merge` operation on `HistoryFilter` to ensure that the destination exists before merging in the changeset.

Fixes #1740

### DIFF
--- a/app/models/history_filter.rb
+++ b/app/models/history_filter.rb
@@ -48,7 +48,9 @@ class HistoryFilter
 
     from_versions.each_with_index do |history_item, index|
       history_item.changeset.each_pair do |key, value|
-        if !["created_at", "updated_at"].include?(key) && value.is_a?(Array)
+        if !["created_at", "updated_at"].include?(key) &&
+            value.is_a?(Array) &&
+            to_versions.size > index
           to_versions[index].changeset.merge!({ "#{key}" => value })
         end
       end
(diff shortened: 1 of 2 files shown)


app/models/history_filter.rb
```diff
@@ -48,7 +48,9 @@ def merge(options)
 
     from_versions.each_with_index do |history_item, index|
       history_item.changeset.each_pair do |key, value|
-        if !["created_at", "updated_at"].include?(key) && value.is_a?(Array)
+        if !["created_at", "updated_at"].include?(key) &&
+            value.is_a?(Array) &&
+            to_versions.size > index
```
zazaian: This all looks good to me :+1: Out of curiosity what do the `to_versions` and `from_versions` methods reference here?

jwright: The `from_versions` is the `Version` history (which contains `changesets`) for the "from" class and `to_versions` is the history for the "to" class.

PR #1742, opened 2016-03-09
Issue 1606: URL authentication for client-encrypted S3 Files

zazaian:
## Overview
The thrust of this PR is that it's impossible to access client-encrypted resources on S3 using the `#presigned_url` helper from `Aws::SDK v2` as Rails or some agent on our server has to decrypt the envelope stored alongside the resource on S3 prior to fetching the decrypted resource.

The desired outcome here is that a user who's requested a `SubmissionsExport` download will receive an email with a download link which will allow them to directly access the download from S3 without having to authenticate via GradeCraft. Because we can't use the `#presigned_url` method on AWS we need to implement some kind of very secure limited-access authentication service in order to download these files. Thus I've created the `SecureToken` model to handle these authentications.

### Encryption Protocol
`SecureToken` uses `scrypt` for authentication which, to date, is the safest cryptographic platform in existence and as of right now there are no reported instances of it having been cracked. Here's a chart demonstrating the amount of hardware that would be required to crack passwords encrypted with 'scrypt', 'bcrypt', 'PBKDF2', and 'MD5' algorithms:

image

For reference the `(time)` amounts in parenthesis are the maximum decryption durations being used in those scenarios in order hinder the effectiveness of brute-force attacks. We're using a maximum decryption time of 1 second which is the default of the `scrypt` gem for ruby (https://github.com/pbhogan/scrypt), and is ample.

The maximum password length demonstrated in the chart above is 80-characters long. We're using a url-safe 254-character base-64 password (190 bit) generated via ruby's `SecureRandom` library:

```
SecureRandom.urlsafe_base64(190)
```

To make matters a bit worse for attackers we're using 512-bit `scrypt` encryption, which produces a hex hash of between 1049 and 1050 characters. It's also worth noting that because `scrypt` is a 'one way' encryption algorithm even if an attacker had the 512-bit encrypted hash in their possession it would be impossible to decrypt the original password used. The only way to authenticate against the encrypted hash is to guess the correct, random 254 character password that was used to generate it in the first place. 

### Workflow
The overall authentication process here is as such:

1. User requests a `SubmissionsExport` archive.

2. If the archive generates successfully, the `SubmissionsExportPerformer` generates a new `SecureToken` object which contains a hashed 190-bit password and a unique 36-character uuid (also generated by `SecureRandom`) which can used to locate the `SecureToken` so we don't have to betray the resource name or id in the secret url emailed to the user.

3. The user receives an email with a link to the resource which contains the `uuid` for the `SecureToken` as well as the 190-bit password (called `secret_key` across the remainder implementation).

4. The user clicks the link, which directs the request to the `SubmissionsExportsController#secure_download` method, which has been stripped of all other normal system authentication and replaced only with authentication from `SecureToken`.

5. The `SecureTokenAuthenticator` class accepts the `uuid` and the `secret_key` from the request and authenticates the secure token against them.

6. If the `secret_key` successfully authenticates against the `encrypted_key` in the `SecureToken` associated with the given `uuid`, then the request is authenticated, and the decrypted S3 resource is streamed via `send_file` in the `SubmissionsExport` controller. If the authentication fails because the `SecureToken` has expired (secure tokens are currently only valid for one week), then we redirect the user to the GradeCraft homepage and notify them of this. Otherwise they're just redirected to the homepage and given a vague 'your request was not successful' notice.

### Slowdown
It should also be noted that the usage of incorrectly formatted `uuid` or `secret_key` values submitted in requests also prompt a one-second slowdown of the response in order to hinder brute-force attacks that might be using random values for either of these attributes.

### Notes
Currently all `SecureToken` objects are being created with an `expired_at` time of one week from the created_at date. It may be desirable to modify this in the future. 

### SecureTokenHelper
There's a new `SecureTokenHelper` which has been tasked with building the `secure_download_url` that we're using in emails. This prevents us from having to build the url_for call in every email and cleans up the whole process of generating this `url` without having to do it in the `SecureToken` itself, which wouldn't really be appropriate.

### New REGEX Constant
Universal regexes for the `UUID`, `190_BIT_SECRET_KEY` and `512_BIT_ENCRYPTED_KEY` have been added to a `REGEX` constant hash in `/config/initializers/regex.rb`. This seemed like a good approach though it forced me to use the `rails_spec_helper` in a couple of specs that otherwise wouldn't have required them. Any thoughts on whether this is a good approach or whether there might be a more flexible way to define these would be welcome.

### S3Manager::Resource
A bunch of S3-related logic was moved out of `SubmissionsExport` and into the new `S3Manager::Resource` module. `SubmissionsExport` is currently the only model using this functionality but it seemed like it really belonged in the S3Manager vendor library rather than in the model. There are a couple of methods there currently that are handling encrypted objects but the method names don't explicitly suggest that they do. These should probably be clarified before this PR or in a future issue.

### Fixed sporadically-failing PredictorEventJob spec example in GradesController
There's a pretty annoying spec example in the `GradesController` spec that's been popping up for sometime because it expressly is testing context for when an error has been raised, and occasionally for whatever reason the rescue block doesn't catch the artificially-raised `RSpec` error and fails the example. I noticed that this was connected to a bunch of logic in the `GradesController` that was basically performing mongo-fallback as we implemented recently in `EventLogger`, so I added an `#enqueue_with_fallback` method to `ResqueJob::Base` and now this is handling the fallback rather than the controller dealing with it.

This allowed me to clean up about 70 lines of unnecessary spec examples from the `GradesController` spec and test only that a PredictorEventJob is being built and enqueued with fallback. I noticed that this `PredictorEventJob` is really doing the exact same thing that the `PredictorEventLogger` is but with a different set of parameters, and it seems that at some point in the near future it would be best to unify all of these through the `PredictorEventLogger` class as it has better functionality for building the default request parameters. For now, at least, this is much better than what was in place and the matter of the sporadically-failing spec has been resolved.

## Concerns
Really the only concern that I have with this PR at this point is that it would be more secure overall to extract SecureToken and the SecureTokenAuthenticator into a private repository and to include it from there. Because `scrypt` is so incredibly secure and we're using 512-bit encryption with a 190-bit password there's really no universe in which I think this is a threat, but it still seems like it would be better overall not to broadcast our authentication workflow even though GradeCraft is an open source codebase.

Additionally it's likely unnecessary to pass the `target_id` or even the `target_class` into the `SecureTokenAuthenticator` considering that the `uuid` for any given `SecureToken` is unique. I simply used these to add an additional security requirement for the authenticator to be able to find the `SecureToken` in the first place, though this is probably needless and may prove to be overkill if SecureToken is extended to other parts of the application.


zazaian:
@chcholman please note that this PR has been submitted.

zazaian:
@jwright this PR is pending your review whenever you get an opportunity. Thanks! :+1:

app/authenticators/secure_token_authenticator.rb
```diff
@@ -0,0 +1,76 @@
+class SecureTokenAuthenticator
+  def initialize(options = {})
```
jwright: Instead of using `required_options`, we could use keyword arguments here instead for the same functionality (https://robots.thoughtbot.com/ruby-2-keyword-arguments)

app/authenticators/secure_token_authenticator.rb
```diff
@@ -0,0 +1,76 @@
...
+    @options = default_options.merge options
+    @slowdown_duration = @options[:slowdown_duration]
```
jwright: Instead of using `default_options` here (since there is only one of them), maybe we should use a `Hash#fetch` with a default specified there. Seems like it would be easier to read.

app/authenticators/secure_token_authenticator.rb
```diff
@@ -0,0 +1,76 @@
...
+  end
+
+  attr_accessor :secure_token_uuid, :target_class, :target_id, :secret_key,
+    :slowdown_duration
+
+  attr_reader :options
+
+  # we should probably add a separate workflow here for just telling the user
+  # if the secure token expired so they know that this is happening instead of
+  #
```
jwright: I agree with this comment but it looks like you are already doing this with `valid_token_expired?`. I think the client can just check the `valid_token_expired?` first, then `authenticates?`.

app/authenticators/secure_token_authenticator.rb
```diff
@@ -0,0 +1,76 @@
...
+  def authenticates?
+    required_options_present? &&
+    uuid_format_valid? &&
+    secure_key_format_valid? &&
+    secure_token &&
+    !secure_token.expired? &&
```
jwright: I think we should replace the check `secure_token && !secure_token.expired?` with a check to `!valid_token_expired?` to eliminate some duplication.

app/authenticators/secure_token_authenticator.rb
```diff
@@ -0,0 +1,76 @@
...
+    secure_token.authenticates_with?(secret_key)
+  end
+
+  def valid_token_expired?
+    return unless secure_token
```
jwright: We should return false here.

app/authenticators/secure_token_authenticator.rb
```diff
@@ -0,0 +1,76 @@
...
+
+  def uuid_format_valid?
+    if secure_token_uuid.match REGEX["UUID"]
+      true
+    else
+      # This should only occur if somebody is attempting to crack a key by
+      # searching for random uuid values. By sleeping for a moment here we
+      # drastically hinder attacks by slowing down the rate at which they can
+      # be made but we do so without sacrificing system resources.
+      sleep slowdown_duration
```
jwright: :+1:

app/controllers/submissions_exports_controller.rb
```diff
@@ -23,13 +27,45 @@ def destroy
   end
 
   def download
-    send_data submissions_export.fetch_object_from_s3.body.read, filename: submissions_export.export_filename
+    stream_file_from_s3
+  end
+
+  def secure_download
+    if secure_download_authenticator.authenticates?
+      stream_file_from_s3
```
jwright: Is the desired behavior here to download the file and then always redirect to the root_url? Since there is no `and return` here, that will be the desired behavior. I think that is ok because they will be redirected to the Dashboard if they are signed in.

app/controllers/submissions_exports_controller.rb
```diff
@@ -23,13 +27,45 @@ def destroy
...
+      redirect_to root_url
+    end
   end
 
   protected
 
-  def delete_s3_object
-    @delete_s3_object ||= submissions_export.delete_object_from_s3
+  def secure_download_authenticator
+    # it's possible that this could be cleaned up by simply passing params into
```
jwright: I think this is fine.

app/mailers/exports_mailer.rb
```diff
@@ -53,6 +57,11 @@ def default_attrs
     }
   end
 
+  def cache_success_mailer_attrs(submissions_export, secure_token)
```
jwright: I don't think this method buys you much since you have to pass in the parameters. I think this reduces the readability a bit as well. Thoughts?

app/models/secure_token.rb
```diff
@@ -0,0 +1,95 @@
...
+  end
+
+  validates :uuid, uniqueness: true, format: { with: REGEX["UUID"] }
+  validates :encrypted_key, uniqueness: true, format: {
+    with: REGEX["512_BIT_ENCRYPTED_KEY"]
+  }
+  validates :target, presence: true
+
+  # double-check to make sure that the authentication process is correct here
+  def authenticates_with?(secret_key)
```
jwright: I really like the `SCrypt::Password`s approach for checking for plain text here. I think it would be neat to use some syntactic sugar here and be able to say `secure_token == "plain text"`. We could easily accomplish that with a `==` override that would make a call to `authenticates_with?`. Thoughts?

I am not in love with the name `authenticates_with?`. Maybe `matches?`?

app/models/secure_token.rb
```diff
@@ -0,0 +1,95 @@
...
+    # Additionally since the documentation for SCrypt is somewhat lacking
+    # it's worth noting here that the #scrypt_options hash only need to be
+    # passed in with SCrypt::Password.create, and not when building the new
+    # SCrypt::Password object below to perform the equivalency/authentication
+    # check.
+    #
+    SCrypt::Password.new(encrypted_key) == secret_key
+  end
+
+  def has_target_of_class?(required_class)
```
jwright: I don't see this used anywhere.

app/models/secure_token.rb
```diff
@@ -0,0 +1,95 @@
...
+    target && target.class.to_s == required_class.to_s
+  end
+
+  def expired?
+    expires_at <= Time.now
```
jwright: Do we want to convert these to utc? Do we need to worry about timezones here?

app/models/submission_file.rb
```diff
@@ -5,6 +5,7 @@ class SubmissionFile < ActiveRecord::Base
   attr_accessible :file, :filename, :filepath, :submission_id, :file_missing, :last_confirmed_at
 
   belongs_to :submission
+  has_many :secure_tokens, as: :tokenable
```
jwright: 1. Should this be `has_one`? When will we have multiple tokens for a single submission?
2. Do we want the `dependent: :destroy` option on this?

spec/authenticators/secure_token_authenticator_spec.rb
```diff
@@ -0,0 +1,319 @@
...
+    before do
+      allow(subject).to receive(:slowdown_duration) { slowdown_duration }
+    end
+
+    describe "#uuid_format_valid?" do
+      let(:result) { subject.uuid_format_valid? }
+      let!(:cached_uuid_regex) { REGEX["UUID"] }
+
+      before do
+        REGEX["UUID"] = /VALID-UUID/
```
jwright: ...and by double, I mean stub it out.

spec/factories/secure_token_factory.rb
```diff
@@ -0,0 +1,6 @@
+FactoryGirl.define do
+  factory :secure_token do
+    association :target, factory: :submissions_export
```
jwright: I wonder since `target` is polymorphic if we use a `trait` here.

```
factory :secure_token do
  trait :submissions_export do
    association :target, factory: :submissions_export
  end
end

## in tests...
let(:secure_token) { create :secure_token, :submissions_export }
```

spec/models/submissions_export_spec.rb
```diff
@@ -0,0 +1,214 @@
...
+
+    let(:expected_base_s3_key) do
+      "exports/courses/40/assignments/50" \
+      "/#{subject.created_at_date}" \
+      "/#{subject.created_at_in_microseconds}/stuff.zip"
+    end
+
+    before(:each) do
+      allow(subject).to receive_messages(course_id: 40, assignment_id: 50)
+      ENV["AWS_S3_DEVELOPER_TAG"] = "jeff-moses"
```
jwright: Is this ok that we do not set this back or use a stub here for the constant?

spec/models/submissions_export_spec.rb
```diff
@@ -0,0 +1,214 @@
...
+    end
+
+    context "env is development" do
+      before do
+        allow(Rails).to receive(:env) { ActiveSupport::StringInquirer.new("development") }
```
jwright: We do this several times in the code. It would be nice to pull this out into `spec/support`.

jwright:
@zazaian This PR is amazingly simple to follow and really great. This is really, really great code. Kudos!

I added many comments but they were pretty minor and some were questions.

As far as your concerns, I don't think it matters if it's a private gem or not. I don't think the implementation makes it less secure in a major way and having a private gem has it's own repercussions on cost, etc.

app/authenticators/secure_token_authenticator.rb
zazaian: Hey @jwright: I made a pass at this using keyword arguments instead of an `options` hash by moving the attr/ivar definition into a new `KeywordArguments::DefineAsIvars` module that I've written which sets ivars for each keyword argument.

I'm not sure if this is more readable, but I'm open to following this approach further if you think it makes sense:

```
# /app/authenticators/secure_token_authenticator.rb
class SecureTokenAuthenticator
  def initialize(secure_token_uuid:, target_class:, target_id:, secret_key:,
                 slowdown_duration: 1)
    # set each keyword argument as an instance variable
    include KeywordArguments::DefineAsIvars
  end
  ...
```

```
# /lib/required_arguments/define_as_ivars.rb
module KeywordArguments
  # This will define all keyword arguments passed into the containing
  # __method__ as ivars. This allows the keyword arguments in the target method
  # to change without having to worry about explicitly changing which attributes
  # are being defined.
  #
  module DefineAsIvars
    method(__method__).parameters.each do |type, key|
      next unless type == :key
      value = eval(key.to_s)
      instance_variable_set("@#{key}", value) unless value.nil?
    end
  end
end
```
Thoughts?

app/authenticators/secure_token_authenticator.rb
zazaian: It seems like this is solved by using required arguments instead of options, and a keyword argument for the `:slowdown_duration`. Additionally we could still pass an options hash in instead of using a slowdown_duration keyword argument, but since `slowdown_duration` is the only option at the moment that might be unnecessary.

app/authenticators/secure_token_authenticator.rb
zazaian: The problem with this change is that if `secure_token` is nil and the `valid_token_expired?` returns nil in that case then `!valid_token_expired?` would actually evaluate to true, which isn't what we want.

That said, I believe we still want the `valid_token_expired?` method to stay as it is as it's being used externally by the `SubmissionsExportsController`.

app/authenticators/secure_token_authenticator.rb
zazaian: :+1: cool, this makes sense.

app/controllers/grades_controller.rb
```diff
@@ -349,7 +349,11 @@ def predict_score
 
     @grade_saved = @grade.nil? ? nil : @grade.save
 
-    enqueue_predictor_event_job
+    # this should be implemented with a PredictorEventLogger instead of a
```
zazaian: :+1: good call.

app/controllers/grades_controller.rb
```diff
@@ -401,19 +405,6 @@ def safe_grade_possible_points
     @grade.point_total rescue nil
   end
 
-  def enqueue_predictor_event_job
```
zazaian: Cool, thanks!  :hamburger:

app/controllers/submissions_exports_controller.rb
zazaian: Hey there @jwright: yes indeed, I should have added a comment here to indicate that fact. Since that's the case though I can probably move `redirect_to root_url` to the bottom of the `#secure_download` method here to expressly redirect to root_url. Does that seem like a good solution?

app/controllers/submissions_exports_controller.rb
zazaian: Cool, thanks.

app/helpers/secure_token_helper.rb
```diff
@@ -0,0 +1,18 @@
+module SecureTokenHelper
+  def secure_download_url(secure_token)
+    # we've enforced presence of SecureToken[:target_type] so this shouldn't be
+    # able to be a nil value. Let's presume that this will work rather than
+    # assume that it might not.
+    target_type = secure_token.target_type.underscore
```
zazaian: Cool, sounds good.

app/models/secure_token.rb
zazaian: I think I was using this in a version of the `SecureTokenAuthenticator` and ripped it out. Good eye :+1:

app/models/secure_token.rb
zazaian: I'm a little torn about using the `==` syntactic sugar here as it feels a little magical to me. Particularly distressing is that if you reverse the order of operations it breaks, which isn't immediately obvious:
```
secret_key = "abc7f70378gcbb29" # the correct secret key
SCrypt::Password.new(encrypted_key) == secret_key
=> true
secret_key == SCrypt::Password.new(encrypted_key)
=> false
```
Maybe I'm overreacting here but that feels really strange to me. I'd rather require the usage of a method that exists only on `SecureToken` so that the direction of the workflow is more obvious. What about something like:
```
secure_token.secret_key?(secret_key)
# or
secure_token.unlocks_with?(secret_key)
```

app/models/secure_token.rb
```diff
@@ -0,0 +1,95 @@
...
+  def random_secret_key
+    # generates a 254-character URL-safe secret key
+    @random_secret_key ||= SecureRandom.urlsafe_base64(190)
+  end
+
+  protected
+
+  # let's expire the token in a week
+  def set_expires_at
+    self[:expires_at] = Time.now + 7.days
```
zazaian: Yeah, I think that makes sense. I started using this style since `rubocop` was complaining about it at some point, but using self.method = seems fine to me now.

app/models/submission_file.rb
zazaian: I actually think this is an oversight, I don't believe that we need to use `secure_tokens` in `SubmissionFile` right now. I'm going to remove this for the time being.

app/models/submissions_export.rb
```diff
@@ -41,75 +48,25 @@ def s3_object_key_pieces(object_filename)
...
   end
 
-  def s3_manager
-    @s3_manager ||= S3Manager::Manager.new
-  end
-
   def downloadable?
     !!last_export_completed_at
   end
-
```
zazaian: :bowling: because there's now :bowing:

config/initializers/regex.rb
```diff
@@ -0,0 +1,11 @@
+REGEX = {
```
zazaian: Yeah, I think that makes sense. I'll add a `SecureTokenValidator` with these values. Thanks for the suggestion.

zazaian: Hey @jwright, how does this direction look to you?
```
# /app/validators/secure_token_validator.rb
require_relative "secure_token_validator/regex"

class SecureTokenValidator < ActiveModel::Validator
  attr_accessor :record

  def validate(record)
    self.record = record
    validate_uuid_format
    validate_encrypted_key_format
  end

  def validate_uuid_format
    return if record.uuid =~ Regex.uuid
    record.errors[:uuid] << "UUID format is not valid."
  end

  def validate_encrypted_key_format
    return if record.encrypted_key =~ Regex.encrypted_key
    record.errors[:encrypted_key] << "Encrypted key format is not valid."
  end
end
```

```
# /app/validators/secure_token_validator/regex.rb
class SecureTokenValidator
  # define the regex formatting for the various secure_token parameters
  module Regex
    class << self
      def uuid
        # this is the regex for testing UUID formatting
        /\A[a-z0-9]{8}-[a-z0-9]{4}-[a-z0-9]{4}-[a-z0-9]{4}-[a-z0-9]{12}\z/
      end

      def secret_key
        # regex for testing formatting of 190-bit secure keys, 254 characters
        /\A[a-zA-Z0-9_\-]{254}\z/
      end

      def encrypted_key
        # regex for testing formatting of 512-bit secure encrypted keys,
        # 1049 or 1050 characters
        /\A[a-f0-9\$]{1049,1050}\z/
      end
    end
  end
end
```

I considered using an `EachValidator` here but it feels a lot less expressive to me since these would end just feeling kind of redundant and magical:
```
```
validates :uuid, :presence => true, :uuid => true
validates :encrypted_key, :presence => true, :encrypted_key => true
```

spec/authenticators/secure_token_authenticator_spec.rb
zazaian: Cool, that seems like a good improvement.

spec/factories/secure_token_factory.rb
zazaian: Cool, this seems like a good improvement. :+1:

spec/models/submissions_export_spec.rb
zazaian: Sorry, I've already implemented this in `UniMock::Rails.stub_env`. I'll be sure to implement that here.

spec/models/submissions_export_spec.rb
zazaian: I think it's fine but I'll be sure to cache it and reset it just in case.

zazaian:
Hey @jwright, I've gone through all of your notes here and bounced them back to you. I haven't updated the specs yet as I'd like to have decision on final implementational direction with these before I move ahead and get the whole build passing, but hopefully this is demonstrative enough for now to give you a sense of what I'm thinking for some of these. Thanks for the help and the feedback :+1:

app/controllers/submissions_exports_controller.rb
jwright: It does and I think there is now :)

zazaian:
@chcholman this is ready to merge. Codeclimate is failing but it looks like it has something to do with being unable to correctly pull the `3572387a9ea272fe7dc34fc730a58fb51b1f6272` HEAD commit from master. I ran the rubocop checks locally with `rubocop-git 3572387 HEAD` and everything is passing.

chcholman:
@zazaian Thanks so much Mike!

chcholman:
@zazaian Leaving the branch open so that you can move the text from the PR description to the Wiki!

zazaian:
Awesome, sounds good!

PR #1758, opened 2016-03-12
Standardizing email templates

chcholman:
This PR refactors mailers so that they rely on a standard pattern of partials for the template content. 

It also fixes a bug where the mailer.css stylesheet was not being applied to HTML emails, and adds/fixes email previews to make them available to investigate at the /emails route.


app/mailer_previews/activation_needed_email_preview.rb
```diff
@@ -1,6 +1,7 @@
 class ActivationNeededEmailPreview
   def activation_needed_email
     user = User.last
+    user.activation_token = 102
```
zazaian: Is `102` an arbitrary value here? Maybe adding a comment above this line would help to clarify why this is being set.

app/mailer_previews/gradebook_export_preview.rb
```diff
@@ -1,5 +1,9 @@
 class GradebookExportPreview
   def gradebook_export
-    NotificationMailer.gradebook_export
+    course = Course.first
+    user = User.first
+    export_type = "gradebook export"
+    csv_data = course.assignments.to_csv
+    ExportsMailer.gradebook_export course, user, export_type, csv_data
```
zazaian: Perhaps rather than setting a bunch of variables and then passing the variables into the `.gradebook_export` method we could allow `.gradebook_export` to take a hash of options?

```
ExportsMailer.gradebook_export({
  course: Course.first,
  user: User.first,
  export_type: "gradebook export",
  csv_data: course.assignments.to_csv
)}
```

app/mailers/exports_mailer.rb
```diff
@@ -27,8 +27,35 @@ def team_submissions_export_failure(professor, assignment, team)
     mail_team_submissions_export("failed to build", professor, assignment, team)
   end
 
+  def grade_export(course, user, csv_data)
+    set_export_ivars(course, user)
```
zazaian: I think it's probably fine to just set the ivars for `@course` and `@user` directly in these methods rather than relying on `set_export_ivars`. The only reason I'd condensed these into various `set_ivar` methods is that `rubocop` was complaining about the duplication of the same ivars being set with the same values in multiple places, but I presume that that would be unlikely now with more relaxed rubocop parameters. Thoughts?
```
def grade_export(course, user, csv_data)
  @course = course
  @user = user
  ...
```

app/mailers/exports_mailer.rb
```diff
@@ -27,8 +27,35 @@ def team_submissions_export_failure(professor, assignment, team)
...
+    attachments["grade_export_#{course.id}.csv"] = { mime_type: "text/csv",
```
zazaian: perhaps we could extract `{ mime_type: "text/csv", content: csv_data }` into a separate method since it's being used reiteratively. Thoughts?

```
def csv_attachment(content)
  { mime_type: "text/csv", content: content }
end
```

app/mailers/notification_mailer.rb
```diff
@@ -53,6 +35,32 @@ def earned_badge_awarded(earned_badge_id)
     send_student_email "#{@course.courseno} - You've earned a new #{@course.badge_term}!"
   end
 
+  def group_status_updated(group_id)
+    @group = Group.find group_id
```
zazaian: Maybe it would make more sense to just reference the attributes of the `@group` within the email template rather than setting all of these instance variables?

```
@group = Group.find group_id
@group.students.each do |group_member|
  mail(to: group_member.email, subject: "#{@group.course.courseno} - Group #{@group.approved}") do |format|
    @student = group_member
    format.text
    format.html
  end
end
```

app/views/layouts/mailers/announcement_layout.html.haml
```diff
@@ -15,35 +15,9 @@
           / content
           .content
             %table
+              %tr= render "layouts/mailers/logo"
```
zazaian: This looks great, much cleaner :+1:

app/views/user_mailer/welcome_email.html.haml
```diff
@@ -1,92 +1,9 @@
-<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
```
zazaian: :+1:

spec/mailers/exports_mailer/grade_export_spec.rb
```diff
@@ -65,5 +66,21 @@
         should_not include "Regents of The University of Michigan"
       end
     end
+
```
zazaian: This all looks good :+1:

zazaian:
Hey @chcholman I think this all looks good. I made a couple of comments on how we might be able to improve the workflow of some of the mailer methods, but these are kind of discretionary. Nice work!

app/mailer_previews/gradebook_export_preview.rb
chcholman: I'm going to skip this for now - both to keep the pattern with the rest of the mailer previews, and because switching to a hash breaks the view.

app/mailers/exports_mailer.rb
chcholman: :+1:

app/mailer_previews/activation_needed_email_preview.rb
chcholman: Comment added!

app/mailers/exports_mailer.rb
chcholman: I'm okay with how it currently is - I think rubocop's complaint here is a good note that we should keep thinking about how to abstract these further. There's still so much common between them we're repeating.

app/mailers/notification_mailer.rb
chcholman: :+1:

app/mailers/exports_mailer.rb
zazaian: Okay, we should decide on a standard format here because on the `1606` pr @jwright made a comment that this felt less readable to him. I'm really okay with it either way, but still agree that there's probably a better way to handle this duplication than either of these solutions.

chcholman: Totally agree

app/mailer_previews/gradebook_export_preview.rb
zazaian: Okay, at the very least then I'd say it's not really necessary to reiterate the local variable definitions. This could also be written as:
```
course = Course.first
ExportsMailer.gradebook_export(
  course,
  User.first,
  "gradebook export", # export type
  course.assignments.to_csv
)
```

zazaian: It's really fine either way, just trying to offer alternatives here.

chcholman: :+1:

PR #1760, opened 2016-03-13
Display correct dates for resubmissions

jwright:
This PR fixes several issues where the `Submission#updated_at` and `Grade#updated_at` dates were being displayed in the views, which was causing discrepancies when it came to resubmissions. The `Submission` was being listed in the `/resubmissions` view but the data looked like it was not a resubmission.

This was changed so that the `Submission#submitted_at` and the `Grade#graded_at` dates were being displayed instead.

In addition, I went through the views to display the `#submitted_at` date instead of the `updated_at` or `created_at` were being used  to signify when the `Submission` was submitted at.

One more thing....

I added a *RESUBMITTED?* column to the `/assignments/:id` view for the instructor. This now corresponds to the number of assignments with resubmissions listed in the `/resubmissions` view.

screen shot 2016-03-13 at 10 35 52 am

In order to achieve this, I simplified the calculation of the `Submission#resubmitted?` attribute. This used to use the `Resubmission` model to achieve whether it **was** a resubmission. However, we wanted to know whether it **is** a resubmission.

The difference is that `Submission#resubmitted?` shows if the `Submission` is **currently** a resubmission. It used to signify if it **was** a resubmission. This did not correspond to the `/resubmissions` view.

The cleanup meant that I could remove the `Resubmission` model since it was not longer being used. If we want to signify anywhere that the `Submission` **was** a resubmission, we can reintroduce the model.


app/models/submission.rb
```diff
@@ -73,16 +73,14 @@ def ungraded?
     !grade || grade.status == nil
   end
```
acousticrobot: I found [this line in the Wiki](https://github.com/UM-USElab/gradecraft-development/wiki/Submissions#user-content-instance-methods) really helpful to understand what this was used for: 

>> Used to report to the user that a change will be a resubmission.

Maybe worth adding as a comment above the method.

jwright:
@acousticrobot Thanks Jonathan! I added that comment (great idea).

chcholman:
@jwright After talking it through with @acousticrobot, we've got a UX change:

- Let's remove the "Resubmitted" column on the assignment page, and change the button text to "Re-grade". 
- Let's also add the '.danger' class to any button that is 'Grade' or 'Re-grade' (but not 'Edit Grade')

jwright:
@chcholman Here is what it looks like with "danger" grade buttons. Too much? I wanted to check before I put it everywhere.

screen shot 2016-03-16 at 12 23 57 pm

chcholman:
@jwright ...and one more iteration: 

- Let's only make the button red if there is a submission that needs to be graded. So the above screenshot, Katie Bell's would be red, and so would Brandon Angel's, but Regulas Black's would not.

jwright:
@chcholman Gotcha!

chcholman:
@jwright Thank you!

jwright:
screen shot 2016-03-16 at 1 22 39 pm

chcholman:
:+1: :+1: :+1:

PR #1761, opened 2016-03-13
Issue 1745: fix EventLogger workflow bug in Production

zazaian:
## Overview
The thrust of this PR is that there was an issue with the `LoginEventLogger` and `PageviewEventLogger` where the `base_attrs` from the `ApplicationEventLogger` were being overwritten by the `base_attrs` from `EventLogger::Enqueue` in the aforementioned subclasses.

In addition to moving the primary `base_attrs` definition from `ApplicationEventLogger` into `EventLogger::Base`, and removing `base_attrs` entirely from `EventLogger::Enqueue`, I've also added a new `application_attrs` method to the `ApplicationEventLogger` which is now used by the child event loggers instead of `base_attrs`. The only class that direcly inherits `base_attrs` is `ApplicationEventLogger`.

## Further Improvements
Beyond simply fixing the bug with the `EventLogger` inheritance workflow I've implemented changes from several of the recommendations made by @jwright from the PR on which this revised `EventLogger` code came into `master`.

### ApplicationController
* The `ApplicationController#record_course_login_event` method now accepts a course as an argument which will overwrite the `current_course` passed into the login event job from the `event_session` hash. This allows us to trigger a new login job elsewhere in any given controller without having to worry about whether the `current_course` session data has been updated.

Additionally this method now looks for `@user` as well as `current_user` in case the `current_user` hasn't been set by the session yet. If it seems desirable we could add a user argument to this method, but for the moment I think this works since users will change courses frequently but will never change user within the course of a given session without logging out first (if there is even a use case for this).

Also both of these methods are now `public`.

### ApplicationEventLogger
As mentioned the `#base_attrs` method here is now called `#application_attrs`, and an additional `#event_session_user_role` method has been added here to help handle circumstances in which `event_session[:user]` is nil or doesn't have a `#role` method. This seemed better than jamming all of the logic for handling these cases into `#application_attrs`.

### LoginEventLogger
The `LoginEventLogger` has been stripped way down and is now basically just serving to include the enqueuing methods from `EventLogger::Enqueue`, to establish the `:login_event_logger` queue, and most importantly to handle the login event through the new `LoginEventPerformer` class. 

Stripping out all of the class methods here help us to avoid any conflicts that may have arisen with setting class-level instance variables, which didn't appear to be happening (and would probably only occur under bizarre threading circumstances), but it seemed safer to move this to a dedicated class to help firm up the implementation here.

### PageviewEventLogger
Not much changed here except the use of `#application_attrs` instead of `#base_attrs`, as well as better workflow for the `#build_page_from_params`.

### PredictorEventLogger
The chief change here is that I ripped out the `EventLogger::Params` implementation for now to simplify the related specs so this is a bit more straightforward until we have time to visit a better implementation here.

@jwright's suggestion of using an `EventLogger::Param` class to wrap each of these param methods and do the filtering there probably makes the most sense, but there's really very little functionality here so just writing it out long-hand here really isn't too big of an issue for now.

### LoginEventPerformer
One of the most important improvements on this PR is the new `LoginEventPerformer`, which has basically inherited the previous workflow from the `LoginEventLogger` but with better encapsulation as an object rather than trying to perform this work entirely inside of the `LoginEventLogger#perform` block or with class methods and class-level instance variables.

Even though they're not technically in the same vendor library this class inherits from `ResqueJob::Performer` as it gives us some tools for determining whether the expected outcome occurred as well as some logging/messaging for those outcomes.

Until we have time to do some more work on the `Analytics` library, we're not attempting to create the login event without at `data[:user_role]` value as this is what was being used in the analytics aggregate classes which were throwing the repetitive error in NewRelic.

Like in the old workflow this performer finds the `CourseMembership` that matches the user_id and course_id from `data`, caches the `last_login_at` timestamp for use in the login event, and then updates the `CourseMembership[:last_login_at]`value with whenever the user logged in originally.

### EventLogger::Base
In order to ensure that there weren't any errors occurring due to class-level ivar threading conflicts I ripped out all of the class-level ivars that I could as @jwright and I discussed and replaced these all with class methods. Instead of these being passed on through `InheritableIvars` these methods are now just passed on through standard class inheritance to descendants.

### EventLogger::Enqueue
A couple of important changes were made here. First off, the `#fallback` methods now rescue only `Redis::BaseError` instead of `StandardError`, which means that only Redis errors will trigger fallback. 

Also the `.perform` behavior that was being called on fallback has been moved into a dedicated `#fallback` method to reduce duplication, and also to allow overwriting the `#fallback` behavior on subclasses without having to also re-write the rescue blocks.

### EventLogger::Params
As mentioned previously `EventLogger::Params` and its corresponding spec were removed entirely until it seems useful to build out a more straightforward implementation.

### ResqueJob::Outcome
I cleaned up some of the logic here but not too much. At some point this may be worth revisiting.

### ResqueJob::Performer
Performers now have accessible `#attrs` methods so `@attrs` doesn't have to be used as an ivar internally.

### ApplicationController EventLogger spec
I stripped out the shared examples that seemed like overkill here such as those for defining the `event_session` behaviors, and for not building `EventLoggers` unless `current_user` is present. Additionally I modified these to reflect some workflow changes such as the new `EventLogger::Enqueue#fallback` method.

### ApplicationEventLogger spec
I stripped out all of the shared examples I could here and just wrote out the spec long-hand to make them easier to read and manipulate. Instead of using shared examples to test behaviors we're now just checking to ensure that the class is inheriting from `EventLogger::Base` to ensure that its functionality is present even though it might not be tested directly.

### LoginEventLogger spec
Like the `LoginEventLogger` this has been stripped down substantially and now only tests that the class is inheriting from `ApplicationEventLogger`, as well as testing the conditions under which the new `LoginEventPerformer` is built and triggered with messaging.

### PageviewEventLogger and PredictorEventLogger specs
Liked with `ApplicationEventLogger` I stripped out all of the shared examples I could here and just wrote out the specs long-hand to make them easier to read and manipulate. Instead of using shared examples to test inherited behaviors we're now just checking to ensure that these inherit from the `ApplicationEventLogger`.

### EventLogger::Base spec
As with the specs above any unnecessary shared examples were replaced with normal examples.

### EventLogger::Enqueue spec
This was written out long-hand in lieu of using shared examples.


zazaian:
@jwright please review this when you have an opportunity. @chcholman please note that this PR has been submitted.

app/controllers/application_controller.rb
```diff
@@ -64,6 +64,25 @@ def redirect_back_or_default(path=root_path, options={})
     end
   end
 
+  # Tracking course logins
+  def record_course_login_event(login_course = nil)
+    return unless request.format.html? || request.format.xml?
+    return unless current_user || @user
+    event_attrs = event_session
```
jwright: Do we need to merge in `@user` here if that is being used? We check for `@user` but then we don't log it from what I can see.

app/event_loggers/application_event_logger.rb
```diff
@@ -22,14 +21,22 @@ def event_type
...
-      user_role: event_session[:user].role(event_session[:course]),
-      created_at: Time.zone.now
-    }
+      user_role: event_session_user_role(event_session)
+    }.merge(base_attrs)
+  end
+
+  def event_session_user_role(event_session)
+    return unless event_session[:user]
+    event_session[:user].role event_session[:course]
```
jwright: We probably want to return unless `event_session[:course]` is here too since `User#role` assumes that `course` is not nil.

jwright:
@zazaian This is much cleaner and easier to follow. Great job.

I just had a few comments on things I think I discovered.

Also, I am not sure if there are any extra "toolkits" lying around that we don't need but I just noticed that there was a lot of references removed from the specs and only a few deleted.

app/controllers/application_controller.rb
zazaian: Hey there @jwright: thanks for the note here, and good catch! I think because there are various scenarios in which we may actually want different attributes than the standard `event_session`, that it makes the most sense to just pass an `event_options` hash into the block and merge that into the `event_session` like so:

```
  # Tracking course logins
  def record_course_login_event(event_options = {})
    return unless request.format.html? || request.format.xml?
    event_attrs = event_session.merge event_options
    return unless [:course, :user].all? { |attr| event_attrs[attr].present? }
    LoginEventLogger.new(event_attrs).enqueue_with_fallback
  end
```

Thoughts?

app/event_loggers/application_event_logger.rb
zazaian: Cool, I think that makes sense. I left this as such because `some_user.role nil`will always return nil, but it's probably best to spare ourselves from the necessary query. :+1:

zazaian: Also I'm looking at the code here and `event_session`should probably be it's own class considering the sort of mess of logic occurring here. That would allow this whole thing to look like this:
```
  def application_attrs
    {
      course_id: event_session.course_id,
      user_id: event_session.user_id,
      student_id: event_session.student_id,
      user_role: event_session.user_role
    }.merge(base_attrs)
  end

  class EventSession
    def user_role
      return unless user and course
      user.role course
    end

     # some logic for handling the user, course, and student ids
  end
```

zazaian:
@jwright good eye on the toolkits, there were a couple that weren't being used, and a couple that still were, but I just ripped them all out for now. There's probably a better way to define some of these things but for now this is fine.

zazaian:
@chcholman this is probably ready to merge. @jwright had two minor comments and they've both been addressed. Build is rebased against `master` and passing all checks.

chcholman:
Awesome, thanks @zazaian!

zazaian:
Noooo problem @chcholman :)

PR #1794, opened 2016-03-21
Feature/1782 standard grade status

acousticrobot:
This is a small enhancement which will remove the possibility of grades submitted without a status. It makes the selection of Status explicit for all grades. If release is necessary the selector will present [In Progress, Graded, Released]. If no release is necessary, the Grader will still need to choose from [In Progress, or Graded], rather then defaulting to an invisible "Graded" value.  Additionally, the index of Grades will always show the Status, so that Grades without release will still display if they are "Graded" or "In Progress".

**The Good**: This standardizes the selectable grade Status with Rubric Graded assignments. Grades can't be submitted without a status. It also places the selector in the same location as the Rubric Grade form.

**The Bad**: This shares no code with the Rubric Grading logic, as it still relies on submitting a standard form, so it can't use the same check for a selected status on submit. Instead, the status bar has no "unselected" option. This may not be an acceptable solution since Graders who are not used to making this selection could end up with a lot of "In Progress" grades by default.

### DIFF
--- a/app/views/assignments/individual/_table_body.html.haml
+++ b/app/views/assignments/individual/_table_body.html.haml
@@ -41,8 +41,7 @@
         .display_on_hover.hover-style
           This feedback has been <strong>marked as read</strong> by the #{term_for :student}.
 
-    - if presenter.assignment.release_necessary?
-      %td= grade.status
+    %td= grade.status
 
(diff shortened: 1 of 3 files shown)


chcholman:
@acousticrobot Unfortunately that is exactly what will happen, and won't work as a solution.

PR #1795, opened 2016-03-22
Grade components

chcholman:
This PR removes the `individual_show` and `group_show` partials, and replaces them with a single show view that references a series of component partials. This will hopefully address scenarios where we've been inconsistent about showing people things on group grades vs. individual, and should make the views much easier to read and maintain.


app/controllers/grades_controller.rb
```diff
@@ -13,7 +13,6 @@ class GradesController < ApplicationController
 
   # GET /assignments/:assignment_id/grade?student_id=:id
   def show
-    @assignment = current_course.assignments.find(params[:assignment_id])
```
jwright: :+1:

jwright:
@chcholman This is awesome! The maintenance on this is greatly improved.

PR #1803, opened 2016-03-23
fixes autosave on standard grade edit

acousticrobot:
This fixes the autosave feature on Grade standard edit.  It introduces a new route `api/grades/:id` that uses strong params to update the grade attributes.

In order to not have submission history run a muck with the autoupdate, there is a history squisher that merges the history with the previous history version and deletes the previous history so they appear as one history item in the submission history.

history_squisher


jwright:
@acousticrobot Ready for your review!

jwright:
@acousticrobot Great catch! I believed these to be sent to the controller at the same time. I believe I need to check to make sure if they are not that it still works as intended.

chcholman:
Just noting - for overly cautious reasons, I'm going to wait until the end of semester to merge this.

PR #1814, opened 2016-03-23
1748 replace canable

### LINKED ISSUE #1748: Exchange Cannable for CanCan
We currently use [Canable](https://github.com/jnunemaker/canable) for authorizing requests in our application. We use this sparingly but that does explain why we have some many different authorization mechanisms for different resources.

This change will allow us to use the same authorization mechanism in our controller and routing everywhere. No more guessing how a resource determines it's authorization.

There are 3 reasons to make this change.

1. Canable has not been updated in 4 years. The Ruby community has made CanCan the default authorization framework over the years and Canable has been left behind.

2. Canable does not support what we use a context. Currently, GC is unique in that it uses 2 resources as the current context. It uses a `User` and a `Course`. Canable assumes that you place "cans" (what a user can do on a resource) on the `User` model. It does not allow multiple classes to have "cans" defined. CanCan on the other hand allows you to redefine how your "current ability" behaves. The default is to use `current_user` but it can be overridden in the controller mixin.

3. The architecture of CanCan is much better than Canable. In Canable, you specify abilities on the resources themselves. This expands the responsibilities of the models (they now need to act like authorization objects) and it reduces reusability. CanCan introduces a new object called an Ability that acts solely as the authorization object where you pass in the resource itself.

jwright:
This PR replaces [Canable](https://github.com/jnunemaker/canable) with [CanCan](https://github.com/CanCanCommunity/cancancan). This was done because Canable has not been upgraded for 4 years and it does not allow a custom context. The "Can" must be placed on an object that is serialized (in this case `User`).

CanCan allows for a custom context. This is needed since there is a `User` and a `Course` that needs to be checked for authorization. The default is to use just a `current_user` from a controller but this PR overrides `ApplicationController#current_ability` and sends in both the `current_user` and `current_course`.

The `Ability` object checks authorization rules for `Grade`, `Announcement`, `AssignmentWeight`, and `Submission`. These were all objects that used the previous Canable implementation.

The `Grade` authorization rules delegate to the `GradeProctor` object. In the future, proctor objects will be used for the `Announcement`, `AssignmentWeight`, and `Submission` objects.

The old Canable `enforce_view_permissions` was replaced with the CanCan `authorize` calls. In the future, all the authorization in the controller and view layers should be handled by CanCan (delegating to the proctors) and all of the authorization calls in the model layers and below should be handled by the proctor objects. The complete implementation of this is out of scope for this and will be handled in future refactorings.

Closes #1748


app/models/ability.rb
```diff
@@ -0,0 +1,19 @@
...
+  include GradeAbility
+  include SubmissionAbility
+
+  def initialize(user, course)
+    define_assignment_weight_abilities user, course
+    define_announcement_abilities user, course
+    define_grade_abilities user, course
+    define_submission_abilities user, course
+  end
+end
```
acousticrobot: This looks really extensible

PR #1823, opened 2016-03-27
Adding indexes

chcholman:
pghero has identified: 

Missing Indexes

```
These tables have a large number of rows but indexes are not used often. Add indexes for faster queries.

Table              % of Time Index Used    Rows in Table
submissions           1                               24,906
submission_files    17                              20,793
unlock_states     1                                   20,594
criterion_grades    8                                15,408
```

### DIFF
--- a/db/migrate/20160327132551_drop_shared_earned_badges.rb
+++ b/db/migrate/20160327132551_drop_shared_earned_badges.rb
@@ -1,0 +1,5 @@
+class DropSharedEarnedBadges < ActiveRecord::Migration
+  def change
+    drop_table :shared_earned_badges
+  end
+end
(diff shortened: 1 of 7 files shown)


db/migrate/20160327132551_drop_shared_earned_badges.rb
```diff
@@ -0,0 +1,5 @@
+class DropSharedEarnedBadges < ActiveRecord::Migration
+  def change
+    drop_table :shared_earned_badges
```
jwright: :+1:

db/migrate/20160327135043_add_indexes_to_unlock_states.rb
```diff
@@ -0,0 +1,6 @@
+class AddIndexesToUnlockStates < ActiveRecord::Migration
+  def change
+    add_index :unlock_states, [:unlockable_id]
```
jwright: Should we use both `unlockable_id` and `unlockable_type` here since it's polymorphic and ActiveRecord will look it up by id and type in most cases?

PR #1843, opened 2016-03-30
add predictor threshold icon and slider

acousticrobot:
This feature extends the threshold assignment functionality into the predictor. Assignments with a threshold will present the scale icon in the icon bar.

Additionally, the threshold will influence the behavior of the slider. Sliders are still draggable and influence the prediction, but until the threshold is met, the fill color is red. If the slider is released before meeting the threshold, it snaps back to zero, and a zero prediction is persisted through AJAX.

I have also added a sample assignment with both levels and a threshold, to verify these two behaviors will work together, if ever implemented on the same assignment.

### DIFF
--- a/app/assets/javascripts/angular/controllers/PredictorCtrl.js.coffee
+++ b/app/assets/javascripts/angular/controllers/PredictorCtrl.js.coffee
@@ -59,10 +59,15 @@
     else
       return true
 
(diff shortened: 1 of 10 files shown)


chcholman:
@acousticrobot This looks amazing!

PR #1849, opened 2016-03-31
1839 - Refactor submission scopes

### LINKED ISSUE #1839: Refactor queries on Grading Status & Ungraded Submissions pages

jwright:
This PR improves the performance of the `Submission.ungraded` scope.

### Implementation

The old scope was this:

```
where("NOT EXISTS(SELECT 1 FROM grades WHERE submission_id = submissions.id OR (assignment_id = submissions.assignment_id AND student_id = submissions.student_id) AND (status = ? OR status = ?))", "Graded", "Released")
```

The new scope is this:

```
  scope :with_grade, -> do
    joins("INNER JOIN grades ON "\
      "grades.group_id = submissions.group_id OR "\
      "(grades.assignment_id = submissions.assignment_id AND "\
      "grades.student_id = submissions.student_id)")
  end

  scope :ungraded, -> do
    includes(:assignment, :group, :student)
      .where.not(id: with_grade.where(grades: { status: ["Graded", "Released"] }))
  end
```

### Performance

This resulted in improved performance for the query runtime.

These times below are based on a course that has 2071 submissions and 15,054 grades.

The old run time for the scope was `7842.8ms`.

The new run time for the scope is `24.8ms`.

In addition, rendering the page used to take 4 queries for each row since the assignment, student, and team were lazy loaded.

This is what the old sql query breakdown looked like for rendering the ungraded submissions view:

screen_shot_2016-04-05_at_5_29_01_pm

This is what the new sql query breakdown looks like for rendering the ungraded submissions view. We cannot eager load the team retrieval so there is still some sql queries back to the server.

screen_shot_2016-04-05_at_5_29_35_pm

Closes #1839

### DIFF
--- a/app/models/submission.rb
+++ b/app/models/submission.rb
@@ -25,20 +25,18 @@ class Submission < ActiveRecord::Base
   has_many :submission_files, dependent: :destroy, autosave: true
   accepts_nested_attributes_for :submission_files
 
-  scope :ungraded2, -> do
-    # TODO: Add index on grades.submission_id
-
-    includes(:grade, :assignment)
-      .where("(assignments.release_necessary = true AND "\
-              "grades.status != :released) OR "\
-             "(assignments.release_necessary = false AND "\
-              "grades.status NOT IN (:graded, :released)) OR "\
-             "grades.id IS NULL", graded: "Graded",
-             released: "Released")
-      .references(:grade, :assignment)
+  scope :with_grade, -> do
+    joins("INNER JOIN grades ON "\
+      "grades.group_id = submissions.group_id OR "\
+      "(grades.assignment_id = submissions.assignment_id AND "\
+      "grades.student_id = submissions.student_id)")
   end
 
-  scope :ungraded, -> { where("NOT EXISTS(SELECT 1 FROM grades WHERE submission_id = submissions.id OR (assignment_id = submissions.assignment_id AND student_id = submissions.student_id) AND (status = ? OR status = ?))", "Graded", "Released") }
+  scope :ungraded, -> do
+    includes(:assignment, :group, :student)
+      .where.not(id: with_grade.where(grades: { status: ["Graded", "Released"] }))
+  end
+
   scope :resubmitted, -> { includes(:grade).where(grades: { status: ["Graded", "Released"] })
                                            .where("grades.graded_at < submitted_at") }
   scope :order_by_submitted, -> { order("submitted_at ASC") }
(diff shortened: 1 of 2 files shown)


jwright:
@acousticrobot @chcholman Can you guys take a quick gander at this early PR and let me know if you think this is more complete or am I missing something?

chcholman:
@jwright No, this not query is not more correct. Ungraded and Needs Release are different statuses - the Ungraded Submissions filter should *only* return items that have not been graded at all. 

What we really care about here is the workflow:
* Ungraded means items that have not yet been dealt with by any grader, and need to be assessed
* Graded-but-needs-release are items that have been assessed by a grader, but someone (usually the instructor) has not yet reviewed/normalized/adjusted, and is not yet visible to the student. 
* What isn't addressed here is "In Progress", where some particular grader is partway through the assessment process and needs to complete it. We need to build this feature out more explicitly.

jwright:
@chcholman Ok. Thanks.

So ungraded should be any submission that does not have a grade or has an In Progress grade. Is that correct? (That is what the last query looked for).

chcholman:
@jwright Doesn't have a grade isn't quite accurate at the moment (but it will be) - it should be any submission doesn't have a grade with a status

jwright:
I was unable to make this query any more performant with the current structure. 

This is the EXPLAIN query for the current query:

```
EXPLAIN SELECT id FROM submissions WHERE NOT EXISTS(SELECT 1
FROM grades WHERE group_id = submissions.group_id OR (assignment_id = submissions.assi
gnment_id AND student_id = submissions.student_id) AND (status = 'Graded' OR status =
'Released'));

#->  Nested Loop Anti Join  (cost=0.81..800.12 rows=60 width=4)
```

This is the EXPLAIN query for another query that uses a join instead:

```
EXPLAIN SELECT id FROM submissions WHERE id NOT IN (SELECT submissions.id FROM submissions INNER JOIN grades ON grades.group_id = submissions.group_id OR (grades.assignment_id = submissions.assignment_id AND grades.student_id = submissions.student_id) WHERE grades.status IN ('Graded', 'Released'));

#->  Seq Scan on submissions  (cost=799.85..805.90 rows=82 width=4)
```

I got the query to act more normal so it no longer depends on the number of grades but it is still not that good.

I do not believe that the query can be optimized based on the current structure. I think that the query can be optimized by placing the `grade_id` on the `submissions` table instead of having `submission_id` on the `grades` table.

This is a pretty large change but I can see what it can do to the query first and then go from there. Do you guys see any issues with this idea? Is this a bad route to go down?

chcholman:
@jwright I think that generally that makes a lot of sense. This will be slightly odd logically as long as predicted scores are still stored on grades, assuming that when you either create a new submission or create a new grade as a student you need to check for the other and store the id. (And there are various paths this workflow actually takes - realizing we may not have caught them all right now)

Would an alternative option be to store a boolean 'ungraded' state on submissions that is updated via a background job?

PR #1850, opened 2016-04-01
Updated gemfile to use froala

chcholman:
This PR updates the version of froala we use in order to address a variety of formatting issues we were running into that they have since fixed. It also relaxes the sanitization done to allow people to add color, tables, and images to their text.


app/assets/javascripts/froala/froala_license.js
```diff
@@ -1,2 +0,0 @@
-$.Editable.DEFAULTS.key = '6Ud1QBRVCDLPAZMBQ==';
```
acousticrobot: Is this an issued license?

Gemfile.lock
```diff
@@ -322,8 +322,8 @@ GEM
     fog-xml (0.1.2)
       fog-core
       nokogiri (~> 1.5, >= 1.5.11)
-    font-awesome-rails (4.4.0.0)
-      railties (>= 3.2, < 5.0)
+    font-awesome-rails (4.5.0.1)
```
acousticrobot: Do we need to lock any of these updates down in the Gemfile itself?

app/assets/javascripts/utilities.js
```diff
@@ -23,7 +23,8 @@ $( "#tabs" ).tabs({
   }
 });
 
- $('.froala').editable({
+ $('.froala').froalaEditor({
+  key: '6Ud1QBRVCDLPAZMBQ==',
```
acousticrobot: Is this an issued license?

Gemfile.lock
chcholman: This was actually purposeful - the newest version of froala takes advantage of some updated froala icons.

app/assets/javascripts/utilities.js
chcholman: Yes - if there's another way to do this I'd be happy to switch.

PR #1856, opened 2016-04-04
BadgeProctor and Invisible Level Badges

acousticrobot:
### BadgeProctor

This PR introduces a BadgeProctor. The badgeProctor takes the `viewable?` functionality of the GradeProctor and adapts it to viewing badges. It does not currently implement the `updatable?` functionality, pending discussion of CanCan and how we want to go about implementing permissions app wide.

It is implemented in this PR to determine the permissions for the new `Invisible Level Badges`. The existing places where the app could use the BadgeProctor to determine viewable permissions are not standardized and have some possible additional requirements, so I have left them for another PR.

One difference from the `GradeProctor` is that the `course` is handled from options params rather than as an optional variable

```
BadgeProctor.new(badge).viewable? some_user, course: some_course
GradeProctor.new(grade).viewable? some_user, some_course
```

I made this change since there is currently no place in the app where we need to determine if the badge is from a particular course, but there are additional factors relevant to the badge.  For instance, I am also currently handling `level` in options:

```
BadgeProctor.new(badge).viewable? some_user

BadgeProctor.new(badge).viewable? some_user, level: some_level
```

This probably warrants a discussion on how to better standardize the calls across Proctors, and I am open to other suggestions here for the BadgeProctor.

### Invisible Level Badges

Currently invisible badges were filtered out from the rubric design page. This PR allows invisible badges to be attached to Rubric Levels similar to the current visible level badges
  * Allows faculty to attach invisible badges to rubric levels
  * Allows faculty to view any invisible badges on a rubric
  * Allow students to see invisible badges they earned through a rubric
  * Prevents students from seeing earned invisible badges on other unearned levels
  * Normal behavior of earned invisible badges for students on the `my_badges` page

### Complete redo of sample badges
   * Ability to define and expand custom sample badge attributes
   * Sample Badges recreated based on available badge features
   * Awards sample badges consistently across students and only if set to be awarded
   * Awards sample badges multiple times only if this is possible for the badge
   * Adds a visible level badge to the top level of every rubric criteria
   * Adds an invisible level badge to the 2nd top level of every rubric criteria

### bugfixes
  * fixed spelling error preventing earning level badges
  * fixed method that is responsible for deleting all previous level badges for a rubric when grading

---
  closes 1832

  closes 1447

### DIFF
--- a/lib/grade_proctor/viewable.rb
+++ b/lib/grade_proctor/viewable.rb
@@ -1,14 +1,20 @@
-# determines if a `Grade` resource can be viewed within a specific
-# user and course context
+# Determines if a `Grade` resource can be viewed by a user. If no user is
+# supplied in the options, it will default to the Grade's student.
+#
+# Options include:
+#   course:  Will verify the grade is for the course
+#   user:    Determines permissions for supplied user rather than the
+#            grade's student
+#
 class GradeProctor
   module Viewable
     include Base
 
-    def viewable?(user=nil, course=nil)
+    def viewable?(options={})
       return false if grade.nil?
 
-      user ||= grade.student
-      course ||= grade.course
+      user = options[:user] || grade.student
+      course = options[:course] || grade.course
 
       grade_for_course?(course) &&
         (user.is_staff?(course) ||
--- a/app/assets/javascripts/angular/controllers/RubricCtrl.js.coffee
+++ b/app/assets/javascripts/angular/controllers/RubricCtrl.js.coffee
@@ -34,11 +34,6 @@
   $scope.pointsMissing = ()->
     $scope.pointsDifference() > 0 and $scope.pointsAssigned() > 0
 
-  # check whether points overview is hidden
-  # $scope.pointsOverviewHidden = ()->
-  #  element = angular.element(document.querySelector('#points-overview'))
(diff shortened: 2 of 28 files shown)


jwright:
@acousticrobot I am confused as to what "invisible badges" are. Can you explain those a bit?

jwright:
@acousticrobot I like the optional parameters. I think we should implement those for the course for the `GradeProctor` so the apis are the same. Thoughts?

jwright:
@acousticrobot I think we should do it as part of this PR

lib/grade_proctor/viewable.rb
```diff
@@ -4,11 +4,11 @@ class GradeProctor
   module Viewable
     include Base
 
-    def viewable?(user=nil, course=nil)
+    def viewable?(options={})
```
jwright: Do we want to add the comments here like we have with `updatable?`?

jwright:
@acousticrobot Looks great! I just had a small comment about comments

lib/grade_proctor/viewable.rb
acousticrobot: Yes we do! I'll get it in.

PR #1897, opened 2016-04-08
1200 meets expectations

acousticrobot:
## Feature: Meets Expectations

This PR adds `meets_expectations` to the Level model. This boolean is used to determine if which, if any, of the levels in a criteria has been marked as meeting expectations.  This does not affect the logic for the grading process in any other way, so the attribute mainly is used in the styling of rubrics, in order to indicate which level achievement is expected from an average student.

`meets_expectations` is set from the rubric design page, using Ajax. The http call is made from the `Level.js` factory, which ideally would be handled from a service, but this follows the pattern for deleting a level already in place on this page.

A new api/levels controller has been added to handle the level update though json.

One tricky thing for this to work is that the rest of the levels in the criterion must be false to insure only one level is marked as meeting expectation. Originally I was going to handle this in the `Level` model with a `before_save`, but this seemed like an anti-pattern, to have one instance of a class forcing changes on other instances. The solution I came up with was:

  * I added `Criterion#remove_expectations` which sets all of a criterion's levels to `meets_expectations = false`
  * This method is called from the `LevelController` for the `level.criterion` first, and then only this level is set to `meets_expectations = true`

My one hesitation is this leaves the onus on the controller to manage the uniqueness per criterion, so I'm open to suggestions for another solution.

### Styling for Meets Expectations

The rubrics are in line for a major overhaul, so I tried to only make changes that worked in concert with the current rubric styles.  I tried to use a "one shade lighter" approach, combined with text that would work for any accessibility issues.

#### Rubric Design Style Changes

  * `ME` levels are indicated with the lighter green, all other levels now use the darker green
  * In the lower right where we currently display "Zero Credit Level", "Full Credit Level", or "Score Level" the `ME` level now says "Meets Expectations" (I removed "Score Level" because it seemed useless)
  * There is click-able text next to the delete button that either says "Mark as Meets Expectations" or "Remove Meets Expectations" depending on the state of the Level.  
  * The delete button is now an off corner red circle
  * The description box text area is now fixed to conform to the bounding box 

screen shot 2016-04-08 at 10 18 38 am

#### Rubric Grading Style Changes

  * All `ME` levels say `Meets Expectations` below the level name 
  * Unselected levels are one shade lighter gray
  * Selected `ME` levels are one shade lighter green 
  * I removed the superfluous radio button to reduce clutter

screen shot 2016-04-08 at 9 18 32 am

#### Rubric Grade Show Page

  * All `ME` levels say `Meets Expectations` below the level name
  * All `ME` levels are one shade lighter gray

screen shot 2016-04-08 at 9 19 54 am

### rubric view components

Rubric views come in two flavors: Angular enhanced views for action pages, and plain old haml views for show pages. I DRYed up the POHVs, broke them into partials, and moved them into `views/rubrics/components/`. This should set the stage for making future changes to these views universal across the application, and easier to locate.

### removes `durable` from levels

This column was used to determine if a level was able to be deleted. Since the only levels that this applied to were either `full_credit` or `no_credit`, I updated the logic to use these fields instead, changing the condition to `required` in the js.

### adds better errors gem

Let me know if you anyone has a reason not to use it -- we use better errors on all of the other AJ projects and I find it reduces the number of times I have to pry in to an error.


app/assets/javascripts/angular/factories/Level.js.coffee
```diff
@@ -40,14 +41,11 @@
...
       @criterion.levels.splice(index,1)
-
-    ##grade rubric ctrl
-    # loadLevelBadge: (levelBadge)->
-    #   self = this
-    #   courseBadge = self.availableBadges[levelBadge.badge_id]
-    #   loadedBadge = new LevelBadge(self, angular.copy(courseBadge))
-    #   self.badges[courseBadge.id] = loadedBadge # add level badge to level
-    #   delete self.availableBadges[courseBadge.id] # remove badge from available badges on level
+    setCriterionExpectation: (index)->
```
jwright: `index` is not being used here

app/assets/javascripts/angular/factories/Level.js.coffee
```diff
@@ -120,13 +118,44 @@
...
+        "Zero Credit Level"
+      else if @fullCredit
+        "Full Credit Level"
+      else if @meetsExpectations
+        "Meets Expectations"
+      else
+        "" #used to say: "Score Level"
+
+    # TODO: move all API calls to a service
+    setAsMeetsExpectations: (index)->
```
jwright: `index` is not being used by `setCriterionExpectation` and so it can be removed from here as well

app/controllers/api/levels_controller.rb
```diff
@@ -0,0 +1,32 @@
+
+class API::LevelsController < ApplicationController
+  before_filter :ensure_staff?
+
+
```
jwright: RuboCop found an issue with this extra line.

app/controllers/api/levels_controller.rb
```diff
@@ -0,0 +1,32 @@
...
+  # PUT api/level/:id
+  def update
+    level = Level.find(params[:id])
+    level.criterion.remove_expectations if updating_meets_expectations?
+
+    if level.update_attributes(level_params)
+      render json: { message: "level successfully updated", success: true }
+    else
+      render json: {
+        errors: [{ detail: "failed to update level" }], success: false
```
jwright: Do we want to return the actual errors from the model here so the user can know what happened?

app/models/criterion.rb
```diff
@@ -28,6 +28,12 @@ def copy(attributes={})
 
   include DisplayHelpers
 
+  def remove_expectations
```
jwright: I think this should be a bang method since it is updating the data `remove_expectations!`

app/controllers/api/levels_controller.rb
```diff
@@ -0,0 +1,32 @@
+
+class API::LevelsController < ApplicationController
+  before_filter :ensure_staff?
+
+
+  # PUT api/level/:id
+  def update
+    level = Level.find(params[:id])
+    level.criterion.remove_expectations if updating_meets_expectations?
```
jwright: I think this is ok.

app/views/rubrics/components/_level_badges.haml
```diff
@@ -0,0 +1,6 @@
+.row.badge-row
+  - level.level_badges.each_with_index do |level_badge, index|
+    - if BadgeProctor.new(level_badge.badge).viewable?(current_user, level: level) && index < 2
```
jwright: I think we should use CanCan here at the view layer that just delegates to the `BadgeProctor`. I am not sure what we will need to do to pass in the level though.

```
- if can? :view, level_badge.level
```

db/samples.rb
```diff
@@ -313,6 +313,9 @@ def add_unlock_conditions(model, config, course_config)
                 level = criterion.levels.create! do |level|
                   level.name = "Level ##{m}"
                   level.points = criterion.max_points - (m * 1000)
+                  if m == 2
+                    level.meets_expectations = true
```
jwright: This can be simplified to `level.meets_expectations = m == 2`

spec/controllers/api/levels_controller_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+      let(:world) { World.create.with(:course, :student, :assignment, :rubric, :criterion) }
+      let(:level) { world.criterion.levels.last }
+      let(:params) do
+        { id: level.id, level: { meets_expectations: true }}
+      end
+
+      it "updates the level attributes" do
+        put :update, params, format: :json
+        level.reload
+        expect(level.meets_expectations).to be_truthy
```
jwright: `reload` returns itself so you can chain methods if you wish to put all on one line

```
expect(level.reload.meets_expectatons).to be_truthy
```

spec/controllers/api/levels_controller_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+      end
+
+      it "insures only one level per criterion meets expectations" do
+        l_1 = world.criterion.levels.first
+        l_1.update_attributes(meets_expectations: true)
+        put :update, params, format: :json
+        l_1.reload
+        expect(l_1.meets_expectations).to be_falsy
```
jwright: Same `reload` comment here. Also, does `be_falsy` work? Should it be `be_falsey`?

jwright:
@acousticrobot Really clean code here. I love the small controller actions and the new directory structure for the views. I just had a few small things.

In terms of the UI, should be change the `Remove Meets Expectations` to maybe just put a `(remove)` link next to the `Meets Expectations` at the bottom? I find it to be a bit weird in the location it's at. cc:\\ @mariehooper @chcholman

app/assets/javascripts/angular/factories/Level.js.coffee
acousticrobot: Good catch -- updated!

app/controllers/api/levels_controller.rb
acousticrobot: fixed

app/controllers/api/levels_controller.rb
acousticrobot: I would like to, but I recently realized that the errors returned from the other Angular services aren't always parsing correctly. I think we may need to add some better error reporting across all the http calls.

app/models/criterion.rb
acousticrobot: :+1: Good call!

app/views/rubrics/components/_level_badges.haml
acousticrobot: Good idea, adding to :musical_note: t-u-e-s-d-a-y  :notes: ​_chats!

db/samples.rb
acousticrobot: :cool: :sunglasses:

spec/controllers/api/levels_controller_spec.rb
acousticrobot: Good catch, changed. Both work, but we've standardized around `be_falsey` in GC. 

In js, I think `falsy` is more common:
[Mozilla's explanation:](https://developer.mozilla.org/en-US/docs/Glossary/Falsy)

>> Sometimes written falsey, although in English usually turning a word into an adjective with a -y, any final e is dropped (noise => noisy, ice => icy, shine => shiny)

mariehooper:
@acousticrobot Cait and I just reviewed the changes and have the below feedback...

1. Meets Expectations level and everything to the right should be green (even if a lighter shade/transparency of the GC green)
2. Use the lighter shade (currently on right) of gray for everything to the left of the Meets Expectation level
3. The text boxes should be white (unless they already are and our eyes are tricking us...)
4. Meets Expectations needs to look more like a clickable button... maybe a thin outline that fills in on hover like a ghost button
5. Badges button - no rounded corners on the bottom and maybe a lighter shade of the same color as the card itself. - striving to make these look more like buttons....

app/models/criterion.rb
```diff
@@ -5,8 +5,13 @@ class Criterion < ActiveRecord::Base
   has_many :levels, dependent: :destroy
 
   has_many :criterion_grades
-  belongs_to :full_credit_level, foreign_key: :full_credit_level_id, class_name: "Level"
-  attr_accessible :name, :max_points, :description, :order, :rubric_id, :full_credit_level_id
+  belongs_to :full_credit_level,
+    foreign_key: :full_credit_level_id, class_name: "Level"
```
jwright: I don't think we need to specify these `foreign_key` options here. I believe these are the defaults.

jwright: I would like to remove them if they are the default for less maintenance.

app/models/criterion.rb
```diff
@@ -28,20 +33,38 @@ def copy(attributes={})
...
-  def meet_expectations_points
-    levels.where(meets_expectations: true).first && levels.where(meets_expectations: true).first.points || 1000000
-  end
-
-  def remove_expectations!
-    levels.each do |level|
-      level.update_attributes(meets_expectations: false)
+  # Mangages a unique level per criteria that meets expectations, and
+  # stores the id and points on the Criteria for queries from other levels
+  def update_meets_expectations!(level, state)
```
jwright: `state` does not seem like a good parameter name here since it is too vague. Maybe `meets_expections`?

app/models/level.rb
```diff
@@ -9,6 +9,8 @@ class Level < ActiveRecord::Base
   validates :points, presence: true, numericality: {greater_than_or_equal_to: 0}
   validates :name, presence: true, length: { maximum: 30 }
 
+  before_save :update_criterion_meets_expectations_points
```
jwright: What if we remove this callback and override the setter for `points`? That way, we can remove a callback and the fields will all be correct while it is in "transition" until it is saved.

spec/models/level_spec.rb
```diff
@@ -3,13 +3,41 @@
...
-    it "copies the badges for the levels" do
-      level.save
-      level.badges.create! name: "Blah", course: create(:course)
-      expect(subject.badges.size).to eq 1
-      expect(subject.level_badges.map(&:level_id)).to eq [subject.id]
+    describe "#above_expectations?" do
+      it "is false if points are equal to or below expectations" do
+        subject.criterion.update(meets_expectations_points: subject.points)
+        expect(subject.above_expectations?).to be_falsey
+      end
```
jwright: Can we add a blank line here?

app/models/criterion.rb
acousticrobot: I believe it's because there is no actual `MeetsExpectationLevel` that I need the explicitly define a foreign key that maps to a `Level`. That being said I looked through the new code and nowhere do I call `criterion.meets_expectation_level`, I always call `criterion.meets_expectation_level_id`, so I'll remove it for now.

The `full_credit_level` foreign key was changed here because, you know, Rubocop 80 characters blah blah, so I don't want to mess with it now I'm not working with this code. But maybe we should look into seeing if it's used, and if not, removing this line too?

app/models/level.rb
acousticrobot: This is a great idea, and I think much more efficient too. I'll update it and test that it works.

spec/models/level_spec.rb
acousticrobot: fixed

chcholman:
@acousticrobot Is it purposeful that the button for "Set" only shows on hover initially, but then once you've selected it, if you remove it, it stays present on all of the levels (even on refresh)?

chcholman:
@acousticrobot My apologies, looks like it was just my browser. I should restart my computer more often!

PR #1906, opened 2016-04-11
Refactor/1902 grade with rubric

acousticrobot:
This PR standardizes the logic around presenting rubrics in the views and the predictor.  Previously we had a number of ways of determining if an assignment was actually rubric graded throughout the app. This logic has been now all been changed to use the existing assignment method `grade_with_rubric?`. Additionally the method `has_rubric?` has been removed.  

When using the presenter, we also now use `presenter.grade_with_rubric` to keep the naming consistent.

When passing this boolean to js, it is set to `is_rubric_graded` to fit with the names of existing booleans.

closes 1902

### DIFF
--- a/app/views/assignments/_show_student.haml
+++ b/app/views/assignments/_show_student.haml
@@ -8,17 +8,17 @@
   #tabs.ui-tabs.ui-widget
     %ul.ui-tabs-nav{role: "tablist"}
       %li
-        %a{ "href" => "#tab1"} Description & Downloads
-      - if presenter.rubric_designed? && !presenter.grades_available_for?(current_student) && ( !current_student || presenter.assignment.description_visible_for_student?(current_student) )
+        %a{ "href" => "#tab1"} Description &amp; Downloads
+      - if presenter.show_rubric_preview? current_student
         %li
           %a{ "href" => "#tab2"} Grading Rubric
-      - if presenter.has_submission_for?(current_student)
+      - if presenter.has_submission_for? current_student
         %li
           %a{ :href => "#tab3"} My Submission
         %li
           %a{ href: "#history" } My Submission History
 
-      - if presenter.grades_available_for?(current_student)
+      - if presenter.grades_available_for? current_student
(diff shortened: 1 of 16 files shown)


app/views/assignments/_show_student.haml
```diff
@@ -27,23 +27,23 @@
 
     #tabt1.ui-tabs-panel.ui-widget-content{role: "tabpanel"}
       #tab1.active
-        = render partial: "assignments/description_and_downloads", locals: { presenter: presenter }
-      - if presenter.rubric_designed? && !presenter.grades_available_for?(current_student) && ( !current_student || presenter.assignment.description_visible_for_student?(current_student) )
+        = render "assignments/description_and_downloads", presenter: presenter
```
jwright: This version of rendering assumes several things. For instance, passing in an AR object or a collection will have different outcomes. I feel like we should be explicit here and use `partial:` with the `locals` options.

Thoughts?

spec/presenters/assignment_presenter_spec.rb
```diff
@@ -96,25 +96,44 @@
     end
   end
 
-  describe "#rubric_designed?" do
-    let(:rubric) { double(:rubric, designed?: true) }
+  describe "#grade_with_rubric?" do
+    it "is not to be used if the assignment deosn't grade with a rubric" do
```
jwright: "deosn't" is spelled incorrectly here.

jwright:
Looks great. I just had a few small comments.

app/views/assignments/_show_student.haml
acousticrobot: I was trying to standardize around the most recent changes @chcholman made in `/app/views/grades/_grade_content.haml` but I don't have a strong preference, other than making sure we use the same format in all our views. I would like it if we are using presenters that they would manage safeguarding against unexpected outcomes as much as possible. @jwright, @chcholman, @zazaian thoughts on this?

PR #1907, opened 2016-04-12
1747 Grade Resource

### LINKED ISSUE #1747: Change GradesController to use Grade as the resource

jwright:
This PR is one of many (more to come) that moves some of the actions off of the `AssignmentsController` to various other controllers and adds restful routes.

#### Issues

1. Reduce the number of methods in the current grades controller
2. Create better user exposed routes (more restful routes)
3. Increase the organization of the controller files and their responsibilities

#### Changes

Here are several current routes surrounding the grades controller and what I propose we change them to.

##### Current

| Url                                                                                                 |  Controller 
| ------------------------------------------------------------------------------------ | ------------------------------- 
| `GET /assignments/:assignment_id/mass_grade`                        | `grades#mass_grade` 
| `PUT /assignments/:assignment_id/mass_grade`                        | `grades#mass_update`
| `GET /assignments/:assignment_id/group_grade`                       | `grades#group_edit`
| `PUT /assignments/:assignment_id/group_grade`                       | `grades#group_update`
| `GET /assignments/:assignment_id/export_grades`                    | `assignments#export_grades`
| `GET /assignments/:assignment_id/export_submissions`           | `assignments#export_submissions`
| `GET /assignments/:assignment_id/download_current_grades` | `assignments#download_current_grades`
| `GET /assignments/:assignment_id/criterion_grades_review`    | `assignments#criterion_grades_review`
| `GET /assignments/:assignment_id/export_team_submissions `| `assignments#export_team_submissions`
| `GET /assignments/:assignment_id/submissions_export `          | `assignments#submissions_export`
| `GET /assignments/:assignment_id/export_team_submissions ` | `assignments#export_team_submissions`
| `GET /assignments/:assignment_id/ grade_import `                   | `assignments#grade_import `
| `GET /assignments/:assignment_id/ rubric_grades_review `      | `assignments#rubric_grades_review`
| `PUT /assignments/:assignment_id/ update_rubrics `                 | `assignments#update_rubrics`

##### Changes

| Url                                                                                                 |  Controller 
| ------------------------------------------------------------------------------------ | ------------------------------- 
| `GET /assignments/:assignment_id/grades/mass_edit`               | `assignments/grades#mass_edit` 
| `PUT /assignments/:assignment_id/grades/mass_update`          | `assignments/grades#mass_update` 
| `GET /assignments/:assignment_id/groups/:group_id/grade`      | `assignments/groups#grade` 
| `PUT /assignments/:assignment_id/groups/:group_id/grades`    | `assignments/groups#graded` 
| `GET /assignments/:assignment_id/grades/export`                     | `assignments/grades#export` 
| `GET /assignments/:assignment_id/export_submissions`           | *NO LONGER USED*
| `GET /assignments/:assignment_id/grades/download`               | `assignments/grades#download` 
| `GET /assignments/:assignment_id/grades`                               | `assignments/grades#index`
| `GET /assignments/:assignment_id/export_team_submissions `| *NO LONGER USED*
| `GET /assignments/:assignment_id/submissions_export `          | *NO LONGER USED*
| `GET /assignments/:assignment_id/export_team_submissions ` | *NO LONGER USED*
| `GET /assignments/:assignment_id/ grade_import `                   | *NO LONGER USED*
| `GET /assignments/:assignment_id/ rubric_grades_review `      | *NO LONGER USED*
| `PUT /assignments/:assignment_id `                                          | `assignments#update`

Closes #1747


jwright:
@chcholman @acousticrobot @zazaian What are your thoughts on the above proposal?

app/controllers/assignments/groups_controller.rb
```diff
@@ -0,0 +1,43 @@
...
+    @assignment_score_levels = @assignment.assignment_score_levels
+
+    if @assignment.grade_with_rubric?
+      @rubric = @assignment.rubric
+      # This is sent to the Angular controlled submit button
+      @return_path = URI(request.referer).path
+    end
+  end
+
+  # PUT /assignments/:assignment_id/groups/:id/graded
```
acousticrobot: I find having the route as a comment like this very helpful, should we do this on all the new controller routes?

jwright: I can see how those would be helpful. I can add them for all the routes I added/changed.

jwright:
I added comments to the controllers and routes. I did not change the name of the `groups#graded` route because I could not think of a different name. I am open to suggestions.

PR #1942, opened 2016-04-21
Line length

chcholman:


chcholman:
@acousticrobot Not at all, it was bugging me too. Take a look at this update, and let me know what you think. I think I like it a little bit better, but I'm also happy to skip these changes. The samples folder has been added to rubocop's excluded list, but if you run it on the file specifically you'd still have to hear the complaints about the line length. Curious what you think!

PR #1943, opened 2016-04-26
Adds student badge index and show pages

acousticrobot:
### Additions

BadgeController#index page now handles the badge index for faculty and students,
as well as the faculty view of the students badge index.

Students now have access to the badge show page with faculty information 
removed. The page includes an index of the student's earned
badges for the badge.

The show page has a presenter, which takes the logic out of the controller for
differentiating a faculty view, student view, or faculty view of the student 
view.

The partial `_context_menu.haml` DRYs up the context menu for all badge pages.

### Subtractions

The `my_badges` route has been removed, which pulls badge resources entirely 
off of the `StudentController`.

### What this PR does not do:

A new `student_visible_earned_badges_for_badge` query is placed on the User,
for now, where all other current badge queries currently live. These should all 
be refactored together, and probably placed on either the Badge or EarnedBadge
model instead.

The faculty views for badge index and show have been broken into components, to
aid in adding student view contexts for these pages, but most of the the 
elements in the components themselves have not been refactored.

The new student view of the show page has almost no styling, as we are about to 
have a redesign for this page and I didn't want to spend too much time on it.


app/controllers/badges_controller.rb
```diff
@@ -161,4 +173,12 @@ def predictor_badge_data
   def find_badge
     @badge = current_course.badges.find(params[:id])
   end
+
+  def find_student
+    if current_student || params[:student_id]
```
jwright: `current_student` already does the check for `params[:student_id]` and so I think we can remove that check here.

app/controllers/badges_controller.rb
```diff
@@ -1,26 +1,33 @@
...
-    @earned_badges = @badge.earned_badges
-    @teams = current_course.teams
-    if params[:team_id].present?
-      @team = @teams.find_by(id: params[:team_id]) if params[:team_id]
-      students = current_course.students_being_graded_by_team(@team)
-    else
-      students = current_course.students_being_graded
-    end
-    @students = students
+    presenter = ShowBadgePresenter.new({
```
jwright: You can use `ShowBadgePresenter.build` here like so which will reduce the need for creating the `presenter` local:

```
render :show, ShowBadgePresenter.build({...})
```

app/controllers/badges_controller.rb
```diff
@@ -59,18 +68,21 @@ def destroy
     @badge.destroy
 
     respond_to do |format|
```
jwright: Can we remove the `respond_to` here since we are only responding to html?

app/controllers/badges_controller.rb
```diff
@@ -59,18 +68,21 @@ def destroy
...
-      format.html { redirect_to badges_path, notice: "#{@name} #{term_for :badge} successfully deleted" }
+      format.html {
+        redirect_to badges_path,
+        notice: "#{@name} #{term_for :badge} successfully deleted"
+      }
     end
   end
 
   def predict_times_earned
     @badge = current_course.badges.find(params[:badge_id])
```
jwright: We do not need this to be an instance variable since we are rendering `json` here without a view. I like to remove instance variables so that the developer knows they do not go beyond this scope.

app/controllers/badges_controller.rb
```diff
@@ -59,18 +68,21 @@ def destroy
...
-    @badgePrediction =
+    badge_prediction =
       PredictedEarnedBadge.where(student: current_student, badge: @badge).first
-    @badgePrediction.times_earned = params[:times_earned]
+    badge_prediction.times_earned = params[:times_earned]
 
     # save the prediction and cache the outcome
-    @prediction_saved = @badgePrediction.save
+    @prediction_saved = badge_prediction.save
```
jwright: We can change this to simply `badge_prediction.save` and check `badge_prediction.valid?` on line 80. That reduces complexity here.

app/controllers/students_controller.rb
```diff
@@ -87,17 +87,6 @@ def teams
     @team = current_student.team_for_course(current_course)
   end
 
-  def badges
```
jwright: 👍

app/models/user.rb
```diff
@@ -388,6 +388,13 @@ def unique_student_earned_badges(course)
       .references(:earned_badges)
   end
 
+  def student_visible_earned_badges_for_badge(badge)
+    @student_visible_earned_badges_for_badge ||= EarnedBadge
+      .where(badge: badge)
+      .where(student_id: self[:id])
```
jwright: Can we use `self.id` here instead of `self[:id]`?

app/views/badges/_all_students_earned_badges_table.haml
```diff
@@ -0,0 +1,41 @@
...
+        - if presenter.badge.is_unlockable?
+          %td
+            - if ! presenter.badge.is_unlocked_for_student?(student)
+              = link_to glyph(:lock) + "Unlock", manually_unlock_unlock_state_path(student_id: student.id, badge_id: presenter.badge.id), :method => :post, class: "button"
+        %td
+          .right
+            - presenter.badge.earned_badges.where(student_id: student.id).each do |badge|
+              %ul.button-bar
+                %li= link_to glyph(:star) + "Awarded #{badge.created_at.strftime("%b %d")}", badge_earned_badge_path(presenter.badge, badge.id), class: "button"
+                %li= link_to glyph(:edit) + "Edit", edit_badge_earned_badge_path(presenter.badge, badge.id), class: "button"
```
jwright: Can we remove `.id` from the path helper so it keeps it consistent?

app/views/badges/_buttons.haml
```diff
@@ -0,0 +1,5 @@
+%ul.button-bar
+  %li= link_to glyph(:check) + "Quick Award", mass_award_badge_path(badge.id), class: "button"
```
jwright: Can we remove .id from the path helper so it keeps it consistent?

app/views/badges/_buttons.haml
```diff
@@ -0,0 +1,5 @@
...
+  %li= link_to glyph(:star) + "Award", new_badge_earned_badge_path(badge), class: "button"
+  %li= link_to glyph(:edit) + "Edit", edit_badge_path(badge), class: "button"
+  %li= link_to glyph(:trash) + "Delete", badge_path(badge),class: "button", data: { confirm: "Are you sure?",  method: :delete }
```
jwright: Space after comma

app/views/badges/_context_menu.haml
```diff
@@ -0,0 +1,27 @@
+.context_menu
+  %ul
+    - if actions.include? :new
```
jwright: 👍

app/views/notification_mailer/earned_badge_awarded.html.haml
```diff
@@ -19,5 +19,5 @@
   Check out what other #{ @course.badge_term.pluralize.downcase } you can
   earn in #{ @course.name }
   = succeed "." do
-    %a{:href => "https://www.gradecraft.com/my_badges"} here
-%hr/
\ No newline at end of file
+    %a{:href => "https://www.gradecraft.com/badges"} here
```
jwright: We should use a path helper here

```
= link_to "here", badges_url
```

app/views/notification_mailer/earned_badge_awarded.text.haml
```diff
@@ -7,4 +7,4 @@ You've earned the #{ @earned_badge.badge.name } #{ @course.badge_term }.
 
 Check out what other #{ @course.badge_term.pluralize } you can earn in #{ @course.name.html_safe } here:
 
-https://www.gradecraft.com/my_badges
+https://www.gradecraft.com/badges
```
jwright: We should use a path helper here

```
= link_to "here", badges_url
```

config/routes.rb
```diff
@@ -163,6 +163,8 @@
     end
   end
 
+  get "(/students/:student_id)/badges", to: "badges#index"
```
jwright: Since we need a `student` resource here, I believe we should use nested resourceful routing here.

```
resources :students, only: [], module: :students do
  resources :badges, only: :index
end
```

This will move the `badges#index` action to `students/badges#index`. This is what I am doing for grades and it works out well because it forces you to have a resource.

spec/features/editing_a_badge_spec.rb
```diff
@@ -20,7 +20,7 @@
       expect(current_path).to eq badges_path
 
       within(".pageContent") do
-        click_link "Fancy Badge"
+        first(:link, "Fancy Badge").click
```
jwright: I am unclear why we need this change.

jwright:
I like that we removed this logic from the `StudentsController` since the resource is a badge.

I do believe we need a few changes here to keep it using the correct resources. I feel like we should move the `badges#index` to `students/badges#index` since we need a `student_id` resource.

In addition, I am not a fan of checking a `@student` instance variable to do certain things. I don't quite fully understand when you would have one and when you would not. I feel like we should split this up into 2 controllers and views. One for students and one for staff. I realize this is a bigger change but I feel that the branching makes this a bit less maintainable. Thoughts?

app/controllers/badges_controller.rb
acousticrobot: 👍

app/controllers/badges_controller.rb
acousticrobot: Agree. I just moved this because of, you know, Rubocop 80 chars blah blah but I don't want to be changing the predictor in this pull request.

app/controllers/badges_controller.rb
acousticrobot: 👍

app/models/user.rb
acousticrobot: Yes we can. I'm not sure why Mike used [:id] in all the other badge methods on user, but I'll change this one.

app/views/notification_mailer/earned_badge_awarded.html.haml
acousticrobot: Good idea, but I don't know how to trigger this view -- the only change was the missing EOL. @chcholman do you want to check this?

app/views/notification_mailer/earned_badge_awarded.text.haml
acousticrobot: How do we test mailers? I haven't done it in GC

spec/features/editing_a_badge_spec.rb
acousticrobot: I was getting an error from this being an Ambiguous selector (more than one on the page).

app/controllers/badges_controller.rb
acousticrobot: done (in concurrent PR for predictor refactor)

app/views/notification_mailer/earned_badge_awarded.text.haml
jwright: You can test to ensure that the correct url is in the body via:

```
describe NotificationMailer do
  describe "#earned_badge_awarded" do
    it "contains the badges url" do
      expect(email.body).to include badges_url
    end
  end
end
```

`email` comes from `let(:email) { ActionMailer::Base.deliveries.last }`

It looks like @zazaian extracted the email stuff into a "toolkit" (`Toolkits::Mailers::EmailToolkit`). You can take a look at `spec/mailers/user_mailer` for some good examples.

app/models/user.rb
```diff
@@ -388,6 +388,22 @@ def unique_student_earned_badges(course)
...
+  def visible_earned_badges_for_badge(badge)
+    EarnedBadge
+      .where(badge: badge)
+      .where(student_id: self.id)
+      .where(student_visible: true)
+  end
+
+  # Number of times a student has earned a particular badge
+  def visible_earned_badges_for_badge_count(badge)
+    EarnedBadge
```
jwright: We should probably reduce duplication here with changing this by 

```
def visible_earned_badges_for_badge_count(badge)
  self. visible_earned_badges_for_badge(badge).count
end
```

jwright:
This looks great. I added one more issue and answered the email testing queston. Other than that, it's ready to go!

app/views/notification_mailer/earned_badge_awarded.html.haml
acousticrobot: done

app/views/notification_mailer/earned_badge_awarded.text.haml
acousticrobot: thanks, done

PR #1948, opened 2016-04-27
1747 grade routes redux 2

### LINKED ISSUE #1747: Change GradesController to use Grade as the resource

jwright:
Moves some of the remaining grade routes to make them more restful or moves them to a controller that represents the resource they are using.

* `grades#edit_status` -> Moved to `assignments/grades#edit_status` since it needs to get the assignment and deals with multiple grades
* `grades#update_status` -> Moved to `assignments/grades#update_status` since it needs to get the assignment and deals with multiple grades
* `grades#import` -> Moved to `assignments/grades#import` since it needs to get the assignment and deals with multiple grades
* `grades#upload` -> Moved to `assignments/grades#upload` since it needs to get the assignment and deals with multiple grades
* `grades#self_log` -> Moved to `assignments/grades#self_log` since it needs to get the assignment. Even though it only deals with one grade, there is a chance that grade does not exist so it needs to create one which is why the `assignment_id` is important. If it did not have to create the grade, then it could be on the `GradesController` with a `/grades/:id/self_log` url.
* `grades#remove` -> Moved the url from `assignments/:id/grades/remove` to `grades/:id/remove` since it uses the id as a grade id but it was labeled as an assignment resource.
* `grades#exclude` -> Moved the url from `assignments/:id/grades/exclude` to `grades/:id/exclude` since it uses the id as a grade id but it was labeled as an assignment resource.
* `grades#include` -> Moved the url from `assignments/:id/grades/include` to `grades/:id/include` since it uses the id as a grade id but it was labeled as an assignment resource.
* `grades#predict_score` -> Moved the url from `assignments/:id/grades/predict_score` to `grades/:id/predict_score` since it did not need to use the assignment as the resource and instead now uses the grade as the resource.

In addition, the `GradesAbility` module and `GradeProctor` needed the ability to pass in a boolean to check whether a grade should be updateable by checking student logged. This was needed with `grades#feedback_read` as I wanted to use `CanCan` but it was not updating the `Grade#score`, only `Grade#feedback_read`.

There will be one more PR for #1747 to consolidate the `resource grade` routes into the existing `GradesController` and `resource grades` routes.

Close #1747


lib/grade_proctor/updatable.rb
```diff
@@ -15,10 +15,13 @@ def updatable?(options={})
 
       user = options[:user] || grade.student
       course = options[:course] || grade.course
+      student_logged = options[:student_logged].nil? ? true :
+        options[:student_logged]
```
acousticrobot: Rubocop has several problems with this:

```
lib/grade_proctor/updatable.rb:13:5: C: Cyclomatic complexity for updatable? is too high. [9/6]
    def updatable?(options={})
    ^^^
lib/grade_proctor/updatable.rb:13:5: C: Perceived complexity for updatable? is too high. [9/7]
    def updatable?(options={})
    ^^^
lib/grade_proctor/updatable.rb:18:24: C: Avoid multi-line ?: (the ternary operator); use if/unless instead.
      student_logged = options[:student_logged].nil? ? true :
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

PR #1955, opened 2016-05-04
Add "purpose" to assignments

### LINKED ISSUE #1946: Add "purpose" to assignments
Transparent teaching practices suggest that instructors should share three pieces of info with students regarding assignments: a clear description of the task itself, an understanding of the purpose behind their doing it, and clear criteria for assessment. We have a description and criteria, but we need to nudge instructors to describe their purpose. 

* Add a purpose textarea (no character limit) to the Assignment table
* Add a form to allows instructors to enter this data on the assignment creation form
* Show this info back to the student when they're looking at the assignment page

ksofa2:
Fixes #1946

### DIFF
--- a/app/models/concerns/unlockable_condition.rb
+++ b/app/models/concerns/unlockable_condition.rb
@@ -92,6 +92,13 @@ module UnlockableCondition
     (!is_unlockable? && visible?)
   end
 
+  def purpose_visible_for_student?(student)
+    (is_unlockable? &&
+      (visible_when_locked? && show_purpose_when_locked? ||
+     is_unlocked_for_student?(student))) ||
+    (!is_unlockable? && visible?)
+  end
+
   def points_visible_for_student?(student)
     (is_unlockable? &&
       (visible_when_locked? && show_points_when_locked? ||
--- a/app/views/assignments/_description_and_downloads.haml
+++ b/app/views/assignments/_description_and_downloads.haml
@@ -11,7 +11,7 @@
   .italic= "Opens: #{presenter.assignment.open_at}" if presenter.assignment.open_at?
   .italic= "Due: #{presenter.assignment.due_at}" if presenter.assignment.due_at?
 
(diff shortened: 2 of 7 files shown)


chcholman:
@ksofa2 is this ready for review?

ksofa2:
@chcholman Yes, ready.

app/models/concerns/unlockable_condition.rb
```diff
@@ -92,6 +92,13 @@ def description_visible_for_student?(student)
     (!is_unlockable? && visible?)
   end
 
+  def purpose_visible_for_student?(student)
```
acousticrobot: I think this is good for now, but seeing how much duplicate logic is shared among the visibility queries, we might want to refactor this file.

app/views/assignments/_description_and_downloads.haml
```diff
@@ -76,6 +76,11 @@
     %h4.uppercase Guidelines:
     %p= raw presenter.assignment.description
 
+  - if presenter.assignment.purpose.present? && ( !current_student || presenter.assignment.purpose_visible_for_student?(current_student))
```
acousticrobot: I think that we could reduce the view complexity here by putting the logic onto the presenter.

I also think we should use `current_user.is_staff?` instead of `!current_student` if this logic
is being used to only show invisible purposes to the professor

something like:

```
if presenter.assignment_has_viewible_purpose?(current_user)
...

# /app/presenters/assignments/presenter.rb
def assignment_has_viewible_purpose?(current_user)
  assignment.purpose.present? && 
  ( current_user.is_staff? || 
    assignment.purpose_visible_for_student?(current_user)
  )
end
```

we should probably do the same refactoring for these existing lines in this file:
```
14: - if current_student.present? && presenter.assignment.accepts_submissions? &&  (presenter.assignment.is_unlocked_for_student?(current_student) || (presenter.assignment.has_groups? && presenter.assignment.is_unlocked_for_group?(current_student.group_for_assignment(presenter.assignment))))`

74: - if presenter.assignment.description.present? && ( !current_student || presenter.assignment.description_visible_for_student?(current_student))
```

ksofa2:
Thanks for the feedback Jonathan. I've added a commit to refactor some of the logic out of the view.

PR #1956, opened 2016-05-04
Refactor/predictor routing

acousticrobot:
This is the first step on the path to moving the predictor JSON endpoints into the api namespace.

#### JSON changes

The JSON has been updated to conform to the json::api standard we are using in the existing `api/` routes. ("data" and nested "attributes"). The Angular service has been updated to process the updated format.

#### route changes: 

  * Routing has been made resourceful
```
/predictor_badges => /api/predicted_earned_badges
/students/:id/predictor_badges => /api/students/:student_id/predicted_earned_badges
/badges/:badge_id/predict_times_earned => POST /api/predicted_earned_badges/:id
```
   * The Angular passes updates directly to the PredictedEarnedBadge instead of the Badge.

#### controller changes

Instead of overloading the controller routes, the faculty preview of a students predictor now exists in it's own controller in the `api/students/` namespace. In order to eliminate duplication, the shared code has been moved into a concern, and both controllers use the same view.


app/controllers/concerns/predictor_data.rb
```diff
@@ -0,0 +1,35 @@
...
+      :point_total,
+      :visible,
+      :visible_when_locked,
+      :can_earn_multiple_times,
+      :position,
+      :updated_at,
+      :icon
+    ).map do |badge|
+      prediction = badge.find_or_create_predicted_earned_badge(student.id)
+      if current_user.is_student?(current_course)
```
jwright: Do we want to use the `student` here that was passed in instead of `current_user`?

jwright:
Looks great. I just had one question. Also, the codeclimate issues seem to be fine but I think we should turn off the "line too long" cop off. Thoughts @chcholman?

chcholman:
@jwright I agree - I want this 80 character limit to be our goal, but not to force us to write less intelligible code to reach it.

app/controllers/concerns/predictor_data.rb
acousticrobot: Nope, `student.is_student?` will always be true, but `currrent_user` could be faculty. This is where we remove prediction information for the faculty view -- logic that should eventually move to a proctor.

chcholman:
@acousticrobot I did the rebase against master to get this to merge - will you check to make sure the tests I took out were appropriate?

PR #1959, opened 2016-05-05
1747 grade routes redux 3

### LINKED ISSUE #1747: Change GradesController to use Grade as the resource

jwright:
This PR moves the `Grade` show route from `/assignments/:assignment_id/grade/:id` to `/grades/:id`.

In order to do this, the `AssignmentPresenter` had to be removed from usage within the `GradesController`. This refactor cleaned up a lot of presenter pollution since the `Grade` resource should not of used the `AssignmentPresenter`.

Some of the more complicated methods that were being used on the `Presenter` were moved to the `Level` model.

Since the param `?student_id=x` was being removed for the faculty view, the `current_student` could no longer be used in certain views (i.e. `rubric_table` partial) and so the `student` was passed down to the partials through the controller.

In addition, anytime a partial was being rendered, the "explicit long form" was used. (This will affect a lot of views @ksofa2).

Closes #1747


app/views/rubrics/components/_criteria.haml
```diff
@@ -7,10 +7,10 @@
...
-      = render "rubrics/components/level", presenter: presenter, level: level, include_grade_info: include_grade_info
+      = render partial: "rubrics/components/level", locals: { level: level, student: student, include_grade_info: include_grade_info }
 
-    - if criterion.levels.size < presenter.rubric_max_level_count
-      %td.filler(colspan="#{presenter.rubric_max_level_count - criterion.levels.size}")
+    - if criterion.levels.size < rubric.max_level_count
+      %td.filler(colspan="#{rubric.max_level_count - criterion.levels.size}")
       - if include_grade_info
         %td.level.level-comments.comments-box
         - unless comments_by_criterion_id.empty?
```
jwright: @acousticrobot I have no idea where `comments_by_criterion_id` is coming from and I wanted to see if you knew. This is something that I did not touch. Any idea?

I came across it because I was looking for usage of https://github.com/UM-USElab/gradecraft-development/pull/1959/files#diff-835ff61547c63e13a58ceed7adf29fe0L18 but line 16 above is not using a presenter to call.

ksofa2:
Thanks for the heads up re partials. I've not started that work yet. I'll keep this PR in mind when doing so.

app/controllers/grades_controller.rb
```diff
@@ -1,35 +1,21 @@
...
-      @title = "#{current_student.name}'s Grade for #{ @assignment.name }"
-    end
+    redirect_to @grade.assignment and return if current_user_is_student?
 
-    render :show, Assignments::Presenter.build({ assignment: @assignment,
-                                              course: current_course,
-                                              view_context: view_context })
+    name = @grade.group.nil? ? @grade.student.name : @grade.group.name
+    @title = "#{name}'s Grade for #{@grade.assignment.name}"
   end
```
acousticrobot: 👍 slick

app/models/level.rb
```diff
@@ -27,6 +27,15 @@ def copy(attributes={})
     ModelCopier.new(self).copy(attributes: attributes, associations: [:badges])
   end
 
+  def earned_for?(student_id)
+    self.criterion_grades.where(student_id: student_id).exists?
+  end
+
+  def hide_analytics?
```
acousticrobot: I know this was logic moved from the presenter, but I think it warrants a comment. I'm not sure what the use case is out of context, and I'm very confused why both `assignment.hide_analytics?` and `course.hide_analytics?` would have to be true. Maybe @chcholman you can add an explanation before merge (unless you know what's going on here @jwright )?

app/views/assignments/_show_instructor.haml
```diff
@@ -34,7 +34,7 @@
         .content#tabt3{role: "tabpanel", "aria-hidden" => false }
           %br
             .tab-container
-              = render "rubrics/components/rubric_table", presenter: presenter, include_grade_info: false
+              = render "rubrics/components/rubric_table", rubric: presenter.rubric, student: nil, include_grade_info: false
```
acousticrobot: We should to change this to use `partial:` and `locals:`, but I'm guessing you left this for @ksofa2

app/models/level.rb
jwright: I pulled the `hide_analytics?` code from the presenter. I looked through the code and we really only check for `Course#hide_anaytics?` before we do anything. I feel like we may only need to look for `Course#hide_analytics?` but it makes we wonder why we have `Assignment#hide_analytics?`.

chcholman: @acousticrobot @jwright Both are important - instructors can choose to hide analytics for all assignments (`Course#hide_analytics`) or just for a specific assignment (`Assignment#hide_analytics`).

jwright:
@acousticrobot Your changes look great! I fixed some issues with mine as well. It's good to pass on if you think so.

PR #1963, opened 2016-05-09
1747 grade routes destroy refactor

### LINKED ISSUE #1747: Change GradesController to use Grade as the resource

jwright:
Changes the `Grades#destroy` controller action to use a simple `Grade` object as the parameter instead of an `Assignment` and `student_id`. 

The path was changed from `DELETE assignments/:id/grade?student_id=:student_id` to `DELETE grades/:id`.

The `GradeProctor` got a new method (`#destroyable?`) which is basically just an aliased `#updatable?` method.

Closes #1747

### DIFF
--- a/app/controllers/grades_controller.rb
+++ b/app/controllers/grades_controller.rb
@@ -1,6 +1,6 @@
 class GradesController < ApplicationController
   respond_to :html, :json
-  before_filter :set_assignment, only: [:edit, :update, :destroy]
+  before_filter :set_assignment, only: [:edit, :update]
   before_filter :ensure_staff?,
     except: [:feedback_read, :show, :predict_score, :async_update]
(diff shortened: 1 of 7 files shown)


app/controllers/grades_controller.rb
```diff
@@ -171,15 +171,19 @@ def include
...
+    authorize! :destroy, grade
+    grade.destroy
+    score_recalculator(grade.student)
 
-    redirect_to assignment_path(@assignment), notice: "#{ @grade.student.name }'s
-      #{ @assignment.name } grade was successfully deleted."
+    redirect_to assignment_path(grade.assignment),
+      notice: "#{grade.student.name}'s #{grade.assignment.name} grade was "\
+              "successfully deleted."
+  rescue CanCan::AccessDenied
```
acousticrobot: 👍

PR #1966, opened 2016-05-09
Updated announcements to go to all users

chcholman:
This PR updates the announcements feature so that they're sent to all course members (not just students).

### DIFF
--- a/app/models/announcement.rb
+++ b/app/models/announcement.rb
@@ -12,15 +12,15 @@ class Announcement < ActiveRecord::Base
 
   default_scope { order "created_at DESC" }
 
-  def self.read_count_for(student, course)
+  def self.read_count_for(user, course)
     AnnouncementState
       .joins(:announcement)
       .where(announcements: { course_id: course.id })
-      .where(user_id: student.id).count
+      .where(user_id: user.id).count
   end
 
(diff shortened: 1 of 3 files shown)


app/models/announcement.rb
```diff
@@ -29,15 +29,15 @@ def abstract(words=25)
...
+      course.users.each do |user|
+        AnnouncementMailer.announcement_email(self, user).deliver_now
       end
     end
   end
 
   def mark_as_read!(user)
-    if user.is_student?(course) &&
-       !states.map(&:user_id).include?(user.id)
+    if (user.is_student?(course) || user.is_staff?(course)) &&
```
jwright: I feel like we should just eliminate the user check here since it doesn't matter. It will reduce the complexity and an unnecessary database check.

jwright:
@chcholman I just had a brief comment about an unnecessary (??) check for the user. Other than that, it looks great!

app/models/announcement.rb
chcholman: I tried that initially - but then it allowed non-course members to mark as read. Maybe that's too edge case to be concerned about? There's no way we'd anticipate that they'd be able to actually see or receive them.

jwright:
Very nice!

PR #1982, opened 2016-05-11
refactor predicted earned challenge routes

acousticrobot:
This PR refactors the predicted_earned_challenge routing, following the pattern used when I refactored the predicted_earned_badges.

New routes are added to the api namespace that directly manage the `predicted_earned_challenges`, and a `students` namespace is used for the faculty preview.

The JSON follows the json api spec, using a "data" array, with entries with "attributes".


app/controllers/concerns/predictor_data.rb
```diff
@@ -32,4 +32,58 @@ def predictor_badges(student)
       end
     end
   end
+
+  def predictor_challenges(student)
```
acousticrobot: Please note, `predictor_challenges` logic was moved here without refactoring. I'm going to clean this method up in my next PR wihch is adding a `predicted_earned_grade`, since they will need to behave the same.

app/controllers/api/predicted_earned_challenges_controller.rb
```diff
@@ -0,0 +1,63 @@
...
+  end
+
+  # POST api/predicted_earned_challenges/:id
+  def update
+    prediction = PredictedEarnedChallenge.where(
+      student: current_student,
+      id: params[:id]
+    ).first
+
+    prediction.points_earned = params[:points_earned]
```
jwright: We should probably return a 404 if the prediction is not found here.

app/controllers/api/predicted_earned_challenges_controller.rb
```diff
@@ -0,0 +1,63 @@
...
+    prediction.save
+
+    # This should be extracted with the rest of the event_loggers
+    PredictorEventJob.new(
+      data: predictor_event_attrs(prediction.challenge, prediction.valid?)
+    ).enqueue
+
+    if prediction.valid?
```
jwright: 👍

jwright:
This looks great (except the build is broken - I restarted it).

I did have one comment where we may want to check for a nil resource.

app/controllers/api/predicted_earned_challenges_controller.rb
acousticrobot: This is a good idea, I'm going to add a 404 check to this controller, and the predicted grades controller in my next PR, once the related migrations have been added.

PR #1984, opened 2016-05-13
Added requirements to assignment types

chcholman:
Added null: false requirements to assignment type fields per our schema cleanup conversation


app/models/assignment_type.rb
```diff
@@ -95,7 +95,7 @@ def raw_score_for_student(student)
   private
 
   def positive_max_points
-    if max_points? && max_points < 1
+    if max_points < 0
```
jwright: Should this now include zero where it did not before?

app/models/assignment_type.rb
```diff
@@ -74,7 +74,7 @@ def weighted_total_for_student(student)
 
   def visible_score_for_student(student)
     score = score_for_student(student)
-    if max_points? && score > max_points
+    if (max_points > 0) && score > max_points
```
jwright: I believe this check for `max_points > 0` can be removed since it will no longer be nil

jwright:
We may need to add a rake task (or include it with the existing migration) to update the existing data in the database.

This is from the Postgres documentation:

```
SET/DROP DEFAULT
These forms set or remove the default value for a column. The default values only apply to subsequent INSERT commands; they do not cause rows already in the table to change. Defaults can also be created for views, in which case they are inserted into INSERT statements on the view before the view's ON INSERT rule is applied.
```

So, the migration may work for future inserts and updates but the code may no longer work if nulls are returned.

jwright:
Everything else looked great!

PR #1987, opened 2016-05-14
Cleaned up froala implementation

chcholman:
This PR trims down the number of froala utilities in use overall, particularly on the Criterion Grade implementation. It makes the utility consistent between the Rubric Grade overall feedback, Grade feedback, and other text editing field implementations. It also fixes 1/2 JS issues that came about as a result of the upgrade.

### DIFF
--- a/app/assets/javascripts/angular/controllers/GradeCtrl.js.coffee
+++ b/app/assets/javascripts/angular/controllers/GradeCtrl.js.coffee
@@ -156,12 +156,12 @@
   $scope.froalaOptions = {
     inlineMode: false,
     minHeight: 100,
-    buttons: [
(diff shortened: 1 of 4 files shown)


app/assets/javascripts/angular/controllers/GradeCtrl.js.coffee
```diff
@@ -156,12 +156,12 @@
   $scope.froalaOptions = {
     inlineMode: false,
     minHeight: 100,
```
acousticrobot: I think you want to change this to use `heightMin`, and probably to `heightMin: 200,` which will match the text field on the rubric grading page.

PR #1988, opened 2016-05-17
Pr/predicted earned grade

acousticrobot:
This PR adds a `PredictedEarnedGrade` model, which pulls all Prediction Logic off of the `Grade` model. Routing follows the pattern of the other Prediction models, through the API namespace.

This PR also:

  * Standardizes prediction models to use `predicted_points` (`predicted_times_earned` for badges)
  * Updates the sample data to create `PredictedEarnedGrades`
  * Includes a rake task `rake grades:transfer_predictions` to move existing predictions from Grades to PredictedEarnedGrades. Once this is run, we can drop the field off of `Grades`.

Note: In `app/models/grade.rb`, `Grade#predicted_points` is now a method that finds the PredictedEarnedGrade with the same student and assignment, to handle view calls that present the predicted grade for a grade. I would like to examine the pages that use both `Grade` and `PredictedEarnedGrade` info (the syllabus page, for one), and come up with a better way retrieve this information in one call.


app/assets/javascripts/angular/services/PredictorService.js.coffee
```diff
@@ -62,13 +62,15 @@
...
         )
 
-    getAssignments = ()->
-      $http.get("predictor_assignments").success( (data)->
-        angular.copy(data.assignments,assignments)
-        termFor.assignment = data.term_for_assignment
-        termFor.pass = data.term_for_pass
-        termFor.fail = data.term_for_fail
-        update.assignments = data.update_assignments
+    getAssignments = (id)->
```
jwright: Should we change this to `student_id` instead of just `id`? I believe that is what it is. It currently seems like it's the assignment id.

app/controllers/concerns/predictor_data.rb
```diff
@@ -20,70 +18,62 @@ def predictor_badges(student)
...
+    current_course.challenges.select(
+      :id,
+      :name,
+      :visible,
+      :description,
+      :point_total
+    ).map do |challenge|
+      prediction =
+        challenge.find_or_create_predicted_earned_challenge(@student.id)
+      if current_user.is_student?(current_course)
```
jwright: We can probably use `current_user_is_student?` here.

app/controllers/concerns/predictor_data.rb
```diff
@@ -20,70 +18,62 @@ def predictor_badges(student)
...
-          }
-        else
-          challenge.grade = {
-            point_total: challenge.point_total,
-            score: nil,
-            points_earned: nil
-          }
-        end
+      grade = grades.where(challenge_id: challenge.id).first
+      if grade.present? && grade.is_student_visible?
```
jwright: Don't we want to use the `GradeProctor` here instead of `is_student_visible?`?

app/controllers/grades_controller.rb
```diff
@@ -195,40 +195,6 @@ def feedback_read
       notice: "Thank you for letting us know!"
   end
 
-  # POST /grades/:id/predict_score
```
jwright: :+1:

jwright:
This looks really good! Great job.

I just had a few small points but it looks great to me.

app/assets/javascripts/angular/services/PredictorService.js.coffee
acousticrobot: Good idea 🎩 !

app/controllers/concerns/predictor_data.rb
acousticrobot: 👍

app/controllers/concerns/predictor_data.rb
acousticrobot: Good idea, but we need to handle ChallengeGrades first.

PR #1990, opened 2016-05-17
1747 grade earned badge routes

### LINKED ISSUE #1747: Change GradesController to use Grade as the resource

jwright:
Moves the routes `POST grade/earn_student_badge` to `POST api/earned_badges` and allows both the `POST api/earned_badges.json` and the `POST earned_badges` to use a service to create an `EarnedBadge`.

The previous version of `POST grade/earn_student_badge` did not send out the new badge notification nor did it recalculate the student's score. Now, with the new `CreatesEarnedBage` service, it does both.

Closes #1747

### DIFF
--- a/app/controllers/earned_badges_controller.rb
+++ b/app/controllers/earned_badges_controller.rb
@@ -1,11 +1,11 @@
-class EarnedBadgesController < ApplicationController
+require_relative "../services/creates_earned_badge"
 
+class EarnedBadgesController < ApplicationController
   # Earned badges are to badges what grades are to assignments - the record of
   # how what and how a student performed
 
   before_filter :ensure_staff?
-  before_action :find_badge,
-    only: [:index, :show, :new, :edit, :create, :update, :destroy ]
+  before_action :find_badge, except: [:mass_edit, :mass_earn]
   before_action :find_earned_badge, only: [:show, :edit, :update, :destroy ]
 
   def index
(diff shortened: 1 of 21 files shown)


app/controllers/earned_badges_controller.rb
```diff
@@ -1,11 +1,11 @@
...
+require_relative "../services/creates_earned_badge"
 
+class EarnedBadgesController < ApplicationController
   # Earned badges are to badges what grades are to assignments - the record of
   # how what and how a student performed
 
   before_filter :ensure_staff?
-  before_action :find_badge,
-    only: [:index, :show, :new, :edit, :create, :update, :destroy ]
+  before_action :find_badge, except: [:mass_edit, :mass_earn]
```
acousticrobot: 👍

spec/services/creates_earned_badge/creates_earned_badge_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+require "active_record_spec_helper"
+require "./app/services/creates_earned_badge/creates_earned_badge"
+
+describe Services::Actions::CreatesEarnedBadge do
+  let(:assignment) { create :assignment, course: course }
+  let(:badge) { create :badge, course: course }
+  let(:course) { create :course }
+  let(:course_membership) { create :student_course_membership, course: course }
+  let(:grade) { create :grade, course: course, student: student }
+  let(:student) { course_membership.user }
```
acousticrobot: You could reduce these 6 lines to:
```
let(:world) { World.create.with(:course, :assignment, :student, :badge, :grade) }
```
if you append word to the corresponding attributes below

spec/services/creates_earned_badge_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+require "./app/services/creates_earned_badge"
+
+describe Services::CreatesEarnedBadge do
+  describe ".award" do
+    let(:assignment) { create :assignment, course: course }
+    let(:badge) { create :badge, course: course }
+    let(:course) { create :course }
+    let(:course_membership) { create :student_course_membership, course: course }
+    let(:grade) { create :grade, course: course, student: student }
+    let(:student) { course_membership.user }
```
acousticrobot: Same change works: 
```
let(:world) { World.create.with(:course, :assignment, :student, :badge, :grade) }
```

jwright:
@acousticrobot This is amazing. Worked like a charm. Thank you.

PR #1992, opened 2016-05-18
Refactor/predictor grade scheme elements

acousticrobot:
This moves the json `grade_scheme_elements` into the api namespace. It also standardizes the json. The json result is the same for students and faculty, so it only required a single route.

The additional 10% added to the course total points (used to push the predictor graph beyond the highest grade) has been moved into the Angular logic, so that the api endpoint will accurately report the total points.


jwright:
@acousticrobot Looks great!

PR #1997, opened 2016-05-19
update path parsing for group rubrics grades

acousticrobot:
This PR updates the Angular path parsing for group rubric grading. 

The new route does not have a `group_id` query (yay!) but we forgot to update this in Angular. We should remember to do this for the student route as well when it is updated.

### DIFF
--- a/app/assets/javascripts/angular/controllers/GradeCtrl.js.coffee
+++ b/app/assets/javascripts/angular/controllers/GradeCtrl.js.coffee
@@ -14,7 +14,6 @@
 
     $scope.rawScoreUpdating = false
     $scope.hasChanges = false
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/angular/controllers/GradeCtrl.js.coffee
```diff
@@ -14,7 +14,6 @@
 
     $scope.rawScoreUpdating = false
     $scope.hasChanges = false
-    $scope.gradeStatuses = ["In Progress", "Graded", "Released"]
```
acousticrobot: Just found this -- it was no longer being used.

jwright:
2 for 2

PR #2000, opened 2016-05-20
refactor api routes using better resources

acousticrobot:
This commit includes a small reorganization of a number of api routes that were nested within assignments and students, to use better `resources` name-spacing.


jwright:
Looks good

PR #2001, opened 2016-05-20
Better Submission File Downloads, Proctor Conditions

zazaian:
Previously `SubmissionFile` records were only available for individual download using the name that was actually stored on S3, which can be too confusing for instructors if a student hasn't taken the time to rename any of the files they've uploaded to the Submission.

This PR seeks to consolidate logic from the `SubmissionFilesExporter` into some independent libraries for managing filename formatting and bulk submission file handling to avoid duplication, and also to begin extracting some of the responsibilities of the very-large `SubmissionExportsPerformer` class.

Additionally it also adds a better file streaming mechanism for individual submission file downloads in order to give us the ability to rename files in place as they're being downloaded by the user, rather than only allowing the user to download the file directly from S3, filename and all.

This PR also makes an effort to add some structure and some more universally applicable components for a new `Proctor` module that includes logic for handling large sets of Proctor conditions with a new `Proctor::Conditions` module that allows us to avoid defining very large sets of if-else statements in order to rein in all of the logic for any give Proctor action.

There's also a new `S3Manager::ObjectStream` class intended to be used as a wrapper for unencrypted files being streamed directly from S3, as well as a new `Formatter` library that should handle all of our filename formatting (`Formatter::Filename`) or other text formatting concerns from here on out.


app/controllers/submission_files_controller.rb
```diff
@@ -0,0 +1,18 @@
...
+      send_data presenter.stream_submission_file, filename: presenter.filename
+    else
+      presenter.mark_submission_file_missing
+      flash[:alert] = "The requested file was not found on the server."
+      redirect_to request.referrer
+    end
+  end
+
+  def presenter
+    @presenter ||= SubmissionFilesPresenter.new params: params
```
jwright: The other presenters are named after their controller/action. I feel like we should name this `Submission::DownloadPresenter`.

app/models/abilities/submission_file_ability.rb
```diff
@@ -0,0 +1,7 @@
+module SubmissionFileAbility
+  def define_submission_file_abilities(user, course)
```
jwright: Can we remove the `course` here since we are not using it?

app/models/group.rb
```diff
@@ -98,7 +98,6 @@ def submissions_by_assignment_id
   end
 
   def cache_associations
-    # TODO: no method assignment for group!
-    self.course_id ||= assignment.try(:course_id)
+    self.course_id ||= assignments.first.try(:course_id)
```
jwright: Why was this change made?

app/presenters/submission_files_presenter.rb
```diff
@@ -0,0 +1,32 @@
...
+    submission_file.submission
+  end
+
+  def submission_file_streamable?
+    return false unless submission_file
+    submission_file.object_stream.exists?
+  end
+
+  def stream_submission_file
+    return false unless submission_file_streamable?
```
jwright: Why return `false` here? Maybe `nil` makes more sense?

app/presenters/submission_files_presenter.rb
```diff
@@ -0,0 +1,32 @@
...
+    submission_file.object_stream.stream!
+  end
+
+  def filename
+    submission_file.instructor_filename params[:index].to_i
+  end
+
+  def mark_submission_file_missing
+    return false unless submission_file
```
jwright: I don't think we want to return anything here. Perhaps: `submission_file.mark_file_missing if submission_file.present?`?

app/presenters/submissions/edit_presenter.rb
```diff
@@ -1,18 +1,20 @@
...
-    if view_context.current_user.is_student?(course)
-      "Editing My Submission for #{assignment.name}"
-    else
-      if assignment.has_groups?
-        "Editing #{group.name}'s Submission"
+    def title
+      if view_context.current_user.is_student?(course)
+        "Editing My Submission for #{assignment.name}"
       else
-        "Editing #{student.name}'s Submission"
```
jwright: A simple change here reduces the small amount of duplication:

```
name = assignment.has_groups? ? group.name : student.name
"Editing #{name}'s Submission"
```

app/presenters/submissions/show_presenter.rb
```diff
@@ -1,54 +1,69 @@
...
+      owner_id_attr = individual_assignment? ? :student_id : :group_id
+      @grade ||= assignment.grades.find_by "#{owner_id_attr}": owner.id
+    end
 
-  def student
-    submission.student
-  end
+    def submission
+      begin
+        @submission ||= ::Submission.find(id)
```
jwright: I would recommend changing this to `@submission ||= Submission.where(id: id).first` and remove the wrapping in a `rescue` block.

app/views/submission_files/download.html.haml
```diff
@@ -0,0 +1 @@
+-# Leave this here, Rails gets angry on send_data calls without a view present.
```
jwright: I think we just need to change the controller to `send_data .... and return` instead of defining this view.

lib/proctor/condition.rb
```diff
@@ -0,0 +1,22 @@
+module Proctor
+  class Condition
+    attr_reader :condition, :name
+
+    def initialize(name:, &condition)
+      @name = name.to_s
+      @condition = condition
+    end
+
+    def outcome
```
jwright: I like the method name `result` instead of `outcome` here. Thoughts?

lib/proctor/conditions.rb
```diff
@@ -0,0 +1,59 @@
+module Proctor
```
jwright: I think this file would benefit a lot with some class and method commenting. I think some example signatures about the methods in the documentation would go a long way.

lib/proctor/override.rb
```diff
@@ -0,0 +1,4 @@
+module Proctor
```
jwright: I think an example of how this should be inherited in some class level documentation would help the reader.

lib/proctor/requirement.rb
```diff
@@ -0,0 +1,4 @@
+module Proctor
```
jwright: I think an example of how this should be inherited in some class level documentation would help the reader.

lib/proctor.rb
```diff
@@ -0,0 +1,9 @@
+require_relative "proctor/condition"
+require_relative "proctor/conditions"
+require_relative "proctor/conditions/class_methods"
```
jwright: I think this should be required from the `proctor/conditions` file since that is it's parent.

lib/proctor.rb
```diff
@@ -0,0 +1,9 @@
...
+require_relative "proctor/requirement"
+require_relative "proctor/override"
+require_relative "proctor/proctors"
```
jwright: This file is not in this PR. It's also the reason that CI is failing.

app/proctors/conditions/submission_file_conditions.rb
```diff
@@ -0,0 +1,46 @@
...
+  class SubmissionFileConditions
+    include Proctor::Conditions
+
+    attr_accessor :group
+
+    # this uses the resources on the proctor rather than defining them here
+    defer_to_proctor :submission, :assignment, :course, :submission_file
+
+    def downloadable_conditions
+      add_requirements :submission_matches_course?, :assignment_present?
```
jwright: What is the difference between `requirements` and `overrides`?

app/views/submissions/components/_files.haml
```diff
@@ -1,6 +1,17 @@
...
+          download_submission_file_path(submission_file, index: index)
+
+
+  %span.smallcaps Missing Attachments:
+  %ul
+    -# otherwise don't render a link
+    - presenter.missing_submission_files.each_with_index do |submission_file, index|
+      %li
+        %i.fa.fa-chain-broken.fa-fw
+        = "#{submission_file.instructor_filename index + presenter.present_submission_files.count}"
```
jwright: No need for string interpolation here I think.

config/routes.rb
```diff
@@ -107,6 +107,8 @@
     end
   end
 
+  get "submission_files/:id/download", to: "submission_files#download", as: "download_submission_file"
```
jwright: Can we use resourceful routing here?

```
resources :submission_files, only: [] do
  get :download
end
```

db/migrate/20160510054444_add_excluded_by_id_to_grades.rb
```diff
@@ -0,0 +1,5 @@
+class AddExcludedByIdToGrades < ActiveRecord::Migration
+  def change
+    add_column :grades, :excluded_by_id, :integer
```
jwright: What is this used for?

db/migrate/20160510054639_add_excluded_at_to_grades.rb
```diff
@@ -0,0 +1,5 @@
+class AddExcludedAtToGrades < ActiveRecord::Migration
+  def change
+    add_column :grades, :excluded_at, :timestamp
```
jwright: What is this used for?

db/migrate/20160511151506_remove_step_booleans_from_submissions_export.rb
```diff
@@ -0,0 +1,35 @@
+class RemoveStepBooleansFromSubmissionsExport < ActiveRecord::Migration
```
jwright: 👍

lib/grade_proctor/updatable.rb
```diff
@@ -10,6 +10,7 @@ class GradeProctor
   module Updatable
```
jwright: We need to move this to the new proctor architecture, correct?

lib/grade_proctor/updatable.rb
```diff
@@ -10,6 +10,7 @@ class GradeProctor
...
     include Base
 
+    # maybe we should have required arguments here
```
jwright: I think we should have an `updatable_by?` with a required user and an `updatable?` with just options because a lot of times it checks the `Grade#student`

spec/controllers/grades_controller_spec.rb
```diff
@@ -1,4 +1,5 @@
 require "rails_spec_helper"
+require "grade_proctor"
```
jwright: How is this working if `GradeProctor` is at `./lib/grade_proctor`?

spec/controllers/submission_files_controller_spec.rb
```diff
@@ -0,0 +1,122 @@
+require "rails_spec_helper"
+
+describe SubmissionFilesController do
+  let(:course) { Course.last }
```
jwright: You may want to change this to `Course.unscoped.last` in case a `default_scope` with an `order` is created.

spec/controllers/submission_files_controller_spec.rb
```diff
@@ -0,0 +1,122 @@
...
+  let(:student) { User.last }
```
jwright: You may want to change this to `User.unscoped.last` in case a `default_scope` with an `order` is created.

spec/lib/formatter/filename_spec.rb
```diff
@@ -0,0 +1,89 @@
...
+  let(:foo_filename) { "some-rando-filename" }
+
+  describe "#initialize" do
+    it "sets the @filename and @original_filename" do
+      expect(subject.filename).to eq foo_filename
+      expect(subject.original_filename).to eq foo_filename
+    end
+  end
+
+  describe "accessible attributes" do
```
jwright: I am not sure this `describe` block is adding anything here. I think it can be removed.

spec/lib/proctor/condition_spec.rb
```diff
@@ -0,0 +1,77 @@
+require 'proctor'
```
jwright: Double quotes -- we should add this to rubocop.

spec/lib/proctor/condition_spec.rb
```diff
@@ -0,0 +1,77 @@
...
+      expect(this_condition.name).to eq "another_thing"
+    end
+
+    it "takes a condition as a block but doesn't call it" do
+      expect(subject.condition.class).to eq Proc
+      expect(subject.condition.call).to eq false
+    end
+  end
+
+  describe "readable attributes" do
```
jwright: I don't think this describe block adds anything here.

spec/presenters/submissions/edit_presenter_spec.rb
```diff
@@ -1,11 +1,87 @@
 require "./app/presenters/submissions/edit_presenter"
-require_relative "showing_a_submission_spec"
+require 'active_record_spec_helper'
+require './app/models/submission'
```
jwright: double quotes here.

spec/presenters/submissions/show_presenter_spec.rb
```diff
@@ -1,31 +1,319 @@
+require 'active_record_spec_helper'
```
jwright: double quotes.

spec/rails_spec_helper.rb
```diff
@@ -102,5 +102,6 @@
   # Remove uploader files, see config/environments/test.rb
   config.after(:all) do
     FileUtils.rm_rf(Dir["#{Rails.root}/spec/support/uploads"])
+    FileUtils.rm_rf(Dir["#{Rails.root}/public/uploads/*"])
```
jwright: Goodbye Jeff Moses!

jwright:
@zazaian Looks good. I like the new proctor stuff!

I had a lot of comments. There is an issue with the build because of a require for a file that is missing.

spec/rails_spec_helper.rb
acousticrobot: We hardly knew ye

app/controllers/submission_files_controller.rb
zazaian: This convention feels pretty weak to me, and like something that should be removed or re-engineered. Why are submissions presenters living under the first-level namespace of `Submissions`? Are we going to be using pluralized constants for namespaces for all other non-Rails-specific app resources moving forward? Why isn't this `SubmissionsPresenter::Show` or `Presenters::ShowSubmissions` or `Presenters::Submissions::Show`? I'm using `SubmissionFilesPresenter` here because it's the most consistent with Rails app resources and doesn't presume a one-to-one relationship between a presenter and a controller action.

app/models/abilities/submission_file_ability.rb
zazaian: That's fine.

app/models/group.rb
zazaian: I'm now sure how this was working before but there is no `assignment` method on `Group`. Calling this method would raise an error since `assignment` is not existent. This was crashing the spec suite for me.

app/presenters/submission_files_presenter.rb
zazaian: It's returning `false` because stream_submission_file suggests an action rather than the stream itself if it doesn't exist. If this were `submission_file_stream` I'd expect `nil`, but since it's the active phrase `stream_submission_file` I'd expect `false`. Seems like this should be consistent with the behavior of `object_stream.stream!`.

app/presenters/submission_files_presenter.rb
zazaian: Implicitly returning nil raise a `rubocop` error based on settings we've previously used. Seems like bad practice.

app/presenters/submissions/edit_presenter.rb
zazaian: That's a fine change.

app/presenters/submissions/show_presenter.rb
zazaian: Is there any reason for this change?

app/proctors/conditions/submission_file_conditions.rb
zazaian: All requirements must pass for the condition set to be considered valid, whereas the presence of one override will automatically mark the condition set as valid.

app/views/submissions/components/_files.haml
zazaian: Cool.

config/routes.rb
zazaian: That's fine.

db/migrate/20160510054444_add_excluded_by_id_to_grades.rb
zazaian: These were used to resolve schema conflicts. I'll pull them out.

lib/grade_proctor/updatable.rb
zazaian: Yes, but let's do this in another PR. I need to rewrite all of the existing Proctor classes.

lib/proctor.rb
zazaian: I pulled that out on the last commit because it was resolving a namespacing issue that I'd fixed. Fixed now.

lib/proctor/condition.rb
zazaian: Either is fine for me, though we're already using result in the RSpec library. Let's use Outcome for now and we can change it later when we rewrite the rest of the proctors if we'd like.

lib/proctor/conditions.rb
zazaian: Sounds good, I'll add some documentation here.

lib/proctor/override.rb
zazaian: :+1:

spec/controllers/grades_controller_spec.rb
zazaian: `require` will also check for files in `/lib`.

spec/controllers/submission_files_controller_spec.rb
zazaian: I think we've decided to stay away from default scopes altogether no?

app/controllers/submission_files_controller.rb
jwright: I think we should bring that up to the group and standardize on them. I don't really care as long as they are consistent. I will add it to 🎶 Tuesday Chats 🎶

app/presenters/submission_files_presenter.rb
jwright: Gotcha. I think that is boss.

app/presenters/submissions/show_presenter.rb
jwright: Yes, I think it's more readable to be explicit with what is going on. 

For example, someone else may not know that calling `find` and without that record being there will cause an `ActiveRecord::RecordNotFound` or someone may add code that causes a different record not to be found instead of the submission. 

In addition, that behavior may change in future versions of Rails. Control flow via exception handling can be confusing and they are less performant as well. If someone wraps this method call in a loop, that could be very damaging to performance.

zazaian: :+1: cool, sounds like a good improvement to me

lib/grade_proctor/updatable.rb
zazaian: :+1:

app/views/submission_files/download.html.haml
zazaian: This isn't a working solution.

zazaian:
@jwright I think all of the important stuff in the PR has been addressed and integrated. Take a look to see if this looks good to you.

app/controllers/submission_files_controller.rb
jwright: @zazaian I think we decided on using the `Presenters::Submissions::Download` here. Are we going to make these changes in a separate PR?

app/controllers/submissions_controller.rb
```diff
@@ -3,19 +3,14 @@ class SubmissionsController < ApplicationController
...
     authorize! :read, presenter.submission
     render :show, locals: { presenter: presenter }
   end
 
   def new
-    render :new, Submissions::NewPresenter.build(assignment_id: params[:assignment_id],
-                                              course: current_course,
-                                              group_id: params[:group_id],
-                                              view_context: view_context)
+    stuff = Submissions::NewPresenter.new(base_presenter_attrs)
```
jwright: `stuff`? 😄  I think we can remove the variable here and just say `render :new, Submissions::NewPresenter.build(base_presenter_attrs)`

app/controllers/submissions_controller.rb
```diff
@@ -79,4 +72,17 @@ def destroy
     assignment.submissions.find(params[:id]).destroy
     redirect_to assignment_path(assignment, notice: "Submission deleted")
   end
+
+  def presenter_attrs_with_id
```
jwright: Can we put this behind a `private` scope?

app/presenters/submission_files_presenter.rb
jwright: Sorry. I was referring to just removing the implicit return here altogether but I think `false` is fine here since `mark_file_missing` calls `update_attributes` which returns a boolean. 👍

app/presenters/submission_files_presenter.rb
```diff
@@ -0,0 +1,32 @@
+require "showtime"
+
+class SubmissionFilesPresenter < Showtime::Presenter
```
jwright: Are we going to move this file/make the standards change in another PR?

app/views/submission_files/download.html.haml
```diff
@@ -0,0 +1 @@
+-# leave this here for streaming data without having to circumvent render or redirect
```
jwright: I don't think we need this file if we just return from `send_data`. Have you tested that?

config/routes.rb
```diff
@@ -381,6 +385,15 @@
     end
   end
 
+  # 17b. Predictor, Student View
```
jwright: Why are these changes here? Is this a rebase mistake?

lib/proctor/condition_set.rb
```diff
@@ -0,0 +1,143 @@
...
+      send "#{condition_set}_conditions"
+      self
+    end
+
+    # Runs the checks for requirements and overrides to determine the final
+    # outcome of the result set. The condition set should be considered as being
+    # passed if all requirements are true or there are overrides that have
+    # succeeded, which preclude the need for checking requirements.
+    #
+    def satisfied_by?(user)
```
jwright: I think this method will not work for most proctors as most proctors require much greater context like a `course`, `level`, etc.

Can we change this to be a splat of args and remove the `user` altogether from the `ConditionSet`? As it stands now, this will not work for the `GradeProctor` and `BadgeProctor`. I am fine with doing this in a later PR when we convert the current proctors.

lib/proctor/condition_set.rb
```diff
@@ -0,0 +1,143 @@
...
+      @user = user
+      requirements_passed? || valid_overrides_present?
+    end
+
+    def reset_conditions
```
jwright: When is this needed? May be good to document this as well

lib/proctor/condition_set.rb
```diff
@@ -0,0 +1,143 @@
+require_relative "condition_set/defer"
```
jwright: I think `overrides` and `requirements` are very confusing. I am not sure if it's the naming or just the concept but I feel like they add a lot of complexity.

Maybe these are better offered up as logically "and'ing" or "or'ing" conditions so they work more like conventional boolean logic?

spec/controllers/submissions_controller_spec.rb
```diff
@@ -1,37 +1,72 @@
...
+            current_ability: ability,
+            presenter: presenter,
+            presenter_attrs_with_id: { some: "attrs", id: 5 }
+          )
+        allow(presenter_class).to receive(:new) { presenter }
+      end
+
       it "returns the submission show page" do
-        get :show, {id: @submission.id, assignment_id: @assignment.id}
+        result
```
jwright: We use `result` here and `make_request` in another spec. Can we standardize on one or the other?

spec/lib/proctor/override_spec.rb
```diff
@@ -0,0 +1,7 @@
+require 'proctor'
```
jwright: I am not sure this spec adds any value. I feel like it can be removed.

spec/lib/proctor/requirement_spec.rb
```diff
@@ -0,0 +1,7 @@
+require 'proctor'
```
jwright: I am not sure this spec adds any value. I feel like it can be removed.

spec/presenters/submissions/edit_presenter_spec.rb
```diff
@@ -1,11 +1,81 @@
...
-require_relative "showing_a_submission_spec"
+require_relative "../../../app/presenters/submissions/grade_history"
+require "./app/models/submission"
 
 describe Submissions::EditPresenter do
   let(:assignment) { double(:assignment) }
   before { allow(subject).to receive(:assignment).and_return assignment }
 
-  it_behaves_like "showing a submission"
+  describe "ported from shared examples" do
```
jwright: This describe block is used to show history and we can use git for that. Can we remove this describe block?

spec/support/test_classes/models/abilities/submission_file_ability_test.rb
```diff
@@ -0,0 +1,12 @@
+require "cancan"
+require_relative "../../../../../app/models/abilities/submission_file_ability"
```
jwright: Can we simplify this to `require_relative ./app/models/abilities/submission_file_ability`?

jwright:
@zazaian I added some more comments. Sorry man. This is a huge PR and it's taken a bit to wrap my head around all the things contained herein. I redact any comment about a new PR with a "yes" so we don't add anything else to this PR.

app/controllers/submission_files_controller.rb
zazaian: @jwright yes, let's convert all of the presenters to the agreed-upon naming format in a separate PR.

app/presenters/submission_files_presenter.rb
zazaian: Yes.

app/controllers/submissions_controller.rb
zazaian: Sure.

app/controllers/submissions_controller.rb
zazaian: So this didn't have any tests at all previously, and the build method is throwing errors when tested in this context. setting this to a separate variable allows stubbing out the new call without build throwing errors. Let's write out a separate issue for fixing the build method to resolve the testing issues.

app/views/submission_files/download.html.haml
zazaian: Yes, three times now. This is definitely needed.

config/routes.rb
zazaian: Yup, this came in from a rebase. Good catch :baseball:

lib/proctor/condition_set.rb
zazaian: Yup, let's address this in a later PR. :+1:

lib/proctor/condition_set.rb
zazaian: This is used to establish the arrays for the requirements and overrides against which the conditions are being checked. I'll add this to the documentation. :+1:

lib/proctor/condition_set.rb
zazaian: Let's address this in another PR. If you understand how they work this is a very logical workflow. What you're describing is basically just rewriting core ruby logic which really serves no practical purpose.

spec/controllers/submissions_controller_spec.rb
zazaian: Sure, I think this should be `make_request` in all instances in which the result is a RESTful controller call.

spec/lib/proctor/override_spec.rb
zazaian: If this was inheriting from a different class it would break the entire Proctor implementation. Let's keep this as-is.

spec/lib/proctor/requirement_spec.rb
zazaian: If this was inheriting from a different class it would break the entire Proctor implementation. Let's keep this as-is.

spec/support/test_classes/models/abilities/submission_file_ability_test.rb
zazaian: I think this would actually have to be `require ./app/models/abilities/submission_file_ability`. That's fine, but is actually more fragile that the existing solution as it's sensitive to the directory from which RSpec is being called. I don't think that would come up at any point, but it's worth noting.

app/views/submission_files/download.html.haml
jwright: @zazaian Ok. I think we want to use `send_file` instead here to remove the render call.

lib/proctor/condition_set.rb
jwright: Hmmm, I am not proposing that I think. I just think this API is confusing for developers. "Once you understand it" is the issue I would like to address. I agree, let's make a new PR for this.

zazaian: Are you suggesting that any learning curve at all in third party libraries is unacceptable? What you're proposing seems at least as confusing and not necessarily useful. We can do it in another PR but again I get the sense that you're just not comprehending the utility of this workflow.

app/controllers/submissions_controller.rb
jwright: Hmm. `.build` is being used in a lot of other controllers. If it is broken, we should probably fix this asap. Did you create an issue or do you want me to?

Also, I think you can stub out `build` just as you can with `new`.

zazaian: I stubbed out `build` and it was still throwing an error. Go ahead and
create an issue for it.

app/views/submission_files/download.html.haml
zazaian: So the problem with using `send_file` here is that we have to actually write the file on the server somewhere in order to pass it through, whereas send_data is literally streaming the data from S3 to the browser and the browser is applying the filename that we give it as an option. In theory send_file might be faster, but it adds the overhead of having to clean up tmp files that have been generated on the server just for the purposes of serving the content within the context of that one action. Also I'm not sure this solves the template issue, but I'll give it a shot.

lib/grade_proctor/updatable.rb
zazaian: Let's address this in the proctor PR :+1:

spec/lib/proctor/condition_spec.rb
zazaian: It provides better organization for these examples. Otherwise they're just nested inside of the top-level spec scope.

spec/lib/formatter/filename_spec.rb
zazaian: It provides better organization for these examples. Otherwise they're just nested inside of the top-level spec scope.

app/controllers/submission_files_controller.rb
```diff
@@ -0,0 +1,21 @@
+require_relative "../presenters/submission_files/base"
+
+class SubmissionFilesController < ApplicationController
+  def download
+    authorize! :download, presenter.submission_file
+
+    # let's use the object_stream here because there's no reason to hit S3 twice
+    if presenter.submission_file_streamable?
+      send_data presenter.stream_submission_file,
```
zazaian: @jwright I'm ultimately just going to re-implement the initial `send_data` streaming solution. It's a very lightweight implementation that does exactly what we need it to do, and since the data is being streamed we can suggest a filename to the browser without actually having to create a file on the server.

Because `send_file` requires a physical file to be sent, and because the overall thrust is this effort is to be able to provide dynamic filenames depending on who's actually downloading the file, using send_file is much more involved and fragile, as it involves first creating a tempdir on the our s3fs persistent disk, downloading the file from S3 to the the tmp dir, renaming it in place, THEN sending it to the client, and then after all of that deleting the tempdir with the renamed file. It's just a much more involved, more fragile implementation, and really isn't worth cleaning up the the empty download.html.haml template, which is simply a Rails limitation.

We could always implement this through javascript, but again, as all we'd be doing would be removing the template, I'm not sure that it's worth the effort.

zazaian:
@jwright this should be good to go.

spec/lib/formatter/filename_spec.rb
jwright: I feel that is fine as I think the extra output makes it less readable.

app/controllers/submission_files_controller.rb
jwright: I implemented this in another project recently. You do not need a view if you set the response type to `csv`. You can either specify that in a route with `defaults: { format: :csv }` or using `respond_to` in the controller.

jwright:
I added a few other comments. I think we should review the proctor stuff a bit before the next PR to make sure we are on the same page there.

app/controllers/submission_files_controller.rb
zazaian: IMO declaring a format of CSV is actually more of a hack than leaving the template in place, which I suppose actually isn't a hack at all but rather a design limitation. Let's leave this alone for now and if it causes any problems we can re-visit the send_file solution.

zazaian:
@jwright sounds good. I can give you guys a more thorough overview of how this is intended to work and we can come up a solution that works for everybody. Let's address the proctor implementation tomorrow.

spec/lib/formatter/filename_spec.rb
zazaian: This is purely a matter of taste. I dislike having laundry lists of spec example output with no organization of examples beyond the heading. I'm going to leave this as-is.

zazaian:
@chcholman bag it.

PR #2002, opened 2016-05-22
Adding rubric grade exporter

chcholman:
This PR adds a Rubric Exporter to allow the instructor to download the rubric for importing into other systems, safe-keep, etc. The download is a .csv file that contains each criteria as a row, and each level as a column on that row. The level has the following data pattern: Level Name (100 points) Full description of the level requirements. 

It also adds a CriterionGrades Exporter to allow the instructor to download student results for the assignment.


app/exporters/rubric_exporter.rb
```diff
@@ -0,0 +1,32 @@
...
+
+  def base_criterion_data(criterion)
+    [criterion.id, criterion.name]
+  end
+
+  def level_data_for(criterion)
+    level_data = []
+    # add the levels for the criteria
+    criterion.levels.ordered.inject(level_data) do |memo, level|
+      memo << "#{level.name} (#{level.points} points) #{level.description}"
```
acousticrobot: Pretty minor thing, but this line and the same line in the `rubric_grades_exporter`, will leave a space before the comma whenever there is no description. 

```
,Criteria #1,No Credit (0 points) ,Level #5 (95000 points) ,etc.
```
I don't think it's an issue since 1) it's a csv file, and 2) I assume most real rubric will have descriptions.

app/exporters/rubric_grade_exporter.rb
```diff
@@ -0,0 +1,44 @@
+class RubricGradesExporter
```
acousticrobot: Maybe you meant to remove this file? It currently is not used anywhere in the app, has some student specific methods that aren't accessible, and Rails wont find it with the plural/singular namespace discrepancies ( rubric_grade_exporter vs. RubricGradesExporter).

app/exporters/rubric_grade_exporter.rb
```diff
@@ -0,0 +1,44 @@
...
+  def export(rubric)
```
acousticrobot: Maybe you meant to remove this file? It isn't used in the app, has some inaccessible student specific methods, and wont be found by Rails with the singular/plural name discrepancies (rubric_grade_exporter vs. RubricGradesExporter).

app/controllers/rubrics_controller.rb
```diff
@@ -28,6 +28,14 @@ def update
     respond_with @rubric, status: :not_found
   end
 
+  def export
+    assignment = current_course.assignments.find params[:assignment_id]
+    rubric = assignment.rubric
+    respond_to do |format|
+      format.csv { send_data RubricExporter.new.export rubric }
```
acousticrobot: Jamie has advocated in the API routes for removing the respond to blocks on controller actions that only respond to one format. I'm not sure what the best way to do this with csv is though (the `send_data` implementations I've seen are much uglier than what we have now.) Thoughts @jwright?

jwright: I think `respond_to` is ok if it's needed. For this case, we do not respond to any other `Content-Type`s so, I would simply remove the `respond_to` block here entirely. The reason is because other developers looking at this will look for tests and views for other formats.

In addition, do we actually respond to both `html` and `json` like it says on line 6.

app/exporters/rubric_exporter.rb
jwright: Should this be `CriterionExporter`? I am always confused as to when to use `Rubric` and when to use `Criterion` in naming :)

chcholman: @jwright A Rubric is a collection of all criteria for that assignment - so this is either the Rubric Exporter, or the Criteria Exporter - I'm up for either.

jwright: @chcholman I think `RubricExporter` is fine.

jwright:
Looks great to me.

chcholman:
Thank you @acousticrobot and @jwright!

PR #2005, opened 2016-05-24
move assignment weights JSON to api

acousticrobot:
This moves the final JSON GET requests from the Predictor to the api namespace: Assignment Types and Assignment Type Weights.

It also does the following:

  * Combines both into one api call
  * Standardizes the weight properties in Angular to use the same names as in the Rails Course model
  * Moves weights properties into the json `meta` and  loads each one explicitly, rather than wrapping them in a `weights` namespace in the json

This will be followed by an refactor of the weights `update` route.


app/controllers/api/assignment_types_controller.rb
```diff
@@ -0,0 +1,15 @@
+class API::AssignmentTypesController < ApplicationController
+  include PredictorData
+
+  # GET api/assignment_types
+  def index
+    if current_user.is_student?(current_course)
```
jwright: Should we use `current_user_is_student?` here?

app/controllers/assignment_type_weights_controller.rb
```diff
@@ -64,32 +64,6 @@ def update
     end
   end
 
-  # Faculty have access to student weights, so we send the same information to
```
jwright: 👍

app/controllers/assignment_types_controller.rb
```diff
@@ -117,27 +117,6 @@ def destroy
     }
   end
 
-  def predictor_data
```
jwright: 😮

config/routes.rb
```diff
@@ -357,34 +358,27 @@
...
 
-    #17b. Predictor, Student View
+    # Student Predictor View, Predictor Preview
     resources :predicted_earned_badges, only: [:index, :update]
     resources :predicted_earned_challenges, only: [:index, :update]
     resources :predicted_earned_grades, only: [:index, :update]
 
-    #17c. Predictor, Instructor View
+    # Instructor View of Student's Predictor
     resources :students, only: [] do
```
jwright: I believe these can be simplified with a `module: :students`.

```
resources :students, only: [], module: :students do
  get "assignment_types", to: "assignment_types#index"
  # ...
end
```

jwright:
Beutimus! Nice refactor. I just had a few comments.

app/controllers/api/assignment_types_controller.rb
acousticrobot: Good idea, I'm going to make this change throughout the app, it looks like there are 3 other places where it still needs to be changed.

config/routes.rb
acousticrobot: Very nice!

PR #2007, opened 2016-05-24
Display Badges in Individual Columns

jonathoy:


app/views/info/awarded_badges.html.haml
```diff
@@ -18,21 +18,25 @@
       %tr
         %th{scope: "col"} First Name
         %th{scope: "col"} Last Name
-        %th{scope: "col"}= "#{term_for :badges} Earned"
         %th{:style => "display: none", } Badge Count
```
acousticrobot: We should change the text to `"#{term_for :badges} Count"`

app/controllers/info_controller.rb
```diff
@@ -30,6 +30,7 @@ def timeline_events
   def awarded_badges
```
acousticrobot: Can change this name to `earned_badges`, or change the route to `awarded_badges`?

chcholman: Let's change the route to `awarded_badges`

jonathoy:
I made a couple changes concerning the proposed route change. Let me know if it makes sense.

PR #2011, opened 2016-05-25
1747 grade routes edit refactor

jwright:
Moves the `#edit` and `#update` endpoints for a `GradesController` to use a `grade_id` instead of an `assignment_id` and `student_id`. This cleaned up the controller methods quite a bit and used resourceful routing for the two actions.

For rubric grades, an update was necessary for the Angular code so it no longer gathered the `assignment_id` and `student_id` or `group_id` from the url. Instead, these are now passed in via `ng-init` to the front end from the erb view.

In addition, the old grade update path performed a `Grade.find_or_create` in the action. This is not possible in the new `Grade#edit` path because a grade id is required. So the path `POST /assignments/:assignment_id/students/:student_id/grade` path was created to perform a `Grade.find_or_create` and then redirect to `Grade#edit` when the grade is not known.

### DIFF
--- a/app/assets/javascripts/angular/services/RubricService.js.coffee
+++ b/app/assets/javascripts/angular/services/RubricService.js.coffee
@@ -29,7 +29,7 @@
     )
 
   getCriterionGrades = (assignment)->
-    if assignment.scope.type == "STUDENT"
+    if assignment.scope.type == "student"
       $http.get('/api/assignments/' + assignment.id + '/students/' + assignment.scope.id + '/criterion_grades/').success((res)->
         addCriterionGrades(res.data)
       )
@@ -50,7 +50,7 @@
     )
 
   getGrade = (assignment)->
-    if assignment.scope.type == "STUDENT"
+    if assignment.scope.type == "student"
       $http.get('/api/assignments/' + assignment.id + '/students/' + assignment.scope.id + '/grade/').success((res)->
         angular.copy(res.data.attributes, grade)
         angular.copy(res.meta.grade_status_options, gradeStatusOptions)
@@ -67,7 +67,7 @@
       )
 
   putRubricGradeSubmission = (assignment, params, returnURL)->
-    scopeRoute = if assignment.scope.type == "STUDENT" then "students" else "groups"
+    scopeRoute = if assignment.scope.type == "student" then "students" else "groups"
     $http.put("/api/assignments/#{assignment.id}/#{scopeRoute}/#{assignment.scope.id}/criterion_grades", params).success(
       (data)->
         console.log(data)
--- a/app/assets/javascripts/angular/services/RubricService.js.coffee
+++ b/app/assets/javascripts/angular/services/RubricService.js.coffee
@@ -14,24 +14,11 @@
 
   getAssignment = (location)->
     assignment.id = parseInt(location.pathname.split('/')[2])
-    # GRADING A STUDENT FOR AN ASSIGNMENT
-    if location.search.match(/student_id=/)
-      assignment.scope = {
-        type: "STUDENT",
-        id: parseInt(window.location.search.match(/student_id=(\d+)/)[1])
-      }
-    # GRADING A GROUP FOR AN ASSIGNMENT
-    else if location.pathname.split('/')[3] == "groups"
-      assignment.scope = {
-        type: "GROUP",
-        id: parseInt(location.pathname.split('/')[4])
-      }
     # DESIGNING A RUBRIC
-    else
-      assignment.scope = {
-        type: "DESIGN_MODE",
-        id: null
-      }
+    assignment.scope = {
+      type: "DESIGN_MODE",
+      id: null
+    }
 
   # TODO: $scope should not be passed around if we want to avoid tight coupling
   getCriteria = (assignment, $scope)->
--- a/app/assets/javascripts/angular/services/RubricService.js.coffee
+++ b/app/assets/javascripts/angular/services/RubricService.js.coffee
@@ -12,14 +12,6 @@
   badges = {}
   criterionGrades = {}
 
-  getAssignment = (location)->
-    assignment.id = parseInt(location.pathname.split('/')[2])
-    # DESIGNING A RUBRIC
-    assignment.scope = {
-      type: "DESIGN_MODE",
-      id: null
-    }
-
   # TODO: $scope should not be passed around if we want to avoid tight coupling
   getCriteria = (assignmentId, $scope)->
     _scope = $scope
--- a/app/assets/javascripts/angular/services/RubricService.js.coffee
+++ b/app/assets/javascripts/angular/services/RubricService.js.coffee
@@ -33,7 +33,7 @@
       $http.get('/api/assignments/' + assignment.id + '/students/' + assignment.scope.id + '/criterion_grades/').success((res)->
         addCriterionGrades(res.data)
       )
-    else if assignment.scope.type == "GROUP"
+    else if assignment.scope.type == "group"
       $http.get('/api/assignments/' + assignment.id + '/groups/' + assignment.scope.id + '/criterion_grades/').success((res)->
 
         # The API sends all student information so we can add the ability to custom grade group members
@@ -56,7 +56,7 @@
         angular.copy(res.meta.grade_status_options, gradeStatusOptions)
         thresholdPoints = res.meta.threshold_points
       )
-    else if assignment.scope.type == "GROUP"
+    else if assignment.scope.type == "group"
       $http.get('/api/assignments/' + assignment.id + '/groups/' + assignment.scope.id + '/grades/').success((res)->
 
         # The API sends all student information so we can add the ability to custom grade group members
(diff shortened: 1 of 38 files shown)


app/views/assignments/individual/_table_body.html.haml
```diff
@@ -66,15 +66,15 @@
...
             - else
-              %li= link_to glyph(:edit) + "Edit Grade", edit_assignment_grade_path(presenter.assignment, student_id: student.id), class: "button"
+              %li= link_to glyph(:edit) + "Edit Grade", edit_grade_path(grade), class: "button"
             %li= link_to glyph(:trash) + "Delete Grade", remove_grade_path(grade), class: "button", data: { confirm: "Are you sure?" }, :method => :post
           - else
             - if presenter.assignment.is_unlockable? && !presenter.assignment.is_unlocked_for_student?(student)
               %li= link_to glyph("unlock-alt") + "Unlock", manually_unlock_unlock_state_path(student_id: student.id, assignment_id: presenter.assignment.id), :method => :post, class: "button"
             - has_submission = presenter.assignment.accepts_submissions? && !submissions_for_assignment.empty?
-            %li= link_to glyph(:check) + "Grade", edit_assignment_grade_path(presenter.assignment, student_id: student.id), class: "button #{has_submission ? "danger" : ""}"
+            %li= link_to glyph(:check) + "Grade", edit_grade_path(grade), class: "button #{has_submission ? "danger" : ""}"
```
acousticrobot: I'm getting an `UrlGenerationError` error here, I believe because `Assignments::Presenter#grade_for`, calls Grade.new, but the grade is not saved, so it doesn't have an id. I couldn't trigger this error on `app/views/assignments/group/_table_body.html.haml:59` with the sample data, but it may be an issue there as well.

app/assets/javascripts/angular/services/RubricService.js.coffee
```diff
@@ -14,24 +14,11 @@
 
   getAssignment = (location)->
     assignment.id = parseInt(location.pathname.split('/')[2])
```
acousticrobot: `app/views/rubrics/design.html.haml` passes `rubricId` to `/angular/controllers/RubricCtrl.js.coffee` as the first parameter of init(), but it isn't even used. If we change this to the `assignment_id` like you did with `_rubric_edit.html.haml` we can get rid of this method entirely, and drop `assignment` from the service, since it is passed to all of the methods anyway. Doesn't even look like the "DESIGN_MODE" is ever used in the services, what a mess. I can take a crack at it if you want, maybe even after this PR is accepted.

app/controllers/grades_controller.rb
```diff
@@ -18,52 +17,34 @@ def show
     @title = "#{name}'s Grade for #{@grade.assignment.name}"
   end
 
-  # GET /assignments/:assignment_id/grade/edit?student_id=:id
+  # GET /grades/:id/edit
```
acousticrobot: 👏

app/controllers/grades_controller.rb
```diff
@@ -18,52 +17,34 @@ def show
...
+    @grade = Grade.find params[:id]
+    @title = "Editing #{@grade.student.name}'s Grade "\
+      "for #{@grade.assignment.name}"
+    @badges = @grade.student.earnable_course_badges_for_grade(@grade)
+    @submission = @grade.student.submission_for_assignment(@grade.assignment)
   end
 
   # To avoid duplicate grades, we don't supply a create method. Update will
   # create a new grade if none exists, and otherwise update the existing grade
   # PUT /assignments/:assignment_id/grade
```
acousticrobot: update comment: `# PUT grade/:id`

app/views/assignments/groups/grade.html.haml
```diff
@@ -15,6 +15,6 @@
         = prop.title
 
   - if @rubric
-    = render "/grades/rubric_edit"
+    = render partial: "/grades/rubric_edit", locals: { rubric: @rubric, scope_type: "GROUP", scoped_id: @group.id, return_path: URI(request.referer).path }
```
acousticrobot: I originally named "GROUP" and "STUDENT" using conventions for constants, but we could change the `type` to "groups" and "students" to align with our json api standards. We'd change it in `app/views/assignments/groups/grade.html.haml`, `app/views/grades/edit.html.haml`, and the service call methods in `/angular/services/RubricService.js.coffee`. Thoughts?

app/views/grades/_standard_edit.html.haml
```diff
@@ -3,31 +3,30 @@
...
 
         .panel(ng-cloak)
-          - if @assignment.pass_fail?
+          - if grade.assignment.pass_fail?
             .switch-label
               = "#{term_for :pass} / #{term_for :fail} Status"
             .pass-fail-grade-toggle
               .switch.pass-fail-contingent{data: {:on => "#{term_for :pass}", :off => "#{term_for :fail}"}}
-                = f.check_box :pass_fail_status, { :checked=> @grade.pass_fail_status == "Pass" }, checked_value = "Pass", unchecked_value = "Fail"
-                - label_text = @grade.pass_fail_status ? (term_for @grade.pass_fail_status) : (term_for :fail)
```
acousticrobot: at-at

app/controllers/assignments/groups_controller.rb
```diff
@@ -9,12 +9,7 @@ def grade
     @submission_id = @assignment.submissions.where(group_id: @group.id).first.try(:id)
     @title = "Grading #{ @group.name }'s #{@assignment.name}"
     @assignment_score_levels = @assignment.assignment_score_levels
-
-    if @assignment.grade_with_rubric?
-      @rubric = @assignment.rubric
-      # This is sent to the Angular controlled submit button
-      @return_path = URI(request.referer).path
```
chcholman: I'm concerned about losing the redirection here - is this being handled in a different way?

app/controllers/grades_controller.rb
```diff
@@ -18,52 +17,34 @@ def show
...
-    @badges = @student.earnable_course_badges_for_grade(@grade)
-    @assignment_score_levels =
-      @assignment.assignment_score_levels.order_by_value
-
-    if @assignment.grade_with_rubric?
-      @rubric = @assignment.rubric
-      @criterion_grades = serialized_criterion_grades
-      # This is sent to the Angular controlled submit button
-      @return_path =
-        URI(request.referer).path + "?student_id=#{current_student.id}"
```
chcholman: Same note here as above.

app/views/assignments/individual/_table_body.html.haml
jwright: @acousticrobot  I believe that `app/views/assignments/group/_table_body.html.haml:59` will never be a `NullGrade` since it's for a group which means that the grade will always have a grade. There is also a check to ensure that the grade is not nil. Why does `NullGrade#is_graded?` return a `true` value?

acousticrobot: I don't think what I'm seeing is an error related to NullGrades.  I'm still able to get this error on this branch running sample data, and going to an assignment show page for any non-team assignment with no grades (In course 1, "STANDARD EDIT + QUICK GRADE WITH CHECKBOX [NO GRADES]", `http://localhost:5000/assignments/13`). I just tried calling class on the "grade.class" within Better Errors and get `Grade`.

I think the error originates on `app/presenters/assignments/presenter.rb` line 32

```
  def grade_for(student)
    grades.where(student_id: student.id).first ||
      Grade.new(assignment_id: assignment.id)
  end
```

because we're sending in a new, unsaved grade, so it doesn't have an id yet.

I'm not sure why `NullGrade#is_graded?` is true, not false, but I think it needs to be true to work with the GradeProctor:

https://github.com/UM-USElab/gradecraft-development/commit/5a180582e9ab733e929cb0bae088df32efcf4c03

jwright:
@acousticrobot Thank you for your review.

1. I removed the `RubricService#getAssignment` by passing in the `assignmentId` like you suggested. Awesome!
2. I replaced `edit_grade_path` with `assignment_student_grade_path` which does a `Grade.find_or_create` before redirecting to `edit_grade_path` to ensure a grade exists for a few other places than the one you suggested. I replaced where I did not know for certain that a grade would exist. I only kept `edit_grade_path` where I knew a grade would not exist (like it is coming from the database, etc). I am not sure this is the best approach. Perhaps we should just use `assignment_student_grade_path` everywhere?

I am passing this back to you for one more look to ensure that the last commits are good for you. If you are good, it can probably be passed onto cait.

app/assets/javascripts/angular/services/RubricService.js.coffee
```diff
@@ -12,31 +12,10 @@
...
-        type: "GROUP",
-        id: parseInt(location.pathname.split('/')[4])
-      }
-    # DESIGNING A RUBRIC
-    else
-      assignment.scope = {
-        type: "DESIGN_MODE",
-        id: null
-      }
-
```
acousticrobot: 💯 excellent!

PR #2017, opened 2016-05-27
Note check

chcholman:
This PR removes the requirement that instructors declare per assignment whether or not students should receive an email receipt for a new grade. 

This feature will be brought back in the reverse order - students will be able to decide if they want to receive emails of notifications or turn them off and see them only in GC Announcements.


jwright:
Beutimus!

Issue #1853, opened 2016-04-01
Make tables more responsive for mobile, etc


mariehooper:

mariehooper:
implement stackable.js

PR #2019, opened 2016-05-27
1853 implement stacktablejs

mariehooper:
This PR implements Stacktable.js plugin to improve responsiveness of tables throughout app. Several variations of Stacktable.js's cardtable layout are provided: (1) for tables with both styled header and footer with button bar, (2) with neither header nor button bar, (3) with just header and (4) with just footer.

### Before
before-tables

### After
after-tables

### What's Next
Some tables would benefit from an additional cell that concatenates first and last name so headers can present full name. Also - there is a "Due Date" field that is originally hidden that will need to be re-hidden on mobile tables.

Closes issue #1853

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -17,6 +17,7 @@
 //= require underscore
 //= require jquery.dynatable
 //= require dynatable
+//= require stacktable.min
 //= require lodash.min
 
 //= require angular/angular
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -18,6 +18,7 @@
 //= require jquery.dynatable
 //= require dynatable
 //= require stacktable.min
+//= require mobile_tables
 //= require lodash.min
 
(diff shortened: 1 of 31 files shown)


chcholman:
@mariehooper This is looking so awesome! A few minor tweaks and this will be ready to merge:

1) Add this to the Badges index page
2) On the Student index, there's an empty row appearing between the last name and the screenname
3) Title case all column headings on mobile (Check the Team index for an example of where this breaks)
4) The Staff index shows an empty row at the top on mobile
5) What's your thinking about the buttons? On mobile right now they are neither merged nor separate - do you want to restyle them to be fully separate? (May make sense to do in another ticket, but let's create that issue if so)
6) Can we standardize our treatment of names? We've got a few where the first name takes the top (blue) row, and a few where we've done no blue row at all. I think we want the latter until we can come up with a better solution (happy to discuss if you think the other is better!)

chcholman:
@mariehooper Looks awesome, nice job!

PR #2023, opened 2016-05-28
1620 team filter

chcholman:
This PR pulls the team filter out into a single helper that can be referenced as:

`= team_filter`

### DIFF
--- a/app/helpers/team_filter_helper.rb
+++ b/app/helpers/team_filter_helper.rb
@@ -2,7 +2,7 @@ module TeamFilterHelper
 
   def team_filter(teams)
     content_tag :div, class: 'team-filter' do
-      form_tag(request.url, name: "team_selector", onchange: ("javascript: document.team_selector.submit();"), method: :get) do
+      form_tag(request.url, class: "team-selector", onchange: "$('.team-selector').submit();", method: :get) do
         label_tag :team_id, "Select #{(term_for :team).titleize}", class: 'sr-only'
         select_tag(:team_id, options_for_select(teams.alpha.map { |t| [t.name, t.id] }, params[:team_id]), prompt: "– Select #{(term_for :team).titleize} –")
       end
--- a/app/helpers/team_filter_helper.rb
+++ b/app/helpers/team_filter_helper.rb
@@ -1,10 +1,10 @@
 module TeamFilterHelper
 
-  def team_filter
+  def team_filter(teams)
     content_tag :div, class: 'team-filter' do
-      form_tag(request.url, name: "see_team", onchange: ("javascript: document.see_team.submit();"), method: :get) do
+      form_tag(request.url, name: "team_selector", onchange: ("javascript: document.team_selector.submit();"), method: :get) do
         label_tag :team_id, "Select #{(term_for :team).titleize}", class: 'sr-only'
-        select_tag(:team_id, options_for_select(current_course.teams.map { |t| [t.name, t.id] }, params[:team_id]), prompt: "– Select #{(term_for :team).titleize} –")
+        select_tag(:team_id, options_for_select(teams.alpha.map { |t| [t.name, t.id] }, params[:team_id]), prompt: "– Select #{(term_for :team).titleize} –")
       end
     end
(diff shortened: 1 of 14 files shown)


app/helpers/team_filter_helper.rb
```diff
@@ -0,0 +1,11 @@
+module TeamFilterHelper
+
+  def team_filter
+    content_tag :div, class: 'team-filter' do
+      form_tag(request.url, name: "see_team", onchange: ("javascript: document.see_team.submit();"), method: :get) do
```
jwright: I think we should change the markup to be standardized. I vote for changing `see_team` to `see-team` or `team-selector` or I believe we can remove the name and change the js to `this.form()` but I could be wrong.

In addition, does this url work for all these instances?

app/helpers/team_filter_helper.rb
```diff
@@ -0,0 +1,11 @@
...
+        label_tag :team_id, "Select #{(term_for :team).titleize}", class: 'sr-only'
+        select_tag(:team_id, options_for_select(current_course.teams.map { |t| [t.name, t.id] }, params[:team_id]), prompt: "– Select #{(term_for :team).titleize} –")
```
jwright: This uses `current_course.teams` but sometimes there is a `@teams` instance variable. I think it would be better if this method took in the teams to display.

app/views/assignment_types/all_grades.html.haml
```diff
@@ -7,12 +7,7 @@
 .pageContent
   = render "layouts/alerts"
 
-  - if @teams.present?
-    .team-filter
-      = form_tag all_grades_assignment_type_path(@assignment_type), name: "see_team", :onchange => ("javascript: document.see_team.submit();"), :method => :get do
-        %label
-          .sr-only Select #{term_for :team}
-        = select_tag :team_id, options_for_select(@teams.map { |t| [t.name, t.id] }, @team.try(:id)), :prompt => "– Select #{(term_for :team).titleize} –"
+  = team_filter if @teams.present?
```
jwright: This could change to `team_filter(@teams)`

app/views/assignments/individual/_individual_show.haml
```diff
@@ -1,9 +1,4 @@
-- if presenter.has_teams? && presenter.individual_assignment?
-  .team-filter
-    = form_tag assignment_path(presenter.assignment), name: "see_team", onchange: ("javascript: document.see_team.submit();"), method: :get do
-      %label.sr-only{:for => "team_id"}
-        Select #{term_for :team} –
-      = select_tag :team_id, options_for_select(presenter.teams.map { |t| [t.name, t.id] }, presenter.team.try(:id)), prompt: "– Select #{(term_for :team).titleize} –"
+= team_filter if presenter.has_teams? && presenter.individual_assignment?
```
jwright: This could change to `team_filter(presenter.teams)`

jwright:
Awesome. So happy to see this. I have been wanting to do this for a long time :)

app/helpers/team_filter_helper.rb
chcholman: This js currently breaks if you use a dash as opposed to an underscore, so for now I'm updating it to be `team_selector`. I tried a couple of versions of revising to be `this.form()` but none worked. 

The URL works for all instances I'm aware of.

app/helpers/team_filter_helper.rb
```diff
@@ -0,0 +1,11 @@
+module TeamFilterHelper
+
+  def team_filter(teams)
+    content_tag :div, class: 'team-filter' do
+      form_tag(request.url, name: "team_selector", onchange: ("javascript: document.team_selector.submit();"), method: :get) do
```
jwright: I think we should change this to be more idiomatic to javascript instead of ruby and use `team-selector`. Thoughts?

jwright: You could change it to `$(".team-selector").submit()`

jwright:
Small change suggestion but everything looks great!

PR #2032, opened 2016-06-01
Refactor/weights update

acousticrobot:
### Weights Refactor Part 1

This pull request moves `Weights` off of `Assignments` in the logic, and moves them onto Assignment Types.  This includes changing the model to `AssignmentTypeWeight`, and standardizing the naming in the Angular and Ruby. The weights terms stored on course have been changed throughout the app:

     assignment_weight_close_at => weights_close_at
     default_assignment_weight => default_weight
     total_assignment_weight => total_weights
     max_assignment_weight => max_weights_per_assignment_type

### Preparation for Part 2

The `mass_upadate` route for weights relies on two strange constructs, an `AssignmentTypeWeight` struct in `\models` and an `AssignmentTypeWeightsForm` in the lib folder. As this route currently works with the changes in this PR, I am saving this refactor for part 2. As a temporary measure, the struct has been renamed `AssignmentTypesWeightStruct` so the namespace doesn't conflict with the actual `AssignmentTypeWeight` model.


app/assets/javascripts/angular/services/PredictorService.js.coffee
```diff
@@ -96,48 +96,48 @@
...
         )
         termFor.challenges = res.meta.term_for_challenges
         update.challenges = res.meta.update_challenges
       )
 
-    postPredictedGrade = (student_id, value)->
+    postPredictedGrade = (studentId, value)->
       if update.assignments
-        $http.put('/api/predicted_earned_grades/' + student_id, predicted_points: value).success(
+        $http.put('/api/predicted_earned_grades/' + studentId, predicted_points: value).success(
```
jwright: This is a weird url since we are passing in the `studentId` as the resource if for `predicted_earned_grades`. I think this should change to `/api/students/:student_id/predicted_earned_grades`. I don't think we should change this in this PR but perhaps in another one?

app/assets/javascripts/angular/services/PredictorService.js.coffee
```diff
@@ -96,48 +96,48 @@
...
           ).error(
             (data)->
               console.log(data);
           )
 
-    postPredictedChallenge = (student_id, value)->
+    postPredictedChallenge = (studentId, value)->
       if update.challenges
-        $http.put('/api/predicted_earned_challenges/' + student_id, predicted_points: value).success(
+        $http.put('/api/predicted_earned_challenges/' + studentId, predicted_points: value).success(
```
jwright: Same comment here as above.

app/assets/javascripts/angular/services/PredictorService.js.coffee
```diff
@@ -96,48 +96,48 @@
...
           ).error(
             (data)->
               console.log(data);
           )
 
-    postPredictedBadge = (student_id, value)->
+    postPredictedBadge = (studentId, value)->
       if update.badges
-        $http.put('/api/predicted_earned_badges/' + student_id, predicted_times_earned: value).success(
+        $http.put('/api/predicted_earned_badges/' + studentId, predicted_times_earned: value).success(
```
jwright: Same comment as https://github.com/UM-USElab/gradecraft-development/pull/2032/files#diff-7eedc7c5bb4ea4ad6dd55def73e4a3a6R110

app/controllers/api/assignment_type_weights_controller.rb
```diff
@@ -0,0 +1,33 @@
+class API::AssignmentTypeWeightsController < ApplicationController
+
+  before_filter :ensure_student?, only: [:create]
```
jwright: We should remove the `only` option here since `create` is the only action on this controller.

app/controllers/api/assignment_type_weights_controller.rb
```diff
@@ -0,0 +1,33 @@
...
+
+  # POST /api/assignment_types/:assignment_type_id/assignment_type_weights
+  def create
+    assignment_type = current_course.assignment_types.find(params[:assignment_type_id])
+    weight = params[:weight]
+
+    if assignment_type && weight && assignment_type.student_weightable?
+      assignment_type_weight = assignment_type.weights.where(student: current_student).first_or_initialize
```
jwright: I think we should use `current_user` here since that is who we are referring to. I would like to remove references to `current_student` or change it's implementation in the future so it does not refer to the query parameter. Thoughts?

app/assets/javascripts/angular/services/PredictorService.js.coffee
acousticrobot: I was just changing the snake case to camel case on all the function attributes in the file -- but you're right this should actually just be `id` or `resource_id` on all of these service calls.  I've actually already changed these routes to use the id of the resource and not the student. I'll change this before merge since `student_id` it's misleading and out of date.

app/controllers/assignment_type_weights_controller.rb
```diff
@@ -40,30 +40,6 @@ def mass_update
     end
   end
 
-  # Updates weights from the predictor
```
jwright: 👍

app/controllers/assignments_controller.rb
```diff
@@ -115,19 +113,4 @@ def export_structure
       format.csv { send_data AssignmentExporter.new.export current_course }
     end
   end
-
-  private
-
-  def set_assignment_weights(assignment)
```
jwright: Yay!

app/models/assignment.rb
```diff
@@ -132,8 +128,8 @@ def point_total_for_student(student, weight = nil)
   # course doesn't have weights
   def weight_for_student(student, weight = nil)
```
jwright: I feel like this should be moved to `AssignmentType` since we do not really use the `Assignment` at all in this method.

app/models/assignment_type_weight_struct.rb
```diff
@@ -0,0 +1,37 @@
+class AssignmentTypeWeightStruct < Struct.new(:student, :assignment_type)
```
jwright: I am unclear why we need this struct. Can we just use `AssignmentTypeWeight`?

app/views/assignment_type_weights/mass_edit.html.haml
```diff
@@ -9,16 +9,16 @@
...
         %h4 RULES OF THE GAME
         %ul
-          - if current_course.total_assignment_weight?
-            %li You have a total of #{current_course.try(:total_assignment_weight)} #{(term_for :weights).titleize} to spend
-          - if current_course.assignment_weight_close_at?
-            %li You have until #{current_course.try(:assignment_weight_close_at).strftime("%A, %B %d, %Y, at %H:%M%p")} to make your #{(term_for :weight).titleize} decision
-          - if current_course.max_assignment_weight?
-            %li You may place up to #{current_course.max_assignment_weight} #{ ( term_for :weights).titleize } on a single #{ (term_for :assignment_type).downcase}
+          - if current_course.total_weights?
+            %li You have a total of #{current_course.try(:total_weights)} #{(term_for :weights).titleize} to spend
```
jwright: We can remove the `try` here since we test for it's existence in the previous line.

app/views/assignment_type_weights/mass_edit.html.haml
```diff
@@ -9,16 +9,16 @@
...
+          - if current_course.weights_close_at?
+            %li You have until #{current_course.try(:weights_close_at).strftime("%A, %B %d, %Y, at %H:%M%p")} to make your #{(term_for :weight).titleize} decision
```
jwright: We can remove the `try` here since we test for it's existence in the previous line.

db/migrate/20160530000000_move_weights_to_assignment_type.rb
```diff
@@ -0,0 +1,9 @@
+class MoveWeightsToAssignmentType < ActiveRecord::Migration
+  def change
+    #rename_index :assignment_weights, "index_assignment_weights_on_student_id_and_assignment_type_id", "index_weights_on_student_and_assignment_type"
```
jwright: Can we remove this comment?

spec/models/assignment_type_weight_struct_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+      assignment_type = create(:assignment_type, course: course)
+      student = create(:user)
+      assignment_type_weight = create(:assignment_type_weight, student: student, assignment_type: assignment_type, course: course, weight: 3)
+      subject = AssignmentTypeWeightStruct.new(student, assignment_type)
+
+      expect(subject.assignment_type_id).to eq(assignment_type.id)
+    end
+  end
+
+  describe "#save" do
```
jwright: Can we remove this describe block and comments?

jwright:
@acousticrobot Looks great. I had several small comments. Most notably is the existence of `AssignmentTypeWeightStruct`.

app/models/assignment_type_weight_struct.rb
```diff
@@ -0,0 +1,37 @@
...
+  end
+
+  def persisted?
+    false
+  end
+
+  private
+
+  def save_assignment_type_weights
+      assignment_weight = assignment_type.weights.where(student: student).first_or_initialize
```
jwright: Rubocop is complaining about this spacing (4 vs 2 spaces).

app/controllers/api/assignment_type_weights_controller.rb
acousticrobot: I would be OK with removing reliance on the `current_student` or changing the implementation.

app/models/assignment.rb
acousticrobot: Good call -- AssignmentType already has a `weight_for_student` so I can use it and remove it here.

PR #2039, opened 2016-06-03
rake task to update nill final score grades

acousticrobot:
This Rake task addresses Grades from before March 2016, when we did not persist the final_score field. 

Currently there is a [patch method in Grade](https://github.com/UM-USElab/gradecraft-development/blob/master/app/models/grade.rb#L90) to address this, which defaults to `raw_score`. Before we change `final_score` to `final_points` and remove this method, we need to update old grades to work without this patch method.


lib/tasks/grades.rake
```diff
@@ -14,6 +14,11 @@ namespace :grades do
     Grade.find_each(batch_size: 500) { |g| g.update_column(:graded_at, g.graded_at) }
   end
 
+  desc "Update grades without a final score to use raw score"
+  task update_final_score: :environment do
+    Grade.where("final_score is NULL and raw_score is not NULL").update_all("final_score=raw_score")
```
jwright: Do you know how many grades this is on production?

If it is a lot (> 4000), we probably do not want to load all of those in memory and we may want to do a [find_in_batches](http://api.rubyonrails.org/classes/ActiveRecord/Batches.html#method-i-find_in_batches) here.

jwright:
Just a quick comment on the number of grades this will bring into memory. If it's not an issue, feel free to push this.

PR #2042, opened 2016-06-03
Fixed tab display bug on group show view

chcholman:
This PR fixes a bug where the assignment presenter method `has_grades?` was returning true the second time it was called (below a table where grades were created but not persisted). This display of tab content without it's associated tab broke the tab interaction.

### DIFF
--- a/app/presenters/assignments/presenter.rb
+++ b/app/presenters/assignments/presenter.rb
@@ -61,7 +61,7 @@ class Assignments::Presenter < Showtime::Presenter
   end
 
   def has_persisted_grades?
-    grades { |grade| grade if grade.persisted? }
+    grades.any?(&:persisted?)
   end
 
   def has_reviewable_grades?
--- a/app/presenters/assignments/presenter.rb
+++ b/app/presenters/assignments/presenter.rb
@@ -60,6 +60,10 @@ class Assignments::Presenter < Showtime::Presenter
     grades.present?
   end
 
+  def has_persisted_grades?
+    grades { |grade| grade if grade.persisted? }
+  end
+
   def has_reviewable_grades?
     grades.instructor_modified.present?
   end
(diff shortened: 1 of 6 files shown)


app/presenters/assignments/presenter.rb
```diff
@@ -60,6 +60,10 @@ def has_grades?
     grades.present?
   end
 
+  def has_persisted_grades?
+    grades { |grade| grade if grade.persisted? }
```
jwright: This returns a collection of `Grade` objects instead of a boolean. Since this is a predicate method, I would expect this method to return a `true` or `false`.

This may work better:

```
def has_persisted_grades?
  grades.any?(&:persisted?)
end
```

I am not sure if this shorthand will work but if not, it can be expanded to:

```
def has_persisted_grades?
  grades.any? { |grade| grade.persisted? }
end
```

app/views/assignments/_show_instructor.haml
```diff
@@ -17,26 +17,22 @@
...
-
       - if presenter.grade_with_rubric?
-        .content#tabt3{role: "tabpanel", "aria-hidden" => false }
+        .ui-tabs-panel#tabt3{role: "tabpanel", "aria-hidden" => false }
           .tab-container
             = render partial: "rubrics/components/rubric_table", locals: { rubric: presenter.rubric, presenter: presenter, student: nil, include_grade_info: false }
-      - if presenter.has_grades?
+      - if presenter.has_persisted_grades?
         .ui-tabs-panel#tabt4{role: "tabpanel", "aria-hidden" => false }
-          - if presenter.group_assignment?
```
jwright: Why do we no longer have to check for a group assignment here?

jwright:
Just a few comments. Looks cleaner.

app/presenters/assignments/presenter.rb
chcholman: So much better! 👍

PR #2061, opened 2016-06-07
Production rake fix

acousticrobot:
This PR fixes the two grade tasks that need to be run on production, and adds a third.

`update_graded_at` has been changed (thanks @jwright) and should be rerun

`transfer_predictions` has been changed to run in batches

`update_final_score` was added and needs to be run in anticipation of value-points-score name changes. ( _This replaces [PR#2039](https://github.com/UM-USElab/gradecraft-development/pull/2039)_).


jwright:
👍

PR #2069, opened 2016-06-07
1857 resourceful routing

### LINKED ISSUE #1857: Resourceful Routing

jwright:
Cleans up all of the routes that are defined in the application. Some routes were duplicated. Some routes were not restricted (e.g. `resources :uploads` when only the `#create` method was defined). Some routes were not even being used. In addition, some routes were changed from using `get`, etc to use the `resources` format. This required changes to the route helpers in some situations.

I did check and triple check most of the route changes. There is a small fear of changes that were not covered by tests.

The routes were reduced by ~ 180 routes 😮 

#### on master

```
> bundle exec rake routes | wc -l
> 500
```

#### on this branch

```
> bundle exec rake routes | wc -l
> 319
```

Closes #1857


app/assets/javascripts/angular/controllers/GradeCtrl.js.coffee
```diff
@@ -75,7 +75,7 @@
         unearnedBadgesPostParams.push $scope.earnedBadgePostParams(badge)
     )
 
-    $http.post("/grade/#{$scope.grade.id}/earn_student_badges", {earned_badges: unearnedBadgesPostParams}).success(
+    $http.post("/api/grades/#{$scope.grade.id}/earned_badges", {earned_badges: unearnedBadgesPostParams}).success(
```
zazaian: So what are we ultimately looking to nest under the api namespace? Is this just going to be stuff that's accessed from frontend/javascript? This seems like a good pattern to me.

chcholman: Just wanted to note here that we need to start both taking data in from other applications and sending it out, if that affects the thinking here.

app/assets/javascripts/angular/factories/EarnedBadgeFactory.js.coffee
```diff
@@ -23,10 +23,5 @@
         $q.reject response.data
 
     deletePath: ()->
-      "/grade/" + this.grade_id + "/student/" + this.student_id + "/badge/" + this.badge_id + "/earned_badge/" + this.id
-
-    deleteParams: ()->
-      student_id: this.student_id,
-      badge_id: this.badge_id,
-      grade_id: this.grade_id
```
zazaian: :+1: This is a nice pattern. I think in general we should be very conscious about sending the minimum amount of information required to perform any given action. Obviously this cleans up the code but more importantly this feels like a move in the former direction. This looks a lot better.

app/assets/javascripts/angular/services/GradeSchemeElementsService.js.coffee
```diff
@@ -30,7 +30,7 @@
         elements[index+1].high_range = newValue-1
 
     getGradeSchemeElements = ()->
-      $http.get('/gse_mass_edit/').success((response) ->
+      $http.get('/grade_scheme_elements/mass_edit.json').success((response) ->
```
zazaian: Why are these `/grade_scheme_elements/` routes not being nested under the `/api` namespace? What's the distinction being drawn here?

app/assets/javascripts/angular/services/PredictorService.js.coffee
```diff
@@ -137,7 +137,7 @@
 
     postAssignmentTypeWeight = (assignmentType_id,value)->
       if update.weights
-        $http.post('/assignment_type_weight', id: assignmentType_id, weight: value).success(
+        $http.put('/assignment_type_weights/' + assignmentType_id, weight: value).success(
```
zazaian: This is cool, but I was already wondering about this as I thought about this PR. Did you run into any instances of ajax calls in the course of writing out this PR in which we're using POST instead of PUT? If this saves us lots of headache I think that's fine, but if we need to use that particular hack we should probably have a systemwide comment format for those so that we can find them easily if we'd like to replace them at some point. Maybe something simple like `# POST->PUT`

I know this is the opposite of what you're implementing here but it brought that matter to mind.

app/assets/javascripts/timeline_settings.js
```diff
@@ -3,7 +3,7 @@ $(document).ready(function(){
   if($("#dashboard-timeline").length) {
     $.ajax({
         type: 'GET',
-        url: '/info/timeline_events',
+        url: '/timeline_events',
```
zazaian: Again, what is the delineation of where `/api` is being used or not? Is that just for routes that are only being accessed via javascript calls?

app/controllers/api/grades/earned_badges_controller.rb
```diff
@@ -0,0 +1,19 @@
+class API::Grades::EarnedBadgesController < ApplicationController
+  # POST /api/grades/:grade_id/earned_badges
+  def create
+    render json: EarnedBadge.create(params[:earned_badges])
+  end
+
+  # DELETE /api/grades/:grade_id/earned_badges/delete_all
+  def delete_all
+    if !EarnedBadge.where(grade_id: params[:grade_id]).destroy_all.empty?
```
zazaian: This seems kind of stilted to me. What about the case of there being no earned badges for a given grade? It would return that earned badges have failed to delete with a 400, which isn't really an accurate response. We should probably have a third case here to return an outcome for an empty collection of EarnedBadge records for the given `grade_id`.

app/controllers/challenge_grades_controller.rb
```diff
@@ -1,8 +1,8 @@
 class ChallengeGradesController < ApplicationController
 
   before_filter :ensure_staff?, except: [:show]
-  before_action :find_challenge_grade, only: [:show, :edit, :update, :destroy]
-  before_action :find_challenge, only: [:show, :edit, :challenge, :destroy, :update ]
```
zazaian: Cool, this is much cleaner.

app/controllers/challenge_grades_controller.rb
```diff
@@ -52,10 +52,10 @@ def destroy
 
   def find_challenge
     find_challenge_grade
-    @challenge = @challenge_grade.challenge
+    @challenge ||= @challenge_grade.challenge
```
zazaian: Caching saves ~~lives~~ ~~cycles~~ lives.

app/controllers/grades_controller.rb
```diff
@@ -48,41 +48,6 @@ def update
     end
   end
 
-  # POST /grades/earn_student_badges
```
zazaian: image

app/controllers/level_badges_controller.rb
```diff
@@ -1,28 +1,13 @@
...
-    @level_badge = LevelBadge.create params[:level_badge]
-    respond_with @level_badge, layout: false,
-      serializer: ExistingLevelBadgeSerializer
+    level_badge = LevelBadge.create params[:level_badge]
+    render json: level_badge, serializer: ExistingLevelBadgeSerializer
   end
 
   def destroy
-    @level_badge.destroy
+    LevelBadge.find(params[:id]).destroy
```
zazaian: Should we be consistent with the `.where(id: params[:id]).first` nomenclature if we're going to use it?

app/views/assignment_type_weights/mass_edit.html.haml
```diff
@@ -5,7 +5,7 @@
 .pageContent
 
   - if current_course.assignment_weight_open? || current_user_is_staff?
-    = simple_form_for @form, as: :student, url: assignment_type_weights_path, method: :put do |f|
+    = simple_form_for @form, as: :student, url: mass_update_assignment_type_weights_path, method: :put do |f|
```
zazaian: Much better :+1:

config/routes.rb
```diff
@@ -112,235 +110,180 @@
...
 
-  #SAML
-  get "saml/init"
-  post "saml/consume"
-  get "saml/metadata"
-  get "saml/logout"
+  resource :saml, only: [] do
+    collection do
+      post :consume
+      get :init
```
zazaian: Maybe this should be `get :initialize`? Seems a bit more expressive.

config/routes.rb
```diff
@@ -112,235 +110,180 @@
...
+      put :predictor_settings, to: "courses#predictor_settings_update"
     end
   end
-  resources :course_memberships
-
+  resources :course_memberships, only: [:create, :destroy]
   get "/current_course/change" => "current_courses#change", as: :change_current_course
-  get "current_course" => "current_courses#show"
-
-  get "leaderboard" => "students#leaderboard"
```
zazaian: Nice clean-up.

config/routes.rb
```diff
@@ -112,235 +110,180 @@
...
-  resources :rubrics
-
-  #11. Rubrics & Grade Schemes
-  resources :grade_scheme_elements do
-    collection do
-      post :destroy_multiple
-      get "mass_edit" => "grade_scheme_elements#mass_edit", as: :mass_edit
-      put "mass_edit" => "grade_scheme_elements#mass_update"
-    end
+  controller :pages do
```
zazaian: :+1:

spec/controllers/api/earned_badges_controller_spec.rb
```diff
@@ -15,5 +15,23 @@
             change {EarnedBadge.count}.by(1)
         end
     end
+
+    describe "DELETE destroy" do
+      it "deletes a badge" do
+        earned_badge = create(:earned_badge, grade: world.grade, student: world.student)
+        params = { id: earned_badge.id }
+        expect { delete :destroy, params }.to \
```
zazaian: This is one of those instances in which it feels a little over the line to use two completely separate expectations in the same example block. Either of these could fail and would in turn obfuscate the failure or success of the other. I think when we're checking for common sets of attributes, like:

```
foo.update_attributes stuff: "great", other_stuff: "not as great"
expect(foo.stuff).to eq "great"
expect(foo.other_stuff).to eq "not as great"
```

...multiple expectations per example is fine. But when it stands to potentially impede the clarity of failure cases it seems cleaner to segregate these into distinct examples:
```
describe "DELETE destroy" do
  context "badge is deleted successfully" do
    let!(:earned_badge) { create(:earned_badge, grade: world.grade, student: world.student) }
    let(:params) { { id: earned_badge.id } }

    it "destroys the badge" do
      expect { delete :destroy, params }.to change { EarnedBadge.count }.by(-1)
    end

    it "renders a successful response" do
      delete :destroy, params
      expect(JSON.parse(response.body)).to \
        eq({"message"=>"Earned badge successfully deleted", "success"=>true})
    end
  end
end

I like the clarity of your solution, but I've already seen a few instances of failures in this situation resulting in confusing outputs due to the multiple expectations per example.

File: spec/controllers/api/grades/earned_badges_controller_spec.rb [hunk 16]
```diff
@@ -0,0 +1,42 @@
...
+                   earned_badges: [{ badge_id: badge_1.id, student_id: world.student },
+                                   { badge_id: badge_2.id, student_id: world.student },
+                                   { badge_id: badge_3.id, student_id: world.student }]}
+        expect { post :create, params }.to change { EarnedBadge.count }.by(3)
+      end
+    end
+
+    describe "DELETE delete_all" do
+      it "deletes all earned badges for a grade" do
+        earned_badge_1 = create(:earned_badge, grade: world.grade)
```
zazaian: Same note as above. It feels like we're testing too much for a single example here.

spec/controllers/api/grades/earned_badges_controller_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+describe API::Grades::EarnedBadgesController do
+  let(:world) { World.create.with(:course, :student, :grade, :badge) }
+  let(:professor) { create(:professor_course_membership, course: world.course).user }
+
+  context "as professor" do
+    before(:each) { login_user(professor) }
+
+    describe "POST create" do
+      it "creates new earned badges from params" do
+        badge_1 = create(:badge)
```
zazaian: :+1:

zazaian:
@jwright I think this looks like a vast improvement over what we had in place originally. I just had some comments on consistency with how we're using the `/api` namespace, and some other bigger-picture considerations here that I was curious about. Overall this is a really nice condensation of our routes, though, and most of this reads quite a bit more cleanly.

app/assets/javascripts/angular/controllers/GradeCtrl.js.coffee
jwright: @zazaian Yes. @acousticrobot has been leading the charge here. Any opportunity to consolidate that in this PR, I added. He is also making changes to follow the [JSONAPI](http://jsonapi.org/) spec as well.

@chcholman That should not matter.

chcholman: @jwright Thanks!

app/assets/javascripts/angular/services/GradeSchemeElementsService.js.coffee
jwright: @zazaian A lot of the json API stuff needs to be cleaned up and have a major refactor. In this case, the route was duplicated and both of these routes went to the same endpoint and so this was simply a consolidation of a route that already existed.

app/assets/javascripts/angular/services/PredictorService.js.coffee
jwright: This should not of been a `POST` since it updates a resource and does not create one.

app/assets/javascripts/timeline_settings.js
jwright: The whole `API` namespace was not updated as part of this PR, just a cleanup of existing routes. Only a few routes were changed to `/api` since they were either created or just needed to add a resource part.

app/controllers/api/grades/earned_badges_controller.rb
jwright: This was moved from the `GradesController` as part of that cleanup. This was taken as is and the additional fixing of the status, etc. was not cleaned up. I agree that 400 is definitely not the correct response here.

app/controllers/grades_controller.rb
jwright: 1997 called...

app/controllers/level_badges_controller.rb
jwright: There are not really a lot of cases to use that in controllers if that is the resource you are working with. This will return a 404 which should occur if the `params[:id]` is not correct.

config/routes.rb
jwright: I was unaware of the `saml` stuff so I did not change it. The goal of this PR was to clean up the routes without blowing out into fixing all the controller, action issues, etc.

spec/controllers/api/earned_badges_controller_spec.rb
jwright: This was moved from another test file without cleanup to not expand the refactoring of this PR. 

I can clean up these specs. I agree with you. Since these are controller specs however, which are expensive in terms of runtime, I struggle with expanding the number of these specs. 

I like the split in this situation. Let me know if you'd like to see this changed in this PR.

jwright:
@zazaian Thanks!

I tried to limit the scope of changes in this PR. I basically touched every controller and action for these routes and I tried not to clean up the `/api`, etc except where it meant removing functionality from an already large controller or if it made the endpoints more restful.

I am just looking for feedback on how much further refactoring we should do on this PR or wait until other endpoint refactoring by developers who are focused and knowledgable on the functionality of the endpoints.

app/assets/javascripts/angular/controllers/GradeCtrl.js.coffee
zazaian: @jwright okay cool. I wasn't sure to what extent we were seeking to implement those api namespacing changes in this PR. I just saw that adjustment and wondered whether it was a matter of inconsistency. If @acousticrobot is going to continue to address this in other PR's then let's not worry about it for now.

app/assets/javascripts/angular/services/GradeSchemeElementsService.js.coffee
zazaian: @jwright cool :+1:

app/assets/javascripts/angular/services/PredictorService.js.coffee
zazaian: Sorry if I was unclear here. I'm not commenting on the correctness of the POST call here, but rather recalling that there was at least one instance in which we used POST as a hack to supplant the PUT call. I'm suggesting simply that in the future if we're going to hack around PUT deficiencies with POST we should have a consistent commenting format. Your implementation here looks clean, though.

app/assets/javascripts/timeline_settings.js
zazaian: :+1: cool

app/controllers/api/grades/earned_badges_controller.rb
zazaian: Okay cool, I'll create a separate issue for this.

zazaian: @jwright FYI I've added issue #2087 to handle this case.

app/controllers/grades_controller.rb
zazaian: Did you take a message for me?

app/controllers/level_badges_controller.rb
zazaian: So 404 should probably only be used in the case of a nonexistent URI rather than simply params being passed for an nonexistent resource. This should probably be returning 400 rather than 404.

config/routes.rb
zazaian: Okay cool -- changing the route name seemed superficial to me so I thought it might fall into the scope of this PR. This and the other SAML-wrangling concerns should be dealt with in a separate issue.

spec/controllers/api/earned_badges_controller_spec.rb
zazaian: Okay, I guess I'm fine with this either way. I understand the desire to preserve speed here even if that costs some example separation. Maybe we should bring this up in #TuesdayChats?

zazaian:
Hey @jwright, back to you here. Maybe only one or two items that need any kind of action here. Let me know if you have any other questions.

app/assets/javascripts/angular/services/PredictorService.js.coffee
jwright: 👍 Gotcha my dude.

app/controllers/api/grades/earned_badges_controller.rb
jwright: You rock 👌

app/controllers/grades_controller.rb
jwright: They just wanted to know if we needed more AOL hours.

app/controllers/level_badges_controller.rb
jwright: This is the base behavior in all Rails apps (http://api.rubyonrails.org/v3.2.6/classes/ActiveResource/Base.html#label-Resource+errors).

It rescues an `ActiveRecord::ResourceNotFound` with a 404. I don't think we should change it. Most restful implementations also suggest this as the convention since 404 is a bad uri as well.

spec/controllers/api/earned_badges_controller_spec.rb
jwright: I agree. I have added it to 🎼 Tuesday Chats 🎼

PR #2083, opened 2016-06-10
Remove timeline settings from course form

chcholman:
This PR removes two form items from the Course setup process: whether or not to use the timeline, and whether or not to add the link for additional details to the event description.


jwright:
@chcholman This looks great!

PR #2086, opened 2016-06-10
Added rake task to update AssignmentType max_points to 0

chcholman:
In preparation for merging the migration to add 'null: false' requirements to AssignmentType fields, we need to deal with the 231 records that current have nil values here. 

I have confirmed that all of the other columns we're adding limitations to do not have any records with nil values.


jwright:
👍

Issue #1541, opened 2015-12-25
Filter To Do List


chcholman:
As a student
I should be able to filter the to do list to only show me assignments I have predicted

Issue #1880, opened 2016-04-06
Smaller version of grading scheme somewhere more visible


mariehooper:
This may be resolved through the implementation of dashboard

PR #2088, opened 2016-06-12
2057 student dashboard

mariehooper:
This PR implements a modular dashboard design to replace the timeline on the current student homepage. Modules are mostly comprised of information from the former student sidebar. Included in this PR is the ability to filter todo list down to "My Planner" and a small version of the grading scheme showing just one level below, current level and one level above.

### Before
screen shot 2016-06-12 at 6 59 04 am

### After
after-dashboard

### What's next
Upcoming Events will be populated with events from the current timeline and timeline.js will be removed entirely. "Since last login" module data will be worked on with issue #2027. More improvements to view on tablet-size devices will be made. Also, all renaming of files will occur to replace instances of  "student-sidebar" with "student-dashboard" or similar and highcharts data visuals will be replaced with D3.js.

Closes issues #1541 #2057 and #1880


chcholman:
@mariehooper A few tiny style changes!

1) Will you switch the ordering of the Course Planner and My Planner sections
2) Darken the 'off' state for the section titles - I'm worried about the density for accessibility considerations
3) Retitle the tab to be "Dashboard" rather than "Timeline"

What do you think about the blue color + lightweight font style for each to do list item? I'm worried it's slightly too light for easy readability.

mariehooper:
Thanks for all the feedback and help @chcholman !!

PR #2090, opened 2016-06-12
Fix issues with student import

jwright:
I could not replicate the issue of a missing reference. I was able to do it once but that was because I modified the server on development and I did not restart it and Rails does not autoload the classes in `/app/importers`. Once I restarted the server, everything was fine. In addition, this line also exists: https://github.com/UM-USElab/gradecraft-development/blob/master/app/importers/student_importer.rb#L2 and so it should be ok.

@chcholman Did the error you reported happen on production?

I did notice an issue with the importer however in that it always tries and creates a student relationship. This commit checks to see if the course membership exists before trying to create it.

### DIFF
--- a/app/importers/student_importer.rb
+++ b/app/importers/student_importer.rb
@@ -103,8 +103,7 @@ class StudentImporter
         last_name: last_name,
         email: email,
         username: username,
-        password: password,
-        team_name: team_name
+        password: password
       }
     end
   end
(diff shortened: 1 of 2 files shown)


chcholman:
@jwright Thanks for this! I encountered the error locally when testing the feature before trying to write documentation for a new user. Weirdly I did restart the server locally to see if that would resolve it, and encountered the same issue. 

Thanks for this update, I'll see if I can find anything else and let you know!

PR #2094, opened 2016-06-13
Analytics Export Hotfix

zazaian:
This PR resolves the issue that was occurring with assignment_id being called on events that had no assignment_id method. I've added a spec for `CoursePredictorExport#assignment_name`, which is where this was coming from, to test all of the cases.

Additionally I've added a `mongoid_spec_helper` class in `/spec` to help with expressly loading the Mongoid config (it wasn't being recognized in some examples), and have tested this patch against the production dump of data for course #41 which is where this error occurred in the first place.

@acousticrobot or @chcholman please take a look at this when you get a chance.

### DIFF
--- a/lib/analytics.rb
+++ b/lib/analytics.rb
@@ -1,9 +1,11 @@
+require "is_configurable"
+require_relative "analytics/configuration"
+require_relative "analytics/event"
+require_relative "analytics/aggregate"
+require_relative "analytics/data"
+require_relative "analytics/export"
+require_relative "analytics/login_event"
+
 module Analytics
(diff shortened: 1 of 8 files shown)


lib/analytics.rb
```diff
@@ -1,9 +1,11 @@
+require "is_configurable"
+require_relative "analytics/configuration"
+require_relative "analytics/event"
+require_relative "analytics/aggregate"
+require_relative "analytics/data"
+require_relative "analytics/export"
+require_relative "analytics/login_event"
+
```
acousticrobot: 👍

PR #2098, opened 2016-06-14
Refactor/points value score

acousticrobot:
This pull request standardizes the naming conventions for all columns related to "points", "score", and "value". 

The details of the new naming conventions have been outlined in the [WIKI](https://github.com/UM-USElab/gradecraft-development/wiki/Points) as part of this refactor.

Most changes revolve around migrations to manage these changes, and simple replacements of the attribute names throughout the code.

One exception is that the `Grade.final_points` has been updated from [a temporary method](https://github.com/UM-USElab/gradecraft-development/pull/2098/files#diff-30a5010b9926513d0a5481926fd1a065L90) to a replacement for `final_score` as outlined in the method's TODO comments from when these changes were first planned.

This PR should NOT be merged until we have verified that the rake task `update_final_score` ([PR#2061](https://github.com/UM-USElab/gradecraft-development/pull/2061)) has run successfully.


zazaian:
@acousticrobot FYI I've seen this PR and I'll take a look tonight or tomorrow morning :+1:

app/assets/javascripts/angular/controllers/PredictorCtrl.js.coffee
```diff
@@ -99,7 +99,7 @@
   $scope.closestScoreLevel = (scoreLevels,value)->
```
zazaian: It looks like the `value` argument here in all instances is actually taking other attributes and methods called `points`. Should this be called `points` as well, or is there a more expressive name that we could use for this like `predicted_points`?

app/assets/javascripts/angular/controllers/PredictorCtrl.js.coffee
```diff
@@ -89,8 +89,8 @@
 
   # Assignments with Score Levels: Defines a snap tolerance and returns true if value is within range
   $scope.inSnapRange = (assignment,scoreLevel,value)->
```
zazaian: Same comment as below, this argument is called `value` but seems like its utility may be more clear if it had a name like `predicted_points`.

app/assets/javascripts/angular/controllers/PredictorCtrl.js.coffee
```diff
@@ -269,18 +269,18 @@
           if $scope.inSnapRange(article,closest,ui.value)
             event.preventDefault()
             event.stopPropagation()
-            angular.element(ui.handle.parentElement).slider("value", closest.value)
+            angular.element(ui.handle.parentElement).slider("value", closest.points)
 
             if articleType == 'assignment'
-              angular.element("#assignment-" + article.id + "-level .value").text($filter('number')(closest.value) + " / " + $filter('number')(article.point_total))
+              angular.element("#assignment-" + article.id + "-level .value").text($filter('number')(closest.points) + " / " + $filter('number')(article.full_points))
```
zazaian: I understand that this isn't the thrust of this PR, but maybe can we clean up some of the duplication here? The only difference between line 275 and line 277 is the id prefix `#assignment-` vs `#challenge-`. Maybe for starters we could try something like this?

```
article_prefix = articleType == 'assignment' ? "assignment" : "challenge"
angular
  .element("#" + article_prefix + "-" + article.id + "-level .value")
  .text($filter('number')(closest.points) + " / " + $filter('number')(article.full_points))
```

app/assets/javascripts/angular/controllers/PredictorCtrl.js.coffee
```diff
@@ -269,18 +269,18 @@
...
             else
-              angular.element("#challenge-" + article.id + "-level .value").text($filter('number')(closest.value) + " / " + $filter('number')(article.point_total))
+              angular.element("#challenge-" + article.id + "-level .value").text($filter('number')(closest.points) + " / " + $filter('number')(article.full_points))
           else
 
             if articleType == 'assignment'
```
zazaian: Same comment here. The only difference between this code and the lines above is the usage of ui.value instead of closest.points. Perhaps this can be extracted into a method to reduce duplication?

app/assets/javascripts/angular/controllers/GradeRubricCtrl.js.coffee
```diff
@@ -190,7 +190,7 @@
           criterion_id: level.criterion_id,
           badge_id:     badge.badge.id,
           description:  badge.description,
-          point_total:  badge.point_total,
```
zazaian: Cool, this is a lot more readable :+1:

app/assets/javascripts/angular/controllers/PredictorCtrl.js.coffee
```diff
@@ -343,27 +343,27 @@
     axis = d3.svg.axis().scale(scale).orient("bottom")
     g = svg.selectAll('g').data(gradeSchemeElements).enter().append('g')
             .attr("transform", (gse)->
-              "translate(" + (scale(gse.low_range) + padding) + "," + 25 + " )")
+              "translate(" + (scale(gse.low_points) + padding) + "," + 25 + " )")
             .on("mouseover", (gse)->
-              d3.select(".grade_scheme-label-" + gse.low_range).style("visibility", "visible")
-              d3.select(".grade_scheme-pointer-" + gse.low_range)
+              d3.select(".grade_scheme-label-" + gse.low_points).style("visibility", "visible")
```
zazaian: Same comment as above -- lines 348-349 are identical to lines 354-355 except for the "visible" and "hidden" strings. Seems like there might be a way to reuse some of this code in a method to avoid duplication. Perhaps we could create a separate issue for this effort?

app/assets/javascripts/angular/directives/GradeSchemeDirectives.js.coffee
```diff
@@ -5,14 +5,14 @@
     element: '=ngModel'
   }
   controller: ($scope) ->
-    this.low_range = (modelValue, viewValue) ->
-      if (modelValue < $scope.element.high_range || $scope.element.high_range == '')
+    this.low_points = (modelValue, viewValue) ->
```
zazaian: :+1: cool cool, man this must've been tedious. :turtle:

app/assets/javascripts/angular/services/GradeSchemeElementsService.js.coffee
```diff
@@ -10,24 +10,24 @@
...
-        high_range: elements[index].low_range-1
+        low_points: ''
+        high_points: elements[index].low_points-1
       })
 
     addFirst = () ->
       elements.push({
         letter: ''
         level: ''
-        low_range: ''
```
zazaian: Was `range` being used in lieu of `score` here?

app/assets/javascripts/angular/templates/ng_predictor_challenges.html.haml
```diff
@@ -27,18 +27,18 @@
 
         .article-graded{'ng-if' => 'articleGraded(challenge)'}
           .grade
-            {{challenge.grade.score | number}} / {{challenge.point_total | number}}
+            {{challenge.grade.score | number}} / {{challenge.full_points | number}}
```
zazaian: :+1:

app/controllers/api/grade_scheme_elements_controller.rb
```diff
@@ -5,13 +5,13 @@ class API::GradeSchemeElementsController < ApplicationController
   def index
     @grade_scheme_elements = current_course
                              .grade_scheme_elements
-                             .order_by_high_range.select(
+                             .order_by_high_points.select(
```
zazaian: Would `highest_points` make more sense than `high_points` here?

app/controllers/api/grade_scheme_elements_controller.rb
```diff
@@ -5,13 +5,13 @@ class API::GradeSchemeElementsController < ApplicationController
...
                                :id,
-                               :low_range,
+                               :low_points,
                                :letter,
                                :level)
     if @grade_scheme_elements.present?
-      @total_points = (@grade_scheme_elements.first.low_range).to_i
+      @total_points = (@grade_scheme_elements.first.low_points).to_i
```
zazaian: Same as above, maybe `lowest_points` instead of `low_points`?

app/controllers/grade_scheme_elements_controller.rb
```diff
@@ -6,7 +6,7 @@ class GradeSchemeElementsController < ApplicationController
   def index
     @title = "Grade Scheme"
     @grade_scheme_elements = current_course
-                             .grade_scheme_elements.order_by_high_range
+                             .grade_scheme_elements.order_by_high_points
```
zazaian: same as above, maybe `highest` and `lowest` would be more expressive than `high` and `low` in this context.

app/helpers/users_helper.rb
```diff
@@ -17,15 +17,15 @@ def total_scores_for_chart(user, course)
                   name: assignment_type.name }
     end
 
-    # TODO: Earned badges pre 2016 have nil scores cached on the model, must convert to integer
-    earned_badge_score = user.earned_badges.where(course: course).pluck("score").sum {|s| s.to_i}
-    if earned_badge_score > 0
-      scores << { data: [earned_badge_score], name: "#{course.badge_term.pluralize}" }
+    # TODO: Earned badges pre 2016 have nil points cached on the model, must convert to integer
```
zazaian: This is unrelated, but the logic here seems too involved and procedural for a single method. We may want to look into extracting the logic in this method into a `UserChart` class or something similar and performing this logic there rather than in the helper method. Extraction here would make this logic much easier to test and might help to improve the clarity of the overall workflow. I can create an issue for this if desired.

app/models/assignment.rb
```diff
@@ -118,9 +118,9 @@ def has_groups?
   end
 
   # Custom point total if the class has weighted assignments
-  def point_total_for_student(student)
-    return 0 unless point_total
-    (point_total * assignment_type.weight_for_student(student)).round
+  def full_points_for_student(student)
```
zazaian: Funsies, this must have been a lot of damn grepping to catch all of this stuff :+1:

app/models/grade.rb
```diff
@@ -79,32 +79,12 @@ def feedback_reviewed!
     update_attributes feedback_reviewed: true, feedback_reviewed_at: DateTime.now
   end
 
-  # Handle raw score attributes with commas (ex "300,000")
-  def raw_score=(rs)
-    if rs.class == String
-      rs.delete!(",").to_i
+  # Handle raw points attributes with commas (ex "300,000")
+  def raw_points=(rp)
```
zazaian: Any thoughts on using `rp` as a method name here? Maybe something like `points_value` might be more expressive?

app/models/grade.rb
```diff
@@ -161,33 +141,33 @@ def excluded_by
...
+  def calculate_full_points
+    assignment.full_points_for_student(student)
   end
 
   # totaled points (adds adjustment, without weighting)
-  def calculate_final_score
-    return nil unless raw_score.present?
-    final_score = raw_score + adjustment_points
-    final_score > assignment.threshold_points ? final_score : 0
+  def calculate_final_points
```
zazaian: points, points, points, points, points, points, points.... :+1:

app/views/assignments/grades/mass_edit.html.haml
```diff
@@ -35,22 +35,22 @@
                 = label_tag do
                   // Checking to see if the grade is actually graded before setting the default value to 0
                   - if @assignment.grade_select? && @assignment.has_levels?
-                    = gf.select :raw_score, options_from_collection_for_select(@assignment_score_levels, :value, :formatted_name, grade.try(:raw_score)), include_blank: true
```
zazaian: fun times... :+1:

db/migrate/20160602140234_change_points_nomenclature.rb
```diff
@@ -0,0 +1,17 @@
+class ChangePointsNomenclature < ActiveRecord::Migration
```
zazaian: Holy points migration, Batman! :boom:

lib/human_history/raw_points_change_description_formatter.rb
```diff
@@ -1,9 +1,9 @@
 require_relative "default_change_description_formatter"
 
 module HumanHistory
-  class RawScoreChangeDescriptionFormatter < DefaultChangeDescriptionFormatter
+  class RawPointsChangeDescriptionFormatter < DefaultChangeDescriptionFormatter
```
zazaian: Naming a class after the only method that it calls feels like an antipattern to me. It seems like this should probably be a `RawPointsFormatter`. We should probably create a separate issue for this.

spec/controllers/assignment_type_weights_controller_spec.rb
```diff
@@ -58,9 +58,9 @@
       end
 
       it "updates points for corresponding grades" do
-        grade = create :released_grade, assignment: @assignment_type_weightable.assignments.first, student: @student, course: @course, raw_score: 1000
-        grade_2 = create :released_grade, assignment: @assignment_type_weightable_2.assignments.first, student: @student, course: @course, raw_score: 1000
-        grade_3 = create :released_grade, assignment: @assignment_type_weightable_3.assignments.first, student: @student, course: @course, raw_score: 1000
+        grade = create :released_grade, assignment: @assignment_type_weightable.assignments.first, student: @student, course: @course, raw_points: 1000
```
zazaian: Seems like there's a lot of redundant logic here. Maybe we can move some of this into a helper method or a dedicated factory namespace?

spec/factories/grade_scheme_element_factory.rb
```diff
@@ -1,19 +1,19 @@
 FactoryGirl.define do
   factory :grade_scheme_element do
     association :course
-    low_range 0
-    high_range 100000
+    low_points 0
```
zazaian: Maybe `highest` and `lowest` points again?

spec/models/assignment_spec.rb
```diff
@@ -106,8 +106,8 @@
     before { subject.save }
 
     it "returns the average score for a graded grade" do
-      subject.grades.create student_id: create(:user).id, raw_score: 8, score: 8, status: "Graded"
-      subject.grades.create student_id: create(:user).id, raw_score: 5, score: 8, status: "Graded"
+      subject.grades.create student_id: create(:user).id, raw_points: 8, score: 8, status: "Graded"
```
zazaian: Can we stub or double out some of these users to avoid having to create a unique user for each grade? It seems like we only need the id.

app/assets/javascripts/angular/controllers/PredictorCtrl.js.coffee
acousticrobot: I think this is a good idea, but yeah for another PR. Right now @mariehooper is writing some Angular for Badges that has overlapping fx with the Predictor. So this is a good time to start thinking about all of the Angular controllers, and how we might want to make them more modular and reusable.

app/assets/javascripts/angular/controllers/PredictorCtrl.js.coffee
acousticrobot: I decided to leave all the generic `value` arguments alone for now, since they aren't tied to the db migrations. They probably could be more expressive, but at least for now they are all the same across the methods.

app/assets/javascripts/angular/directives/GradeSchemeDirectives.js.coffee
acousticrobot: Needles in a haystack 🔍

zazaian:
@acousticrobot I took a look at this and overall it looks like you did a really good job of catching all of the instances of 'score' and 'range' being renamed to points, which I'm sure wasn't a small task as there was clearly an enormous amount of score-related logic littered around the app. I did have a couple of notes about javascript method arguments named `value` instead of `points` or `predicted_points`, and we should probably make a decision about how we're sandboxing those concerns within the context of a method, i.e. are we more concerned about where it came from to get to the method, or how it's being used now that it's been passed into the method as an argument?

I made several comments on code that wasn't related to the actual changes that you made, but rather that I reviewed as a result of your changes being made within the context of various methods. If you'd like I can submit separate issues for all of these and we can keep this PR just about the renaming of [score|range] to points. Let me know how you'd like to deal with these and we can go from there. Maybe @chcholman can offer some input on how she'd like to see these handled.

Also it looks like there's a `# TODO:` comment in the users_helper that rubocop is complaining about. Extracting this into a github issue might easily resolve that hiccup. Additionally there was an S3Manager related spec hiccup on the run of this branch which obviously wasn't related to any of the work that you've done here so I've re-run the build to see if it'll get a clean bill of health.

app/assets/javascripts/angular/services/GradeSchemeElementsService.js.coffee
acousticrobot: I think the term range was used because GSEs have a range defined by a lowest and highest point value. I changed them to `points` because by themselves these values are single `point` values, and the `range` is the area of difference between the two points.

app/controllers/api/grade_scheme_elements_controller.rb
acousticrobot: Yeah I think your right, good call! I'll change these two before submitting the PR

app/helpers/users_helper.rb
acousticrobot: I know we already ran into a problem with this failing back in February, which is where the `{|s| s.to_i}` and the comment came from. My guess is this entire file was very old logic pulled off of the User in a quick attempt to make this model a little less of a nightmare but I'm not sure about that. Feel free to create an issue, or add it into Tuesday Chats.

app/models/grade.rb
acousticrobot: I think `points_value` gets back into the weeds of when is it a `point`, when is it a `value`, and when is it a `score`. I'm inclined to leave it as it was when I found it for now, since it's a four line method that handles internal cleanup.

spec/controllers/assignment_type_weights_controller_spec.rb
acousticrobot: Probably. I have another PR in the works that cleans up the way weights are handled. I'll take a look at the specs as well. There's a good chance a lot of this will be deleted entirely.

spec/models/assignment_spec.rb
acousticrobot: Looks like `Assignment.average` comes from `/app/models/concerns/gradable.rb`. So it probably should be extracted into a spec in a `spec/models/concerns` folder if we're going to test it with doubles.

chcholman:
@acousticrobot In regards to the comment, we need to write a rake task to update the 1422 earned badges that have a score of nil. It's on my db cleanup list, and you can remove it here as a comment.

zazaian:
@acousticrobot sounds good, I'm glad that the feedback here was helpful. I'm going to go through the items that I've stumbled onto here and add separate issues for them so we can keep track of everything that we discussed. Feel free to bounce this back to me whenever you've made those additional changes but I think with those minor `high`->`highest` etc. nomenclature adjustments we should be good to go.

app/assets/javascripts/angular/services/GradeSchemeElementsService.js.coffee
zazaian: Cool, that makes sense. This was the first time that came up within the context of this PR so I thought I'd ask.

app/models/grade.rb
zazaian: @acousticrobot I suggest value here because the default behavior for `attr_accessor` and `attr_writer` in ruby is to use the argument `new_value`. Conventionally it seems that abbreviated argument names are only really used for block arguments. I think this is fine as the utility is clear, but for a setter `new_value` is conventional.

spec/controllers/assignment_type_weights_controller_spec.rb
zazaian: @acousticrobot cool cool, sounds good.

spec/models/assignment_spec.rb
zazaian: @acousticrobot point taken. I actually had trouble finding these methods but I should have just grepped `models/concerns`. We should probably be testing the concern modules in dedicated specs included in a test-only class. Once these behaviors are tested in those dedicated module specs we can then simply test the inclusion of the module in the `Assignment` module etc. and we should be good to go. I'll create an issue for this.

PR #2100, opened 2016-06-16
Course data

chcholman:
This PR makes it possible to download a variety of different pieces of data about many courses from a single location (the Course index). It also adds a descriptive naming pattern to most file downloads for easier storage/tracking.

### DIFF
--- a/app/controllers/assignments/grades_controller.rb
+++ b/app/controllers/assignments/grades_controller.rb
@@ -10,11 +10,7 @@ class Assignments::GradesController < ApplicationController
   def download
     assignment = current_course.assignments.find(params[:assignment_id])
     respond_to do |format|
-      format.csv {
-        send_data GradeExporter.new
-          .export_grades(assignment, current_course.students),
-        filename: "#{ assignment.name } Import Grades - #{ Date.today }.csv"
-      }
+      format.csv { send_data GradeExporter.new.export_grades(assignment, current_course.students), filename: "#{ assignment.name } Import Grades - #{ Date.today}.csv" }
     end
   end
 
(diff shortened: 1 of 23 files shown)


app/controllers/assignments/grades_controller.rb
```diff
@@ -53,21 +54,21 @@ def update_status
...
     assignment = current_course.assignments.find(params[:assignment_id])
     respond_to do |format|
-      format.csv do
-        send_data CriterionGradesExporter.new.export assignment.course,
-          assignment.rubric
-      end
+      format.csv {
+        send_data CriterionGradesExporter.new.export(assignment.course, assignment.rubric),
+        filename: "#{ assignment.name } Rubric Grades - #{ Date.today }.csv"
+      }
```
acousticrobot: I see we turned `Style/BlockDelimiters` off in Rubocop, which would have complained about this.  I don't have strong feelings about it, but I always thought this was a no-no in Ruby

```
Style/BlockDelimiters:
  Description: >-
                Avoid using {...} for multi-line blocks (multiline chaining is
                always ugly).
                Prefer {...} over do...end for single-line blocks.
  StyleGuide: 'https://github.com/bbatsov/ruby-style-guide#single-line-blocks'
  Enabled: false
```

app/controllers/assignments_controller.rb
```diff
@@ -109,8 +109,12 @@ def destroy
   end
 
   def export_structure
+    course = current_user.courses.find_by(id: params[:id])
     respond_to do |format|
-      format.csv { send_data AssignmentExporter.new.export current_course }
+      format.csv {
+        send_data AssignmentExporter.new.export(course),
+      filename: "#{ course.name } #{ (term_for :assignment).titleize } Structure - #{ Date.today }.csv"
```
acousticrobot: same as above, plus I think `filename...` should be indented 2 more spaces

PR #2108, opened 2016-06-17
Removed unused fields on assignments

chcholman:
This PR cleans up the fields on the assignment model, by dropping 5 unused fields, and adding 4 rake tasks to clean up fields that should have a default of 0. I will follow up this PR with another one adding `null: false` to 26 fields that will then all have present data.


app/models/assignment.rb
```diff
@@ -15,7 +15,7 @@ class Assignment < ActiveRecord::Base
     :course, :course_id, :description, :due_at, :grade_scope, :hide_analytics,
     :include_in_predictor, :include_in_timeline, :include_in_to_do,
     :mass_grade_type, :name, :open_at, :pass_fail,
-    :full_points, :points_predictor_display, :purpose, :release_necessary,
```
acousticrobot: Yay! 🎆

PR #2113, opened 2016-06-17
1857 fix student paths

jwright:
Fixes student routes and controllers that were causing the facility view of a student to no longer be displayed.

Fix for #1857

### DIFF
--- a/app/controllers/students_controller.rb
+++ b/app/controllers/students_controller.rb
@@ -68,8 +68,9 @@ class StudentsController < ApplicationController
   end
 
   def teams
+    student = params[:id].present? ? User.find(params[:id]) : current_student
     @title = "#{term_for :teams}"
-    @team = current_student.team_for_course(current_course)
+    @team = student.team_for_course(current_course)
     @teams = current_course.teams.order_by_rank.includes(:earned_badges)
   end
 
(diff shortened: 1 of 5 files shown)


chcholman:
@jwright I fixed the haml issue and did the rebase - will you take a look and see if you think it's okay?

jwright:
@acousticrobot @chcholman I am still seeing the broken dashboard for faculty. I am looking into it.

jwright:
@chcholman @acousticrobot It also looks like the `student_id` parameter is back for the grading scheme. I will leave that it since it looks intentional.

chcholman:
@jwright If there's a way to do it such that it doesn't need the `student_id` param, but does allow the instructor to see it as having each student's data, that should come out. I think I've mis-articulated the requirements here - on the student side, these views don't need a param because we have a current_student id. On the instructor side, they need to be able to see views as customized for each student, thus the param.

jwright:
@chcholman Most of the routes were changed to use a more restful syntax and removed the `student_id` param. I will leave it for now since this needs to be merged ASAP.

jwright:
@chcholman @acousticrobot I am not sure this is a bug on the dashboard. It looks like the admin dashboard just does not display anything. Is that correct?

jwright:
I added the student profile tabs to the predictor. The dashboard does not display anything but I think that is a result of this commit: https://github.com/UM-USElab/gradecraft-development/commit/9b865f36ebd2426dfa03dbbe10e2eb6e0870531e

Let me know if there is any other issues.

chcholman:
@jwright Yup, the staff dashboard currently has no content (but renders) and that is correct. Thank you!

PR #2121, opened 2016-06-20
replace weights form with interactive weights page

acousticrobot:
This PR:
  * removes the `AssignmentTypeWeightStruct` and `AssignmentTypeWeightForm`
  * replaces the assignment type weight form with an interactive weights page using resourceful routing
  * moved the Angular Weights off of the Predictor and into its own Controller, Directives, and Service

Left for the next Weight PR:
  * We need to have the UX team style the new weights page, right now if simply repeats all the elements from the predictor
  * This restyling should include a refactor of the predictor sass file, right now these class names are not very usable in the context of the weights page.
  * I pulled methods off of the Angular PredictorController, but I think they could be cleaned up a lot more. We need a strategy for determining where methods and model management should live in the Angular.
  * We are hoping to remove the `default_weight` entirely from the App. Once this is done, there is a lot of cleanup that can happen to the `Weights` logic.  I would also like to add a validator method to AssignmentType that runs before a weight is updated at this point.


app/assets/javascripts/angular/controllers/PredictorCtrl.js.coffee
```diff
@@ -126,36 +126,20 @@
...
-    assignments = $scope.assignmentsForAssignmentType($scope.assignments,assignmentType.id)
+
     if includePredicted
+      assignments = $scope.assignmentsForAssignmentType($scope.assignments,assignmentType.id)
       total = $scope.assignmentsPointTotal(assignments)
     else
-      total = $scope.assignmentsPointTotal(assignments, false)
+      total = AssignmentTypeService.weightedEarnedPoints(assignmentType)
 
     if includeWeights
```
jwright: Since we are modifying `total` here, I think we should add these changes to `total` as part of the else statement above.

app/assets/javascripts/angular/helpers/serviceHelpers.js.coffee
```diff
@@ -0,0 +1,37 @@
...
+        pass: "Pass"
+        fail: "Fail"
+        badges: "Badges"
+        challenges: "Challenges"
+        weights: "Weights"
+    }
+
+    # used to destinguish student and faculty routes.
+    # On init, students views do not send student id.
+    uri_prefix = (studentId)->
```
jwright: I am not sure this is the case for all routes, except for predictor. I am ok with leaving this here but if it's not the case, we may want to keep this inside predictor.

app/assets/javascripts/angular/services/AssignmentTypeService.js.coffee
```diff
@@ -0,0 +1,91 @@
+@gradecraft.factory 'AssignmentTypeService', ['$http', 'GradeCraftAPI', ($http, GradeCraftAPI) ->
```
jwright: Love the commenting here and throughout the PR.

app/assets/javascripts/angular/helpers/serviceHelpers.js.coffee
```diff
@@ -0,0 +1,37 @@
+# Common Methods and Objects for Services using the GradeCraft api routes
+angular.module('helpers')
+  .factory('GradeCraftAPI', ()->
```
jwright: Could this just be `API` rather than `GradeCraftAPI` since we should be in a `GradeCraft` namespace already?

app/controllers/assignment_type_weights_controller.rb
```diff
@@ -1,46 +1,10 @@
...
-
-  private
-
-  def student_params
-    params.require(:student).permit(assignment_type_weights_attributes:
-      [:assignment_type_id, :weight])
+  # GET /assignment_type_weights
+  def index
+    @title =
+      "Editing My #{term_for :weight} Choices" if current_user_is_student?
```
jwright: We don't need the `current_user_is_student?` check here since the `before_filter` manages that.

app/controllers/students/assignment_type_weights_controller.rb
```diff
@@ -0,0 +1,12 @@
+class Students::AssignmentTypeWeightsController < ApplicationController
+
+  before_filter :ensure_staff?
+
+  # GET /students/:id/assignment_type_weights
+  # faculty view of student's weights index view
+  def index
+    @title =
+      "Editing #{current_student.name}'s #{term_for :weight} Choices" if current_user_is_staff?
```
jwright: We don't need the `current_user_is_staff?` check here since the `before_filter` manages that.

app/models/assignment_type.rb
```diff
@@ -92,6 +92,10 @@ def raw_points_for_student(student)
     student.grades.student_visible.where(assignment_type: self).pluck("raw_points").compact.sum || 0
   end
 
+  def final_points_for_student(student)
+    student.grades.student_visible.where(assignment_type: self).pluck("raw_points").compact.sum || 0
```
jwright: Do we need to worry about `nil` `raw_points` here? If so, the `compact` call will fail.

jwright:
@acousticrobot Great PR! This really cleans up a lot.

I made a few comments but nothing really big. My only big concern here is more js without testing. At some point, we should introduce it.

app/controllers/assignment_type_weights_controller.rb
acousticrobot: good catch!

app/controllers/students/assignment_type_weights_controller.rb
acousticrobot: 👍

app/assets/javascripts/angular/helpers/serviceHelpers.js.coffee
acousticrobot: after discussion, we decided to keep this so that it remains explicit when added to the controllers

app/assets/javascripts/angular/helpers/serviceHelpers.js.coffee
acousticrobot: Let's keep an eye on this one, we may need to handle the rubric routes in the future, where the student isn't the top level resource.

app/models/assignment_type.rb
acousticrobot: Just confirmed in irb: `[1,2,nil,4,5].compact => [1, 2, 4, 5]`

PR #2122, opened 2016-06-21
2116 fix grading scheme module when no 0 level ch

chcholman:


app/presenters/students/dashboard_presenter.rb
```diff
@@ -0,0 +1,41 @@
+require "./lib/showtime"
+
+class Students::DashboardPresenter < Showtime::Presenter
```
jwright: Since this is specific to the `dashboard_grading_scheme` partial, I think this should be named to something like `Students::DashboardGradingSchemePresenter` or something similar. I am not sure where we ended up on the presenter naming standard either. We may want to name it according to that.

app/views/students/dashboard/_dashboard_grading_scheme.haml
```diff
@@ -2,32 +2,39 @@
   %h3 Course Progress
 .card-body
 
-  - next_element = current_student.get_element_level(current_course, :next)
-  - if !next_element.nil?
+  - if presenter.current_element.present?
+    - if presenter.next_element.present?
```
jwright: The embedding here is a bit confusing. Can we just check `if presenter.next_element.present?` here and move line 5 to line 13?

app/views/students/dashboard/_dashboard_grading_scheme.haml
```diff
@@ -2,32 +2,39 @@
...
+          %span=points presenter.next_element.lowest_points
+
+    .progress.bar_magic.current.center
+      %span.your-rank Your rank:
+      %span.your-rank= presenter.current_element.name
+      .striped-bar-container
+        .striped{ style: "width: #{ presenter.progress_percent }%;" }
+      .left-points
+        %span Your Points:
+        %span.bold=points presenter.score_for_course
```
jwright: The method is called `score_for_course` but the label is `Points`. Since we decided to use `points`, can we rename this to `points_for_course`?

jwright:
@chcholman This is great! I just had a few comments but I think the presenter naming is the most important here. It will eliminate the bloat that this presenter may take on later. I like the focus of this presenter.

mariehooper:
Thanks for the help and comments @chcholman and @jwright. I think this one is ready to go...

PR #2123, opened 2016-06-21
Fixing errant spec fail

chcholman:
This PR fixes a sporadically failing spec that was checking student visible grades, and occasionally receiving them in a different order than the `eq` matcher wanted. By switching it to an `includes` check this should pass consistently.


zazaian:
Looks good! Nice use of the `include` matcher.

zazaian:
@chcholman this looks ready to merge.

PR #2124, opened 2016-06-21
Delete all grades

jonathoy:


app/controllers/assignments/grades_controller.rb
```diff
@@ -137,7 +137,28 @@ def mass_update
...
+  # Delete all grades for a given assignment id
+  def delete_all
+    assignment = Assignment.find(params[:assignment_id])
+    assignment.grades.each do |grade|
+      student = grade.student
+      grade.destroy
+      ScoreRecalculatorJob.new(user_id: student.id, course_id: current_course.id).enqueue
+    end
+
+    if assignment.grades.all? { |grade| grade.destroyed? }
```
jonathoy: I'm curious to know if I can get some input on this. I need to delete all grades for an assignment, but I am not sure what the best practice is for validating that the grades were, in fact, deleted. Does this seem like a logical check to do?

Issue #2132, opened 2016-06-21
Refactor Analytics::Export


zazaian:
On current master the `Analytics::Export` module contains multiple behaviors for generating specific subsets of exportable data. This is a really nice pattern because all of these exports are going to deal with Mongoid and Analytics::Event records, so condensing all of these behaviors into a separate library makes a lot of sense.

The problem is that this process is very complicated, so the `Analytics::Export` module is very involved and some the methods doing the heaviest lifting like `#generate_csv` should be extracted to separate classes and not performed in-place as they're being here.

There's currently a nearly-complete implementation of this refactor on the `2132-refactor_analytics_export` branch, but it's not possible to test with production data at present because the method responsible for assembling the exportable data on `AnalyticsController#export` is broken. Once this functionality is fixed by integration of code from issue #2118, it will be possible to rebase this branch and test it against the properly-assembled analytics records.

Issue #2131, opened 2016-06-21
Professors should be able to schedule their analytics exports


zazaian:
Once we've added a scheduler for exporting course analytics on a regular basis, we should give professors control over how often those are generated. This will allow us to turn this off for professors which aren't using it, or generate them more frequently only for professors that rely heavily on this data, which will save lots of system resources.

Obviously professors will still have the ability to generate these files manually whenever they'd like to, but  adding this flexibility should both conserve resources for Gradecraft, and conserve time for professors who need to generate these exports consistently.

This can only be implemented once issue #2130 has been resolved.

Issue #2130, opened 2016-06-21
Analytics Exports should be performable at scheduled intervals


zazaian:
We need some kind of scheduled job to generate new analytics exports for classes on a regular basis of some kind so professors don't need to do this manually. This should be done after issues #2118 and #2128 have been completed.

Issue #2129, opened 2016-06-21
All Exporter classes should be performed with s3fs and Resque


zazaian:
Currently all `Exporter` classes are performing exports in-place. This is problematic for a few reasons:

1. The user has to remain present on the the page for the export to finish generating.
2. The export has to be generated in memory.
3. Exports have no persistence and must be re-generated on new requests even if the underlying data hasn't changed.

This isn't so bad for smaller courses or smaller export files, but for larger courses this could equate to long export generation times and very resource-intensive exports. Ideal solutions for these problems would be as follows:

1. Every export should be performed through either a dedicated job or a larger `Exporter` job that has the flexibility to accept multiple Exporter types and parameters.
2. Rather than being generated in place CSV files should be built on our `s3fs` persistent disk and then uploaded to the user's resource.
3. Like the rest of our exported data, final exports should be moved to S3 and then made available in the download center. For smaller files with very little generation time, this should probably render while the user is still on the page and also make it available in the download center for later access.

PR #2142, opened 2016-06-22
Fix Analytics Exports

zazaian:
The thrust of this PR is that the analytics export process was broken in multiple places and @chcholman needed this to be fixed to be able to generate a working export in order to continue work on her dissertation.

This PR should be considered a patch of the existing code to be used to get the existing analytics export implementation into a working state, and not a long-term design solution for the analytics export suite. Additional issues #2118, #2128, #2129, #2130, #2131, and #2132 have been created to incrementally issue enhancements to the overall functionality of the analytics export suite.

Some points of interest here:

- the `CourseEventExport` and `CoursePredictorExport` were using outdated method names for referencing predicted points and possible points values, and additionally had some unimplemented placeholder columns that were generating blank columns on the export CSV.

- the `AnalyticsController#export` method was generating one batch of files for each role and additionally a batch of files for all roles, which meant that all of the events were being queried for at least 9 times. Large lecture classes of 300+ students were taking over half an hour to complete an export.

- exports were persisting files on disk but not using s3fs, meaning that a docker container shift mid-export could destroy any already-exported data.

- CSV files were being concatenated rather than written individually in-place, so each CSV file was dumped into one master CSV file rather than being written individually.

Additionally please note that no additional spec files have been added to support this code -- these will come in with the pending issues once further refactoring is performed on the players involved.

**Note also** that this has been tested against multiple courses in production and it seems that any bugs related to data malformations have been ironed out.

### DIFF
--- a/app/analytics_exports/course_event_export.rb
+++ b/app/analytics_exports/course_event_export.rb
@@ -5,6 +5,7 @@ class CourseEventExport
 
   set_schema  username: :username,
               role: :user_role,
+              user_id: :user_id,
               page: :page,
               date_time: lambda { |event| event.created_at.to_formatted_s(:db) }
 
--- a/app/analytics_exports/course_event_export.rb
+++ b/app/analytics_exports/course_event_export.rb
@@ -6,7 +6,6 @@ class CourseEventExport
   set_schema  username: :username,
               role: :user_role,
               page: :page,
-              action: :page_name,
               date_time: lambda { |event| event.created_at.to_formatted_s(:db) }
 
   def schema_records_for_role(role)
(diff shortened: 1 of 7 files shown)


zazaian:
@acousticrobot please take a look at this when you get an opportunity. @chcholman please note that this is in the pipeline.

zazaian:
Awesome, @acousticrobot, I'm glad that this was so helpful in understanding the workflow here. There are so many moving pieces that it's been kind of a process to break this down into bite-sized chunks but delegating each component to a separate issue definitely helps to keep the process organized.

Thanks for your comments and for taking a look at this. Much appreciated :fireworks:

zazaian:

PR #2156, opened 2016-06-24
Refactor controllers/assignments/grades_controller_spec.rb to use World

chcholman:
This PR does a first pass at removing instance variables in the grades controller spec and replacing them with World

### DIFF
--- a/spec/controllers/assignments/grades_controller_spec.rb
+++ b/spec/controllers/assignments/grades_controller_spec.rb
@@ -3,6 +3,8 @@ require "rails_spec_helper"
 describe Assignments::GradesController do
   let(:world) { World.create.with(:course, :assignment, :student, :grade) }
   let(:professor) { create(:professor_course_membership, course: world.course).user }
+  let(:assignment) { world.assignment }
+  let(:student) { world.student }
 
   context "as professor" do
     before(:each) do
@@ -11,7 +13,7 @@ describe Assignments::GradesController do
 
     describe "GET download" do
       it "returns sample csv data" do
-        get :download, assignment_id: world.assignment, format: :csv
+        get :download, assignment_id: assignment, format: :csv
         expect(response.body).to \
           include("First Name,Last Name,Email,Score,Feedback")
       end
@@ -19,9 +21,9 @@ describe Assignments::GradesController do
 
     describe "GET edit_status" do
       it "assigns params" do
-        get :edit_status, { assignment_id: world.assignment.id, grade_ids: [world.grade.id] }
-        expect(assigns(:title)).to eq("#{world.assignment.name} Grade Statuses")
-        expect(assigns(:assignment)).to eq(world.assignment)
+        get :edit_status, { assignment_id: assignment.id, grade_ids: [world.grade.id] }
+        expect(assigns(:title)).to eq("#{ assignment.name } Grade Statuses")
+        expect(assigns(:assignment)).to eq(assignment)
         expect(assigns(:grades)).to eq([world.grade])
         expect(response).to render_template(:edit_status)
       end
@@ -29,22 +31,22 @@ describe Assignments::GradesController do
 
     describe "PUT update_status" do
       it "updates the grade status for grades" do
-        put :update_status, { assignment_id: world.assignment.id, grade_ids: [world.grade.id], grade: { status: "Graded" }}
+        put :update_status, { assignment_id: assignment.id, grade_ids: [world.grade.id], grade: { status: "Graded" }}
         expect(world.grade.reload.status).to eq("Graded")
       end
 
       it "redirects to session if present"  do
         session[:return_to] = login_path
-        put :update_status, { assignment_id: world.assignment.id, grade_ids: [world.grade.id], grade: { status: "Graded" }}
+        put :update_status, { assignment_id: assignment.id, grade_ids: [world.grade.id], grade: { status: "Graded" }}
         expect(response).to redirect_to(login_path)
       end
     end
 
     describe "GET export" do
       it "returns sample csv data" do
-        submission = create(:submission, grade: world.grade, student: world.student,
-                            assignment: world.assignment)
-        get :export, assignment_id: world.assignment, format: :csv
+        submission = create(:submission, grade: world.grade, student: student,
+                            assignment: assignment)
+        get :export, assignment_id: assignment, format: :csv
         expect(response.body).to \
           include("First Name,Last Name,Email,Score,Feedback,Raw Score,Statement")
       end
@@ -52,12 +54,12 @@ describe Assignments::GradesController do
 
     describe "GET export_earned_levels" do
       it "returns example earned levels data" do
-        rubric = create(:rubric_with_criteria, assignment: world.assignment)
+        rubric = create(:rubric_with_criteria, assignment: assignment)
         rubric.criteria.each do |criterion|
           level = Level.create(criterion_id: criterion.id, name: "Sushi Success", points: 2000)
-          CriterionGrade.create(criterion: criterion, level_id: level.id, student: world.student, points: 2000, assignment: world.assignment)
+          CriterionGrade.create(criterion: criterion, level_id: level.id, student: student, points: 2000, assignment: assignment)
         end
-        get :export_earned_levels, assignment_id: world.assignment, format: :csv
+        get :export_earned_levels, assignment_id: assignment, format: :csv
 
         expect(response.body).to \
           include("First Name,Last Name,Email,Username,Team")
@@ -66,9 +68,9 @@ describe Assignments::GradesController do
 
     describe "GET import" do
       it "displays the import page" do
-        get :import, { assignment_id: world.assignment.id}
-        expect(assigns(:title)).to eq("Import Grades for #{world.assignment.name}")
-        expect(assigns(:assignment)).to eq(world.assignment)
+        get :import, { assignment_id: assignment.id}
+        expect(assigns(:title)).to eq("Import Grades for #{assignment.name}")
+        expect(assigns(:assignment)).to eq(assignment)
         expect(response).to render_template(:import)
       end
     end
@@ -79,44 +81,44 @@ describe Assignments::GradesController do
       let(:file) { fixture_file "grades.csv", "text/csv" }
 
       it "renders the results from the import" do
-        world.student.reload.update_attribute :email, "robert@example.com"
+        student.reload.update_attribute :email, "robert@example.com"
         second_student = create(:user, username: "jimmy")
         second_student.courses << world.course
-        post :upload, assignment_id: world.assignment.id, file: file
+        post :upload, assignment_id: assignment.id, file: file
         expect(response).to render_template :import_results
         expect(response.body).to include "2 Grades Imported Successfully"
       end
 
       it "renders any errors that have occured" do
-        post :upload, assignment_id: world.assignment.id, file: file
+        post :upload, assignment_id: assignment.id, file: file
         expect(response.body).to include "3 Grades Not Imported"
         expect(response.body).to include "Student not found in course"
       end
 
       it "adds error and redirects without a file" do
-        post :upload, assignment_id: world.assignment.id
+        post :upload, assignment_id: assignment.id
         expect(flash[:notice]).to eq("File is missing")
-        expect(response).to redirect_to(assignment_path(world.assignment))
+        expect(response).to redirect_to(assignment_path(assignment))
       end
     end
 
     describe "GET index" do
       it "redirects to the assignments show view if the assigment is not a rubric" do
-        allow(world.assignment).to receive(:grade_with_rubric?).and_return false
-        get :index, assignment_id: world.assignment.id
-        expect(response).to redirect_to assignment_path(world.assignment)
+        allow(assignment).to receive(:grade_with_rubric?).and_return false
+        get :index, assignment_id: assignment.id
+        expect(response).to redirect_to assignment_path(assignment)
       end
     end
 
     describe "GET mass_edit" do
       it "assigns params" do
-        get :mass_edit, assignment_id: world.assignment.id
-        expect(assigns(:title)).to eq("Quick Grade #{world.assignment.name}")
-        expect(assigns(:assignment)).to eq(world.assignment)
-        expect(assigns(:assignment_type)).to eq(world.assignment.assignment_type)
-        expect(assigns(:assignment_score_levels)).to eq(world.assignment.assignment_score_levels)
+        get :mass_edit, assignment_id: assignment.id
+        expect(assigns(:title)).to eq("Quick Grade #{assignment.name}")
+        expect(assigns(:assignment)).to eq(assignment)
+        expect(assigns(:assignment_type)).to eq(assignment.assignment_type)
+        expect(assigns(:assignment_score_levels)).to eq(assignment.assignment_score_levels)
         expect(assigns(:grades)).to eq([world.grade])
-        expect(assigns(:students)).to eq([world.student])
+        expect(assigns(:students)).to eq([student])
         expect(response).to render_template(:mass_edit)
       end
 
@@ -124,7 +126,7 @@ describe Assignments::GradesController do
         student_2 = create(:user, last_name: "zzimmer", first_name: "aaron")
         student_3 = create(:user, last_name: "zzimmer", first_name: "zoron")
         [student_2,student_3].each {|s| s.courses << world.course }
-        expect{ get :mass_edit, assignment_id: world.assignment.id }.to \
+        expect{ get :mass_edit, assignment_id: assignment.id }.to \
           change{Grade.count}.by(2)
         expect(assigns(:grades)[1].student).to eq(student_2)
         expect(assigns(:grades)[2].student).to eq(student_3)
@@ -133,9 +135,9 @@ describe Assignments::GradesController do
       context "with teams" do
         it "assigns params" do
           team = create(:team, course: world.course)
-          team.students << world.student
-          get :mass_edit, assignment_id: world.assignment.id, team_id: team.id
-          expect(assigns(:students)).to eq([world.student])
+          team.students << student
+          get :mass_edit, assignment_id: assignment.id, team_id: team.id
+          expect(assigns(:students)).to eq([student])
           expect(assigns(:team)).to eq(team)
         end
       end
@@ -143,7 +145,7 @@ describe Assignments::GradesController do
 
     describe "PUT mass_update" do
       let(:grades_attributes) do
-        { "#{world.assignment.reload.grades.index(world.grade)}" =>
+        { "#{assignment.reload.grades.index(world.grade)}" =>
           { graded_by_id: professor.id, instructor_modified: true,
             student_id: world.grade.student_id, raw_points: 1000, status: "Graded",
             id: world.grade.id
@@ -152,14 +154,14 @@ describe Assignments::GradesController do
       end
 
       it "updates the grades for the specific assignment" do
-        put :mass_update, assignment_id: world.assignment.id,
+        put :mass_update, assignment_id: assignment.id,
           assignment: { grades_attributes: grades_attributes }
         expect(world.grade.reload.raw_points).to eq 1000
       end
 
       it "timestamps the grades" do
         current_time = DateTime.now
-        put :mass_update, assignment_id: world.assignment.id,
+        put :mass_update, assignment_id: assignment.id,
           assignment: { grades_attributes: grades_attributes }
         expect(world.grade.reload.graded_at).to be > current_time
       end
@@ -167,7 +169,7 @@ describe Assignments::GradesController do
       it "only sends notifications to the students if the grade changed" do
         world.grade.update_attributes({ raw_points: 1000 })
         run_background_jobs_immediately do
-          expect { put :mass_update, assignment_id: world.assignment.id,
+          expect { put :mass_update, assignment_id: assignment.id,
                    assignment: { grades_attributes: grades_attributes } }.to_not \
             change { ActionMailer::Base.deliveries.count }
         end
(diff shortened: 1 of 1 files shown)


spec/controllers/assignments/grades_controller_spec.rb
```diff
@@ -1,74 +1,63 @@
...
-    @student = create(:user)
-    @student.courses << @course
-  end
-  before(:each) do
-    @grade = create :grade, student: @student, assignment: @assignment,
-      course: @course
-  end
-  after(:each) { @grade.delete }
+  let(:world) { World.create.with(:course, :assignment, :student, :grade) }
+  let(:professor) { create(:professor_course_membership, course: world.course).user }
```
acousticrobot: Since there are so many references to `world.assignment` in the specs, you can opt to add in here:

```
let(:assignment) { world.assignment }
```

Then all the remaining references to `@assignment` can more succinctly be just `assignment`. Also true with `world.student`, `world.grade` etc. Very much a judgement call though, how many spec references warrant an additional `let` statement, and one I'd like to hear other opinions on.

zazaian:
I'm actually really not fond of that syntax because it seems dangerously magical to me:

```
let(:world) { World.create.with(:course, :assignment, :student, :grade) }
```

Do these objects have any relationship to one another? Am I just creating each one independently? Is creation all that's happening? What's creating them? I think actually the lack of explicit relationships or any kind of hierarchy is what makes this feel really kind of cryptic from an outsider's perspective. If I'd written the code for this class I'm sure I'd understand what's happening, but it seems like an outsider should be able to understand the outcome without having to pick apart the innards of the library itself.

Also I find the `create` method really confusing here since it's actually just instantiating a new object. Why is that a valuable syntax? It seems really confusing to me since `create` is a designated `ActiveRecord` method for immediately persisting a record, and the `World` itself isn't being persisted anywhere.

I think something like this feels much more expressive:

```
let(:world) do
  World.new.create course: [assignment, :student, :grade]
end
```

Initializing a World instance with `#new` rather than cryptically instantiating an instance of the `Instance` just seems more declarative to me: I understand what's probably taking place inside the class itself. Also reserving a `create` method to suggest that these resources are in fact being created additionally seems straightforward.

With this syntax you could additionally do the following without having to really change this logic at all:

```
# just build these objects but don't persist them
World.new.build course: [assignment, :student, :grade]

# create these relationships with doubles
World.new.double course: [assignment, :student, :grade]
```

If it feels valuable to pull out the `new` keyword then you could just create class methods for '.build`, '.create`, and '.double` that return an instance of world rather than simply being calls on a world instance itself.

spec/controllers/assignments/grades_controller_spec.rb
zazaian: From a first glance I agree that the addition of `world.` to almost every variable drastically impacts readability. If nearly everything is appended by `world.` I wonder whether we need it. Using let blocks to clean these up sounds like a good idea to me:

```ruby
let(:assignment) { world.assignment }
let(:student) { world.student }
```

But that I guess brings me back to the question of what we're getting out of this implementation. Is it valuable to obfuscate all of the logic that's providing these resources? Should we just presume that a `:student` or an `:assignment` has been built or created with all of its resources? Are we just generating the associations that we need for the specs we've built or is every association on this object being accounted for?

I guess the reason I'd be hesitant to continue moving forward with the `World` implementation as it currently stands is that it's very difficult to work backward from. If I were trying to translate the code for the current world implementation:

```ruby
World.create.with(:course, :assignment, :student, :grade)
```

With a more straightforward implementation, or to strip it out altogether, I'd presume from looking at it that I should be able to do the following:

```ruby
let(:course) { create(:course) }
let(:assignment) { create(:assignment) }
let(:student) { create(:student) }
let(:grade) { create(:grade) }
```

...which is a bit of an oversimplification, but it's not unthinkable that this is what's being asked for when the `World` is constructed`.

At least with the slightly longer-handed code...
```ruby
let(:course) { create(:course_accepting_groups) }
let(:assignment) { create(:assignment, course: course) }
let(:student) { create(:user) }
before { student.courses << course }
```

...I don't have to wonder what relationships I'm hoping to get out of the `World` definition, and I don't have to dig around in the `World` class to figure that out. And even in digging through `World`, because there are so many `||` cases etc, I really wouldn't know what the spec is expecting to be there beyond just presuming that the spec was probably depending too much on the `World` method to magically provide what you hope will be there. 

@chcholman If this all seems totally irrational let me know and i'll move forward and just comment out the implementation, but the bigger issue as it stands in my eyes is that we should probably have a more coherent DSL for the `World` before we expend the effort to swap out all of our test setup logic with it.

jwright:
@chcholman Overall, I think this is a huge improvement. I agree with @acousticrobot and @zazaian above regarding adding shortcut `let`s will eliminate some of the duplication.

In regards to the `World` implementation, I think it is great. I love the simplicity that @acousticrobot  added around being able to create objects easily with just symbols.

Because our object model is so interdependent, I don't think there is any advantage of creating objects that show you how they were created. This is because the specs for those objects should describe how they are created. Too much explicit setup logic would result in a further complicated spec suite. The `World` is for higher level abstractions as to how the "world" is generated. 

A lot of setup logic in some instances is probably an artifact that we are testing things sometimes at the wrong level. Using a controller spec to test a model feature etc. I think we would benefit from models that take on actions (similar to our simple service library we talked about a while ago). As a simple example, the [`PasswordsControllerSpec`](https://github.com/UM-USElab/gradecraft-development/blob/b9c9cf4b56663b778c0f86fda3ab59e7c9a17e57/spec/controllers/passwords_controller_spec.rb#L11) should probably just check if a `ResetsPasswords` service was called (using a spy) and the specs for a mythical `ResetsPasswords` service should check the actual implementation. If we did that, then we would probably would have less need for a `World` in setup.

I do believe the `World` may benefit from a `build` vs. `create` methods and maybe even a `double` method but perhaps not if we reserve that the `World` is for creating an actual world for higher level specs (controller and above). If we do add those methods, I would recommend that `World` be moved into a lib folder and tested like an OSS gem since that is test logic and it would need tests for anything more complicated. I have not run into an instance where we needed those methods for faster specs since most of the time I used it, those objects needed to live in the database. I could see where a `double` method may be of benefit if you don't really want to create all of the ceremony around running a background job or sending a mailer for example.

My 3 cents.

zazaian:
Hey @jwright and @jonathan -- I didn't see these responses until earlier this afternoon. I'll respond to them fully either tonight or early tomorrow.

zazaian:
I'm going to try to respond to everything bit by bit here because there's a lot flying around and I want to make sure that I don't miss anything.

@jwright:
> Because our object model is so interdependent, I don't think there is any advantage of creating objects that show you how they were created. This is because the specs for those objects should describe how they are created. Too much explicit setup logic would result in a further complicated spec suite. The World is for higher level abstractions as to what the "world" is needs.

This isn't really a matter of being shown how the objects were created, it's a matter of knowing what's there, and how the things that are there are associated. As I mentioned before, if tell anything to create a one-dimension array of objects I'm going to assume that either they have no relationship to each other, or they have a relationship to the object that created them. Beyond saying "well you should just know how those objects are connected" I feel like there's a massive logical disconnect in how these are being provided.

Additionally this code is pretty much impossible to work backward from. We're pretty much tethering ourselves to our world implementation, which means that the specs themselves are becoming dependent on this specific manner of provisioning test resources rather than building only what needs to be present for testing. Consider that the code `let(:world) { World.create.with(:course, :assignment, :student, :grade) }` is the equivalent of:

```ruby
let(:world) { World.create }

let(:course) do
  course = FactoryGirl.create :course
  world.courses << course
  course
end

let(:assignment) do
  course = world.course || FactoryGirl.build(:course)
  assignment_type = FactoryGirl.create(:assignment_type, course: course)
  world.assignment_types << assignment_type
  assignment = FactoryGirl.create(:assignment, course: course, assignment_type: assignment_type)
  world.assignments << assignment
  assignment
end

let(:student) do
  course = world.course || FactoryGirl.build(:course)
  user = FactoryGirl.create(:user)
  course_memberships << FactoryGirl.create(:course_membership,
    course: course,
    user: user,
    role: "student") if course
  users << user
  user
end

let(:grade) do
  assignment = world.assignment || FactoryGirl.build(:assignment)
  assignment_type = attributes.delete(:assignment_type) || assignment.assignment_type
  course = world.course || FactoryGirl.build(:course)
  student = world.student || FactoryGirl.build(:user)
  grade = FactoryGirl.create :grade,
    assignment: assignment,
    assignment_type: assignment_type,
    course: course,
    student: student
  grades << grade
  grade
end
```

Is that what you'd expect `let(:world) { World.create.with(:course, :assignment, :student, :grade) }` to do? Where am I being told that a grade will build an assignment and an assignment type and a course and a user? Why can't many of these default association behaviors be defined using `FactoryGirl` sub factories? Why should I presume that those things are being provisioned? What if I genuinely just want a grade but without the 5 additional resources that it builds, should I not use `World` at all in this context?

The lack of an explicit, declarative DSL for how we're provisioning test objects, regardless of how compact the code might be, makes the testing process something of a guessing game, and one in which you need to become intimately familiar with the innards of the World in order to avoid provisioning resources that don't need to exist for testing the `subject`.

@jwright:
> A lot of setup logic in some instances is probably an artifact that we are testing things sometimes at the wrong level. Using a controller spec to test a model feature etc. I think we would benefit from models that take on actions (similar to our simple service library we talked about a while ago). As a simple example, the PasswordsControllerSpec should probably just check if a ResetsPasswords service was called (using a spy) and the specs for a mythical ResetsPasswords service should check the actual implementation. If we did that, then we would probably would have less need for a World in setup.

This I absolutely agree with, and think that we should be much more vigilant about what resources are being created for testing (creating extras just for the sake of it can lead to false-positives), and about confining our spec examples to testing only the `described_class` and `subject` and as few of the behaviors of other classes as possible. Additionally stubbing and doubling where possible will help us cut way, way back on the resource overhead for this stuff.

@acousticrobot:
>  Do we now need to use world.course.assignment in the specs? and if not, isn't world.assignment even more ambiguous?

This is a really astute observation and might be the lynch pin for the current implementation of the world. By trying to implicitly manufacture as many of these relationships as possible we inevitably lose sight of the ability to figure out how these things are actually connected. As @jwright said maybe that's not important, but it still feels kind of curious to me to be to be carving out actually a very specific workflow for these relationships in `World`, but then masking all of them in `World`'s instantiation.

@acousticrobot:
> What balance do we find between full featured / custom overrides VS. rapid setup of a group of behaves-as-expected models? Currently we have a lot of custom overrides that we don't yet take advantage of in the specs, and the "magic" method which would only require a small subset of current code to operate. If it it too confusing to read and use as-is perhaps stripping away functionality, rather that adding to it, is the solution.

So this seems like maybe the most important distinction to make here, which @jwright also brought up, namely where do we need a tool that gives us very simple setup, and where do we need more explicit setup for specs? I wonder if this question points to the idea that we actually need something like the World, but not that acts as a one-size-fits-all solution to provisioning spec resources.

Maybe what we really need is a few smaller, more refined iterations of `World` rather than one giant `World` that attempts to handle every case. Maybe something like a `World::Context` class, which models the data needed for a given user scenario, gives us the best of both worlds.

An additional need seems to be the ability to quickly define associations with the other objects being created without having to go through the laundry list of everything-is-related-to-everything for reach individual resource.

I'm just spitballing here, but maybe something like this gives us a better balance of explicitness and simplicity:

```ruby
class StudentInACourse < World::Context
  # define the resources that we want our context to have
  default_resources :student, :grade, :course, :assignment, :assignment_type

  # how should we construct resources by default?
  default_constructor :build

  def grade
    # instantiate a grade using :build or :create, and then create associations with the other resources
    @grade ||= FactoryGirl.send constructor, :grade, attrs.delete(:grade, :assignment_type)
  end

  def assignment
    @assignment ||= FactoryGirl.send constructor, :assignment, attrs.delete(:assignment)
  end

  def course
    @course ||= FactoryGirl.send constructor :course
  end

  def student
    # define the student in a factory rather than in the world context
    @student ||= FactoryGirl.send constructor, :student, attrs.delete(:student)
  end
end

module World
  class Context
    cattr_accessor :default_resources, :default_constructor    
    attr_accessor: :resources, :constructor

    def initialize(constructor: nil, resources: nil)
      @constructor = constructor || class.default_constructor
      @resources = resources || class.default_resources
    end
    
    # don't use some of the resources we've defined in default_resources
    def without(*omitted_resources)
      resources.delete! omitted_resources
      self
    end

    # use only the resources we're defining here
    def only(*only_resources)
      self.resources = only_resources
      self
    end

    # produce a hash in format of { class: class, student: student, assignment: assignment } etc...
    def attrs
       class.resources.inject({}) do |memo, attr|
         memo[attr] = send attr
       end
    end

  end
end
```

Or a slightly more involved version that abstracts some of the `FactoryGirl` logic and moves us closer to our own DSL:

```ruby
class StudentInACourse < World::Context
  default_resources :student, :grade, :course, :assignment, :assignment_type
  default_constructor :build

  def grade
    @grade ||= build_with_associations :grade, except: [:assignment_type]
  end

  def assignment
    @assignment ||= create_with_associations :assignment
  end

  def course
    @course ||= create :course
  end

  def student
    # define the student in a factory rather than in the world context
    @student ||= construct_with_associations :student, except: [:assignment_type]
  end
end

module World
  class Context
    cattr_accessor :default_resources, :default_constructor    
    attr_accessor: :resources, :constructor

    def initialize(constructor: nil, resources: nil)
      @constructor = constructor || class.default_constructor
      @resources = resources || class.default_resources
    end

    def without(*omitted_resources)
      resources.delete! omitted_resources
      self
    end

    def only(*only_resources)
      self.resources = only_resources
      self
    end
   
    def build_with_associations(resource_name, except: [])
      FactoryGirl.build resource_name, attrs.delete(resource_name, *except)
    end

    def create_with_associations(resource_name, except: [])
      FactoryGirl.create resource_name, attrs.delete(resource_name, *except)
    end

    def construct_with_associations(resource_name, except: [])
      FactoryGirl.send constructor, resource_name, attrs.delete(resource_name, *except)
    end

    def create(resource_name)
      FactoryGirl.create resource_name
    end

    def build(resource_name)
      FactoryGirl.build resource_name
    end

    def construct(resource_name)
      FactoryGirl.send constructor, resource_name
    end

    def attrs
       class.resources.inject({}) do |memo, attr|
         memo[attr] = send attr
       end
    end

  end
end
```

and then in a spec you would be able to do any of these:

```ruby
let(:context) { StudentInACourse.new }
# or
let(:ungraded_context) { StudentInACourse.new.without :grade }
# or
let(:empty_course) { StudentInACourse.new.only :course, :student }
```

Maybe this suffers from some of the same opaqueness that I've been complaining about in the `World` implementation, but at least it seems like it organizes the resources in a more flexible way, and one that seems to more easily suggest where I can find the resources I'm building.

jwright:
@zazaian Thanks for the great response. Very nice to see your thinking here.

I agree that it is valuable to make it easy for the reader to know what the "context" is...but personally, I have not come across this need yet but it may be because I am too close to the implementation and the all of us knowing what is being created.

I do like the "context" implementation. It's similar to fixtures. I would see a DSL being very valuable where we could do that. I think this is a larger discuss, much larger than this PR. I think this PR is good where it's at but I think we should create another PR or have another discussion to come up with this new `World`.

JangoSteve:
Hey everyone, sorry just catching up and reading through all this. Keep in mind as well, this is the first I've heard of `World` in GC and the first I've looked at it, so I'm probably missing some contextual and historical info.

Overall, I agree with all of Mike's points, and I like the reusability and composability of his proposed `World::Context`s, though like Jamie, I wonder how practical and useful that is immediately and would caution against premature optimization in that regard and rather view that as an eventual goal which everything we're building now can take baby steps toward accomplishing over time. So, like Jonathan suggested, I'd probably watch the time spent on this and focus on what's needed for now. As long as each PR makes the code a little better than it was before the PR, and we have a vision for the future, we're moving in the right direction.

Like Jamie said, I think this PR is almost good enough to merge and then start a new PR with a discussion focused on improving the DSL of `World`. The only change I'd make before merging this PR is adding the

```ruby
let(:assignment) { world.assignment }
let(:student) { world.student }
```

that Jonathan suggested, which will make the commit more readable and the PR a bit cleaner.

For the future though, I especially agree with Mike's points about not being a fan of this:

```ruby
World.create.with(:course, :assignment, :student, :grade)
```

Especially consider that the following will result in completely different results:

```ruby
World.create.with(:course, :grade)
#=> Will give World one Course, and a grade that belongs to the course
World.create.with(:grade, :course)
#=> Will give World two Courses, with a grade that belongs to the first course
```

The above may be intended behavior given the presumption that we're substituting order for hierarchy in flatly-declared arguments to `World`, but you'd literally need to read through the three methods (as I did), `World#with`, `World#create_course`, and `World#create_grade`, line-by-line to understand the difference between the two above instantiations.

A better API I think would be something along the lines of what's been proposed already (maybe a little different), like:

```ruby
let(:world) { World.new.with(:student, :student, :course, :course, course: {assignment: [{student: :grade}, {student: :grade}]}).create }
```

It's not perfect in that it doesn't allow a full range of relationships to be declared (since we're using nesting to declare hierarchy, it only allows for tree-like structures and doesn't necessarily reflect e.g. the multi-assignment nature of `:grade` in that it belongs directly to both the `:student` and the `:assignment`), but at least it's obvious without knowing much about `World` that I've generated two unrelated students, two unrelated courses, and a third course with an assignment having a third and fourth student, each with a grade.

By moving `create` to the end, we allow for reusing the `with` builder and swapping out `build` or whatever else to decide how we want to handle persistence/memory.

I haven't fully thought it through, but with the above syntax, my initial thought is that it'd add complexity to the definition of the `with` method (in each `item` loop, it'd need to check if the current item is a `Symbol` or `Hash` to determine how to structure the new instance association), but it seems like it could reduce the complexity of the various `create_#{item}` methods by getting rid of a lot of those `||` statements.

Also, everything would just `build` instead of `create`, and the `create` method would loop through the instances in `World` and call `save` on each of them essentially.

With the above refactoring and API, the original example would look like this:

```ruby
World.create.with(course: {assignment: {student: :grade}})
```

Again though, I think just creating those convenience `let` methods is probably good enough at least to merge this for now and then start thinking about a potential (time-boxed) refactoring.

zazaian:
So I like @JangoSteve's nested world declaration solution, and like too the idea of performing the create on the world after the relationships and data hierarchies have been established, but I suppose if we're going this route I might take it a step further to simplify some of the construction logic:

```
let(:world) do
  World.new.
    .students(2)
    .courses(2)
    .course({
      assignment: [{student: :grade}, {student: :grade}]
    }).create
end
```

Just another way to do it. That said we could probably hack DSL iterations to death without really getting to the root of the problem, which seems to be figuring out what problem we're trying to solve.

- Are we just trying to get as much setup code as possible out of the specs files?
- How much do we care about cycles lost from provisioning unneeded resources?
- How much do we care about the potential of bugs arising from unintentionally-instantiated resources as in Steve's example of `:course, :grade` producing different results than `:grade, :course`?

I guess those are all pretty important questions to answer before going much further down the path to a different solution here.

So I agree overall that using the world and lets instead of instance variables is definitely an improvement, my key gripe at the moment being that because of the de facto orientation of the current `World` to over-provision resources that, if we end up implementing a different solution later, we will probably end up with some spec bugs from the swap-out. If that's worth getting all of the instance variables out of the specs right now then we should probably go ahead with it and just keep having the world discussion.

Personally I haven't run into many issues where I felt that I needed the current iteration of `World`, and think that just swapping everything out with `let` calls is probably a safer bet before we get to a more final iteration of `World`. Perhaps to pare down some of the setup complexity that is being complained about we can just use more situationally-specific sub factories?

Either way I'm fine with the work done in this commit, just trying to save us from some potential headache down the road.

JangoSteve:
I also share @zazaian's points in the last comment, though anyone who's coded with me knows I tend to lean toward explicitness versus magic and declarations versus abstractions. But I'm not the one coding on GC everyday, so I don't know the pain.

One thing I would advise would be to think carefully about what functionality FactoryGirl provides and how/why this can't be done with that (I'm sure it can't, but I'm a bit fuzzy on exactly what benefit this provides over just using FactoryGirl a bit more toward its full capabilities, which I think gets into Mike's points about really figuring out what the problem is we're trying to solve).

Also, I think that latest iteration of proposed syntax looks sweet.

JangoSteve:
I guess this is where my feedback may be limited by my lack of context, but generally when I've come across the issue of generating contexts of associations, I've just used the built-in way FactoryGirl has of doing that (particularly the bit at the end of the ["Associations" section](https://github.com/thoughtbot/factory_girl/blob/master/GETTING_STARTED.md#associations) of the Getting Started guide).

I would just define the associations I want in Factory girl and then in the spec, I'd do something like:

```ruby
create(:grade_with_associations)
# or
create(:course_with_student_and_grades)
```

Or something like that. Similarly to what @acousticrobot found with how many specs use `World` currently, I usually end up with 20 or whatever specs using e.g. my `:object_with_other_objects` factory throughout the test suite.

FactoryGirl has a lot of interesting features and capabilities, including some that could be relevant to this discussion in ["Inheritence"](https://github.com/thoughtbot/factory_girl/blob/master/GETTING_STARTED.md#inheritance), ["Modifying Factories"](https://github.com/thoughtbot/factory_girl/blob/master/GETTING_STARTED.md#modifying-factories), ["Building or Creating Multiple Records"](https://github.com/thoughtbot/factory_girl/blob/master/GETTING_STARTED.md#building-or-creating-multiple-records), and ["Custom Construction"](https://github.com/thoughtbot/factory_girl/blob/master/GETTING_STARTED.md#custom-construction).

Again though, I'm probably missing some nuance of the problem being solved here.

zazaian:
Hey guys -- so I've got some more thoughts here but I want to make sure that I've digested everything that's been mentioned prior to commenting further. I'll chime in with some additional observations tomorrow.

zazaian:
Hey guys, so I've been thinking about this a lot and I spent a lot of time yesterday reviewing how `World` is being used across the app as per Jonathan's list above, and I have some thoughts that I'll just enumerate to keep things organized:

#### 1. If we're staying the course with this current implementation, we should probably just be using `World` in controllers.

When we're building contextually-rich instances `World`, such as with `:course, :student, :assignment, :grade, :criterion`, we have to create instances of each of those resource classes every time the world is used at all, regardless of whether we need those specific resources for the given example being run. For a few examples this isn't such a big deal, but for hundreds of them it's probably slowing down the spec suite quite a bit. Being cautious about using `World` only in instances where it's truly needed will help make up for the processing overhead it might be responsible for.

#### 2. If we're going to position this as a "development tool" that doesn't intrinsically require a clean DSL, we should probably at least reduce some of the duplicated syntax for building objects to make it easier to quickly check what each method is provisioning.

For example, I wonder if the "open world" concept might be more usable if instead of this:

```ruby
    def create_criterion_grade(attributes={})
      assignment = attributes.delete(:assignment) || self.assignment || FactoryGirl.build(:assignment)
      student = attributes.delete(:student) || self.student || FactoryGirl.build(:user)
      rubric = attributes.delete(:rubric) || self.rubric || FactoryGirl.build(:rubric, assignment: assignment)
      criterion = attributes.delete(:criterion) || self.criterion || FactoryGirl.build(:criterion, assignment: assignment)
      criterion_grades << FactoryGirl.create(:criterion_grade, \
        attributes.merge(assignment: assignment, student: student, criterion: criterion))
      self
    end
```

...we condensed some of the logical redundancy so it looks more like this:

```ruby
    def create_criterion_grade(attributes={})
      assignment = assign_or_build :assignment, attributes
      student = assign_or_build :user, as: :student
      rubric = assign_or_build :rubric, assignment: assignment
      criterion = assign_or_build :criterion, assignment: assignment
      create_for_collection :criterion_grade,
        assignment: assignment,
        student: student,
        criterion: criterion

      self
    end

    private

    def assign_or_build(resource_name, attrs={})
      attrs.delete resource_name ||
        self.send resource_name ||
        FactoryGirl.build resource_name, attrs
    end

    def create_for_collection(resource_name, attrs={})
      resource = FactoryGirl.create resource_name, attrs
      send(resource_name.pluralize) << resource
      resource
    end

```

#### 3. We should be clear that one of the central ideas of this design is its flatness.

I think one of the things @JangoSteve and I were keying in on is that this implementation is really only useful for single-level resources rather than a broader, richer spec context for more elaborate relationship declarations. Maybe this implementation would be more succinct if we could enforce this aspect of the `World` somehow?

#### 4. The order of object instantiation matters, so maybe we should enforce it.

The example that @JangoSteve mentioned where `:course, :grade` creates a different outcome than `:grade, :course` seems to undermine the integrity of the simplicity `World` is reaching for. By hardcoding one or two workflows into the world we could at least guarantee that it's being used as it's been intended to, rather than perhaps getting some unintended outcomes from not following the exact order of resource instantiation that's required.

For example, something like this:

```ruby
class CourseWorld < World::Base
  order :course, :student, :assignment, :grade, :criterion, 
end
```

Would raise an `OrderError` if we tried to instantiate it with:
```ruby
CourseWorld.new  :assignment, :student, :course, :grade
# or
CourseWorld.new  :grade, :student
```

It seems like almost everyone likes the simplicity of the `:create, :these, :resources, :by_name` format, so perhaps some of these workflow reinforcements might help reduce the instance of error, and enhance clarity of workflow if built into the utility of the class itself.

#### 5. It might add some clarity to mutually exclusive workflows if they were housed in separate Worlds

Course-centric `World` resources, for example, seem to behave differently than rubric-centric `World` resources, because almost all of them need a course. Rather than explicitly merging a course into the attributes for each course-centric method, maybe `CourseWorld#create_for_collection could associate the course by default to save lots of course association declarations.

Here's a quick condensation of some of these ideas that might provide some structure to how we're constructing our worlds.

```ruby
class CourseWorld < World::Base
  members :course, :assignment, :grade, :criterion
  member_order :course, :assignment, :student, :grade

  def create_assignment(attributes={})
    assign_or_build :course, attributes
    assignment_type = create_for_collection :assignment_type
    create_for_collection :assignment,
      attributes.merge(assignment_type: assignment_type)
    self
  end

  def create_assignment_type(attributes={})
    assign_or_build :course, attributes
    create_for_collection :assignment_type
    self
  end
end

module World
  class Base
    cattr_accessor :member_names, :member_order, :requested_members

    def initialize(*requested_members)
      @requested_members = requested_members
      raise MetaWorld::Errors::OrderError unless valid_order?
      define_collections
      define_getters
    end

    class << self
      def members(*member_names)
        @member_names = member_names
      end

      def order(*member_order)
        @member_order = member_order
      end
    end

    private

    def valid_order?
      # check whether the requested members have been requested in the order
      # compliant with the defined member_order
      #
      requested_members == self.class.member_order.inject([]) do |memo, member|
        memo << member if requested_members.include? member
        memo
      end
    end

    # define empty collections by name for each member type. For example, calling:
    # members :course will set @courses = []
    def define_collections
      self.class.member_names.each do |name|
        collection_name = "@#{name.to_s.pluralize}".to_sym
        instance_variable_set collection_name, []
      end
    end

    # define methods for grabbing the first member of a given resource type out of its collection
    def define_getters
      self.class.member_names.each do |name|
        define_method name do
          send(name.to_s.pluralize).first
        end
      end
    end

    def assign_or_build(resource_name, attrs={})
      attrs.delete resource_name ||
        self.send resource_name ||
        FactoryGirl.build resource_name, attrs
    end

    # creates a resource by name and adds it to the World collection of the same name
    def create_for_collection(resource_name, attrs={})
      resource = FactoryGirl.create resource_name, attrs
      send(resource_name.pluralize) << resource
      resource
    end

    def create_for_course_collection(resource_name, attrs={})
      create_for_collection resource_name, attrs.merge course: course
    end
  end
end
```

PR #2159, opened 2016-06-24
Not null fields on assignments

chcholman:
This is the final PR in a series cleaning up the Assignments table. I've confirmed that all data we're adding not-null values to has been properly cleaned up in production.


jwright:
@chcholman This looks good. I am a little skittish that we have all of the instances where we create/update assignments are covered by having these now required columns. Are we sure these are covered by tests?

chcholman:
@jwright Excellent point - I'll confirm and have you take another look if need be. Thank you!

PR #2162, opened 2016-06-24
1521 count top grades only

chcholman:
This PR adds a feature that allows instructors to declare a maximum number of grades that count from a single assignment type and properly totals the student's course score.

### DIFF
--- a/app/models/assignment_type.rb
+++ b/app/models/assignment_type.rb
@@ -4,7 +4,7 @@ class AssignmentType < ActiveRecord::Base
   acts_as_list scope: :course
 
   attr_accessible :max_points, :name, :description, :student_weightable,
-    :position
+    :position, :top_grades_counted
 
   belongs_to :course, touch: true
   has_many :assignments, -> { order("position ASC") }, dependent: :destroy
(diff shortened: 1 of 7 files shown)


app/models/assignment_type.rb
```diff
@@ -73,27 +79,53 @@ def weighted_total_for_student(student)
   end
 
   def visible_score_for_student(student)
-    score = score_for_student(student)
-    if (max_points > 0) && score > max_points
-      return max_points
+    if count_only_top_grades?
+      return summed_highest_scores_for(student)
+    elsif is_capped?
+      return max_points_for_student(student)
```
acousticrobot: 💘 Love how much better this reads!

app/models/assignment_type.rb
```diff
@@ -73,27 +79,53 @@ def weighted_total_for_student(student)
...
     else
-      return score
+      return score_for_student(student)
     end
   end
 
-  def score_for_student(student)
+  def grades_for(student)
```
acousticrobot: I feel like this should be `visible_grades_for(student)`, but that also isn't entirely accurate (really it would be something horrible like `grades_that_count_towards_score_for(student)`. 

I like that `grades_for` is succinct, considering how much it is used internally in AssignmentType for calculations.

Outside of the file, I would expect `AssignmentType.grades_for(student) == AssignmentType.grades.where(student: student)`, but I'm not sure if are serious repercussions for code written under this assumption. If someone saw the method was available, and tried to use it to find a count for incomplete grades by assignment type, for instance.

If it is a method better used internally only, perhaps it should be private?

I'm not necessarily recommending a change, just trying to lay out my thoughts.

chcholman: I really appreciate all of this thinking. I am going to keep working with some of these issues in other work, so for the moment I'm going to hold off on making any further changes while I think about it some more.

PR #2164, opened 2016-06-25
Removed limitation on the upper bound of grade scheme element

chcholman:
This bugfix addresses the grade scheme element form rejecting point values that are higher than 1) is set on the Course model, or 2) the total sum of assignments created. Because the grading scheme is often entered far before all of the assignments are in, this is key to the UX feeling functional. It also adds the ability to edit the highest_points value, and for changes here to smartly update the lowest bounds of the above grade scheme element.

### DIFF
--- a/app/assets/javascripts/angular/directives/GradeSchemeDirectives.js.coffee
+++ b/app/assets/javascripts/angular/directives/GradeSchemeDirectives.js.coffee
@@ -7,12 +7,13 @@
   controller: ($scope) ->
     this.lowest_points = (modelValue, viewValue) ->
       if (modelValue < $scope.element.highest_points || $scope.element.highest_points == '')
-        GradeSchemeElementsService.update_scheme($scope.index, modelValue)
+        GradeSchemeElementsService.update_high_range($scope.index, modelValue)
         true
       else
         false
     this.highest_points = (modelValue, viewValue) ->
-      if (modelValue > $scope.element.lowest_points && modelValue < GradeSchemeElementsService.getTotalPoints())
+      if (modelValue > $scope.element.lowest_points || $scope.element.lowest_points == '' )
+        GradeSchemeElementsService.update_low_range($scope.index, modelValue)
         true
       else
         false
@@ -30,5 +31,4 @@
   require: ['^gradeSchemeRanges', '^ngModel']
   restrict: 'C'
   link: (scope, elm, attrs, ctrls) ->
-   ctrls[1].$validators.range = ctrls[0].highest_points
-    # create an on change binding
+    ctrls[1].$validators.range = ctrls[0].highest_points
--- a/app/assets/javascripts/angular/directives/GradeSchemeDirectives.js.coffee
+++ b/app/assets/javascripts/angular/directives/GradeSchemeDirectives.js.coffee
@@ -6,17 +6,11 @@
   }
   controller: ($scope) ->
     this.lowest_points = (modelValue) ->
-      if (modelValue < $scope.element.highest_points || $scope.element.highest_points == '')
-        GradeSchemeElementsService.update_high_range($scope.index, modelValue)
-        true
-      else
-        false
+      GradeSchemeElementsService.isLowestPointValue($scope.element, $scope.index, modelValue)
+
     this.highest_points = (modelValue) ->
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/angular/directives/GradeSchemeDirectives.js.coffee
```diff
@@ -7,12 +7,13 @@
   controller: ($scope) ->
     this.lowest_points = (modelValue, viewValue) ->
```
zazaian: It seems to me like the fact that we're calling the `GradeSchemeElementService` is being called in both the `lowest_points` method and the `highest_points` method means that both methods should be defined in the service itself rather than in the controller for this directive.

app/assets/javascripts/angular/directives/GradeSchemeDirectives.js.coffee
```diff
@@ -7,12 +7,13 @@
...
       if (modelValue < $scope.element.highest_points || $scope.element.highest_points == '')
-        GradeSchemeElementsService.update_scheme($scope.index, modelValue)
+        GradeSchemeElementsService.update_high_range($scope.index, modelValue)
         true
       else
         false
     this.highest_points = (modelValue, viewValue) ->
```
zazaian: Why do we need a `viewValue` here if it isn't being used?

app/assets/javascripts/angular/services/GradeSchemeElementsService.js.coffee
```diff
@@ -25,10 +25,14 @@
...
       (value < elements[parseInt(index)].highest_points)
 
-    update_scheme = (index, newValue) ->
+    update_high_range = (index, newValue) ->
       if(index != elements.length-1)
         elements[index+1].highest_points = newValue-1
 
+    update_low_range = (index, newValue) ->
+      if(index != elements.length+1)
+        elements[index-1].lowest_points = newValue+1
```
zazaian: Was a decision made to remove all spaces around math operators or is just for consistency? I feel like removing the spaces around the `+` and `-` is confusing and really hurts readability.

app/assets/javascripts/angular/templates/ng_gradeSchemeRanges.html.haml
```diff
@@ -1,4 +1,4 @@
 %span.form_label Low Point Threshold
 %input.low-range{:type => 'number','index' => '{{index}}','name' => 'lowest_points', 'ng-model' => 'element.lowest_points'}
 %span.form_label High Point Threshold
-%input.high-range{:type => 'number', 'name' => 'highest_points', 'ng-model' => 'element.highest_points', 'ng-disabled' => 'index > 0'}
+%input.high-range{:type => 'number', 'index' => '{{index}}','name' => 'highest_points', 'ng-model' => 'element.highest_points' }
```
zazaian: Yayy Angular :tada:

app/assets/javascripts/angular/services/GradeSchemeElementsService.js.coffee
```diff
@@ -56,7 +60,8 @@
         getGradeSchemeElements: getGradeSchemeElements
         postGradeSchemeElements: postGradeSchemeElements
         checkLowRange: checkLowRange
-        update_scheme: update_scheme
```
zazaian: :+1: the consistent naming here definitely helps clarify the relationship between these methods.

app/assets/javascripts/angular/directives/GradeSchemeDirectives.js.coffee
zazaian: Can you explain the naming `modelValue` and `viewValue` here? What do these represent?

app/assets/javascripts/angular/services/GradeSchemeElementsService.js.coffee
acousticrobot: Looks like I was the last to [touch these lines](https://github.com/UM-USElab/gradecraft-development/commit/43a9d86de62c197800e17c754ddb400178d228fb) but it just for a find and replace `low_points` to `lowest_points`.  The lack of spaces is inconsistent with the rest of the app and I agree w/ @zazaian that it is confusing and hard to read.

zazaian:
@chcholman I had a few questions about method organization and argument naming. Let met know if these are unclear at all.

PR #2168, opened 2016-06-27
1864 predictor style

acousticrobot:
NOT READY TO MERGE

No need to review the code for now, please just take a look at this "sketch" for the Predictor sidebar. I'd like to get some feedback before fully styling and implementing, and in advance of adding it to the other student pages.

Things I'd like to discuss tomorrow:

  * Can we fix the tabs to the top, and remove the H3.pagetitle (GRADE PREDICTOR) title on this and maybe all the other student pages? If the graph and sidebar did not need to scroll-to-stick I think it would improve the UX, not to mention simplify the js.
  * I put a working remove 'X' button in the sidebar, but I think we might be OK with just the "Okay, got it!"
  * I added an animated attention point instead of a red arrow, but am open to other ideas or refinements. My thoughts go to reducing the speed and opacity, but I'd like to see some other 'hint' cards first.
  * I've added all the icons into the side panel as requested. Please view `Group Assignment + Submissions` Assignment, and some of the Unlock Settings examples (both locked and key examples)
  * I've just added, unsettled, the name, description, due-date, and points.  Additional properties on the Assignment will be really easy to add now, which ones should we add?
  * Do we keep the due date in the predictor widgets?
  * If you have a chance to run the sample data, I've added an Assignment in the Predictor Type called "Assignment With Icons and Levels" -- that has all three 'main' icons. I'd like to work on the spacing of these, and also -- The icons in the example spec float right -- should I implement this? 
  * I have't built the sidebar detail page for the badges yet -- how similar can we make them here and on the badge page? I think it would be cool if these detail blocks could be completely autonomous and not reliant on the predictor context, but this could take some forethought.  
  * I haven't started to tackle any responsive implementation, but we need to, and this might be better discussed face-to-face tomorrow.

Thanks for any feedback!


mariehooper:
Hi @acousticrobot - @chcholman and I have addressed your Qs above as discussed... here is an itemized list of changes based on that feedback! 

- [ ] fix the tabs to the top, and remove the H3.pagetitle on all student pages
- [ ] remove the 'x' to close the side panel since the panel will not be opening and closing and just leave the 'okay got it!' as a way to dismiss that initial instruction block.
- [ ] make whole card headers link to info rather than just the info icon (and reduce the size of the info icon to ~ 0.85rem to match font-size
- [ ] yes, let's float the icons right... I think they stand out a little better that way.
- [ ] we need to standardize icons for different states.. lets make 'required' the 'fa-asterisk' icon, 'late' the 'fa-clock-o' icon, and 'group assignment' should be 'fa-users' everywhere - some places we use this and some we use the building blocks icon
- [ ] Let's leave the due date per Cait request but slide a 'fa-calendar' icon ahead of it and reduce the size of the font to 0.7rem

I will look into all the possibilities for assignment properties to add to the side panel, but for now let's have the details you mentioned with a truncated version of the assignment description that links out to the assignment show page. 

Thanks so much! This is looking awesome! 🎉  ✨

PR #2176, opened 2016-06-29
Add Ability to Copy Courses with Students

### LINKED ISSUE #1607: Copy Course & students
Particularly in the K-12 context we have spaces where instructors need to copy a course AND bring the students along with them. We should support a workflow that allows for this within the Course Copy feature, allowing the instructor mark to "Copy students & roles", and thereby creating new course memberships that match the roles of the course being copied.

jonathoy:
Skips After Create Callback for Courses Model When Copying With Associations
to Prevent Duplication Issues

Closes #1607


spec/controllers/courses_controller_spec.rb
```diff
@@ -154,6 +154,21 @@
       end
     end
 
+    describe "POST copy_with_students" do
+      let(:course_with_students) { create(:student_course_membership, user: student, course: @course).course }
+      let(:student) { create :user }
```
acousticrobot: You don't have to declare the student here if you aren't going to use it in the expect block. Also, it works (because of the magic of let, I think), but it's confusing have `let(:student)` on 159, but to use the variable on line 158.

I would change these two lines to:
```
let(:course_with_students) { create(:student_course_membership, course: @course).course }
```
and the `student_course_membership` factory will create the user.

spec/models/course_spec.rb
```diff
@@ -49,7 +49,7 @@
   end
 
   describe "#copy" do
-    let(:course) { build :course }
+    let(:course) { create :course }
     subject { course.copy }
```
acousticrobot: This is much cleaner!

spec/models/course_spec.rb
```diff
@@ -83,20 +80,38 @@
...
+      subject.copy_with_students
+      expect(Course._create_callbacks.map(&:filter)).to include :create_admin_memberships
+    end
+
+    it "copies the students" do
+      duplicated = subject.copy_with_students
+      expect(duplicated.reload.users).to include student
+    end
+  end
+
```
acousticrobot: This may be a personal style preference, but agian I find the `subject` declared on line 97 but used on line 96 confusing. 

Additionally in this case I would probably not assign the original course to `subject` because I feel like it is only tangentially the scrutinized subject. I would expect the subject to be `duplicated`, or `callbacks`, if anything.

I probably would have written this block as:

```
  describe "#copy_with_students" , focus: true do
    let!(:course) { create(:student_course_membership, course: subject).course }

    it "turns off the create_admin_memberships callback" do
      expect_any_instance_of(Course).to_not receive(:create_admin_memberships)
      course.copy_with_students
    end

    it "turns back on the create_admin_memberships callback" do
      course.copy_with_students
      expect(Course._create_callbacks.map(&:filter)).to include :create_admin_memberships
    end

    it "copies the students" do
      duplicated = course.copy_with_students
      expect(duplicated.reload.users).to include course.students.first
    end
  end
```

app/controllers/courses_controller.rb
```diff
@@ -3,6 +3,7 @@ class CoursesController < ApplicationController
 
   before_filter :ensure_staff?
 
+  # rubocop:disable AndOr
```
jwright: Why do we need to disable this `rubocop` rule here?

app/controllers/courses_controller.rb
```diff
@@ -31,19 +32,15 @@ def edit
...
-    else
-      redirect_to courses_path,
-        alert: "#{course.name} was not successfully copied"
-    end
+    @course = Course.find(params[:id])
+    duplicated = @course.copy
+    validate_copy(duplicated)
+  end
+
+  def copy_with_students
```
jwright: I feel like we should not have 2 action endpoints here. They both seem like they are accomplishing the same thing, one just has the option to copy students. I think a parameter for the option could be passed in. This would eliminate the need for the private `validate_copy` method and the small duplication of finding the course.

app/models/course.rb
```diff
@@ -161,14 +161,14 @@ def user_term
...
   def copy(attributes={})
-    ModelCopier.new(self).copy(attributes: attributes,
-                               associations: [
-                                 :badges,
-                                 { assignment_types: { course_id: :id }},
-                                 :challenges,
-                                 :grade_scheme_elements
-                               ],
-                               options: { prepend: { name: "Copy of " }})
+    copy_with_associations(attributes, [])
```
jwright: 👍

app/models/course_membership.rb
```diff
@@ -24,6 +25,10 @@ class CourseMembership < ActiveRecord::Base
 
   validates :instructor_of_record, instructor_of_record: true
 
+  def copy(attributes={})
```
jwright: This method definition is not necessary since it will just use the one from `Copyable`.

spec/models/course_spec.rb
jwright: 👍

spec/models/course_spec.rb
jwright: @acousticrobot I disagree with this since this is a `Course` spec. The subject under test should be a `course`. 

I could see it being the `duplicated` course except it's the result of the `"when"` action and so I don't think that should be the subject under test. This would result in the following:

```
describe "#copy_with_students" do
  let(:course) { create :course }
  let!(:student) { create(:student_course_membership, course: course).user }
  subject { create(:course).copy_with_students }

  it "turns off the create_admin_memberships callback" do
    expect_any_instance_of(Course).to_not receive(:create_admin_memberships)
    subject # <--- weird
  end
end
```

jwright:
@jonathoy Looks good man! I had a few comments but nothing major. Nice work 👌

spec/models/course_spec.rb
acousticrobot: I hear you that that is very strange. I wouldn't advocate for changing what `subject` refers to, but getting rid of it entirely. I guess what's bothering me is this is an situation where an instance method acts like a class method (constructor) , so `subject` doesn't seem to correlate to the instance. And it breaks with the pattern from the test directly above:

```
describe "#copy" do
    let(:course) { create :course }
    subject { course.copy }

...

describe "#copy_with_students" do
    let!(:student) { create(:student_course_membership, course: subject).user }
    subject { create :course }
```

so it necessitates a lot of extra scrolling to keep track of the context of `subject` and makes me question if it's good enough that `subject` is *any* instance of `course`. This may be a larger discussion though about the merits of requiring the use of a `subject` in all tests.

app/controllers/courses_controller.rb
jonathoy: Rubocop was having a fit on lines 123 and 126 where I had used the `and return` statement. I originally had it just above line 123 but I realized that if you put `# rubocop:disable AndOr` once it will suppress for the entire file. I had figured by that logic it made more sense to have it at the top of the class.

spec/models/course_spec.rb
jonathoy: Being an RSpec newb, I would be interested in hearing said discussion.

acousticrobot: I'll add it to Tuesday chats, but feel free to submit the PR with these tests as-is in the meantime.

Issue #2167, opened 2016-06-27
Perform course analytics export with background job


zazaian:
We're getting a cryptic error-free timeout for generating large course analytics export archives in production. The logic in the `AnalyticsController#export` method still needs to be refactored, but perhaps at least we can move this into a job for now, or condense it down into the necessary logic for a simple `CourseAnalyticsExport` class without getting too bogged down in the development of that class.

zazaian:
This was closed by PR #2250

Issue #2118, opened 2016-06-20
AnalyticsController#export needs to be refactored


zazaian:
The `AnalyticsController#export` method is doing way too much heavy lifting for a controller action at present. We've basically put all of the system analytics export logic here, when that information should be implemented through a separate class, the sole purpose of which would be to orchestrate which individual Event/Mongo record export classes to include.

This is currently blocking the work that I'm doing on extracting the logic from `Analytics::Export` into multiple classes and modules as all of the export logic is being jammed into a single CSV file when each Export class should have a dedicated CSV file.

I've gone through and commented out the entire workflow in Analytics#export to add some clarity to what's happening here and how the data's being used, but all of this should ultimately be condensed down into a class or two that functions expressly to orchestrate this export process:

```ruby
# /app/controllers/analytics_controller.rb

class AnalyticsController
  def export
    respond_to do |format|
      format.zip do
        # create a temporary directory in which to generate all of the export
        # files
        export_dir = Dir.mktmpdir

        # create a new zip file using ApplicationController#export_zip
        export_zip "#{ current_course.courseno }_anayltics_export_#{ Time.now.strftime('%Y-%m-%d') }", export_dir do

          # iterate over each role and perform the export process for each role
          (Role.all + ["total"]).each do |role|

            # create a sub-directory just for the current role and store
            # exported files for that role in that sub-directory
            role_subdir = File.join(export_dir,role.pluralize)

            # let's remember that id is just the id for the current_course
            course_id = current_course.id

            # it seems evident to me immediately that we're searching for
            # all events of each type every time we iterate to the next role.
            #
            # additionally we're fetching all Analytics::Event records as
            # well as fetching predictor-specific event records, which means that
            # predictor analytics records are being fetched a number of times
            # equal to this equation = 2 x (total_number_of_roles + 1)
            #
            # when we clean this up it will certainly become more performant
            # as we'll drastically cut down on the number of export queries.

            # fetch all analytics events for the current course
            events = Analytics::Event.where course_id: course_id

            # fetch all predictor events for the current course
            predictor_events = Analytics::Event.where \
              course_id: current_course.id,
              event_type: "predictor"

            # fetch all CourseUserPageview analytics aggregates for all time
            # for the scope of the the given course
            ## this is the range of time across which to fetch data
            user_pageviews = CourseUserPageview.data :all_time, \
              nil,
              { course_id: course_id },
              { page: "_all" }

[...]

PR #2186, opened 2016-07-01
2185 - Format Course Numbers for Analytics Exports

zazaian:
The entire point of this PR is that we were using unformatted `Course#courseno` values to construct the filenames for `Analytics::Export` archives, but some of those `courseno` values have forward slashes and other characters that were breaking export filenames.

I've added a `#url_safe` method to the `Formatter::Filename` library as well to give us url-safe and therefore filename-safe output for strings, and then implemented that in the current process for performing analytics exports in `AnalyticsController#export`.

### DIFF
--- a/app/controllers/analytics_controller.rb
+++ b/app/controllers/analytics_controller.rb
@@ -204,8 +204,10 @@ class AnalyticsController < ApplicationController
         export_tmpdir = Dir.mktmpdir nil, s3fs_prefix
 
         # create a url-safe course number for the export's root directory
-        course_number = Formatter::Filename.new(current_course.courseno)
-                          .url_safe.filename
+        # be sure to replace forward-slashes with hyphens
+        course_number = Formatter::Filename.new(
+                          current_course.courseno.gsub(/\/+/,"-")
+                        ).url_safe.filename
 
         # create a named directory to generate the files in
         export_dir = FileUtils.mkdir \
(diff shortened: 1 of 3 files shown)


lib/formatter/filename.rb
```diff
@@ -20,6 +20,17 @@ def sanitize
       self
     end
 
+    # just remove erroneous characters, strip out whitespace, and replace
```
jwright: 👍 for the commenting here

app/controllers/analytics_controller.rb
```diff
@@ -203,8 +203,15 @@ def export
         # create a working tmpdir for the export
         export_tmpdir = Dir.mktmpdir nil, s3fs_prefix
 
+        # create a url-safe course number for the export's root directory
```
jwright: I wonder instead of putting this directly in the controller, we create a `CourseNumberFilenameFormatter` or something like that. It would help in testing and would not pollute the controller logic. Thoughts?

jwright:
@zazaian Looks great man. I just had a suggestion.

app/controllers/analytics_controller.rb
zazaian: So this whole process is being carried out in 4 different branches right now. The work to extract the code from the `AnalyticsController` is being done in the #2118 branch. It actually removes the `AnalyticsController` from the equation entirely and does the routing through an `CourseAnalyticsExportsController`.

If you'd like to keep track of everything the branch for extracting the process into a job is being done in #2167, and the branch for refactoring `Analytics::Export` is #2132.

zazaian: This branch is basically a bugfix so we can produce exports for production courses with non-standard characters in their `courseno` values.

zazaian:
Awesome, thanks for taking a look :eyeglasses:  @chcholman  this branch is @jwright-approved and ready for merge.

PR #2193, opened 2016-07-06
Sample updates

chcholman:
This PR adds Course Events, Flagged Students, Assignment Descriptions, Rubric Criterion & Level Descriptions, Team Leaderships, and an Assignment that is past with no grade but with a prediction to the sample data. 

It also fixes a big where Assignment Descriptions weren't being shown, and pleases the Rubocop gods with 3 removed !! items.

### DIFF
--- a/db/samples/assignments.rb
+++ b/db/samples/assignments.rb
@@ -31,9 +31,9 @@
     accepts_submissions_until: nil,
     resubmissions_allowed: false,
     accepts_text: false,
-    description: nil,
+    description: "Basilisk venom Umbridge swiveling blue eye Levicorpus, nitwit blubber oddment tweak. Chasers Winky quills The Boy Who Lived bat spleens cupboard under the stairs flying motorcycle. Sirius Black Holyhead Harpies, you’ve got dirt on your nose. Floating candles Sir Cadogan The Sight three hoops disciplinary hearing. Grindlewald pig’s tail Sorcerer's Stone biting teacup. Side-along dragon-scale suits Filch 20 points, Mr. Potter.",
     grade_scope: "Individual",
-    hide_analytics: nil,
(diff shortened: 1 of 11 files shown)


db/samples/assignments.rb
```diff
@@ -595,6 +595,26 @@
...
 }
 
+@assignments[:predictor_past_assignment_with_prediction] = {
+  quotes: {
+    assignment_created: nil,
+  },
+  assignment_type: :predictor,
+  attributes: {
+    name: "Past Assignment no Grade but Prediction",
+    description: "Displays a Slider. Displays the info icon with this text \
```
acousticrobot: I think you want this to say "Displays 0 / total points". We can skip the info icon message -- I will update the rest of the sample descriptions on my branch since we are removing hover states.

If you did actually want to show a slider with a prediction for a passed grade, we will have to update the logic in the predictor.

db/samples/courses.rb
```diff
@@ -322,3 +322,26 @@
     twitter_hashtag: "si101",
   }
 }
+
+@courses[:no_dates_no_info] = {
+  quotes: {
+    course_created: "“To live is the rarest thing in the world. Most people \
+    exist, that is all.” ― Oscar Wilde",
+    grade_sceme_elements_created: " ",
```
acousticrobot: Update to  `grade_scheme_elements_created` here too, and on the four other courses in this file.

db/samples/assignments.rb
chcholman: Great catch, thanks! I definitely meant to change the description.

db/samples/courses.rb
chcholman: Thank you!

Issue #2171, opened 2016-06-28
Import Canvas assignment


ksofa2:
Depends on #2080 and #2170

jwright:
** Import

- [ ] `description` -> `description`
- [ ] `due_at` -> `due_at`
- [ ] `points_possible` -> `possible_points`
- [ ] `grading_type` -> `pass_fail` (only if grading_type is `pass_fail`

jwright:
-  [ ] Only import visible assignments

Issue #2170, opened 2016-06-28
Authenticate Canvas with OAuth


ksofa2:
Depends on #2080

PR #2194, opened 2016-07-06
2080 list canvas assignments

### LINKED ISSUE #2080: List Canvas assignments
Start with a one-time import of assignments from a Canvas course. Link to this from the course_creation_wizard list.

jwright:
Lists assignments from the [Canvas API](https://www.canvaslms.com) based on an access token provided in an environment variable.

The ability to OAuth for multiple users will be implemented in the next story #2170 and the ability to import these assignments from Canvas will be implemented in a following story #2171.

The flow is as follows:

A staff member visits `/courses` and there is now an "Import Courses" on the button bar.

screen shot 2016-07-06 at 4 08 11 pm

Once on the `/courses/import`, the user can select the provider. For now, we only provide Canvas.

screen shot 2016-07-06 at 4 09 45 pm

Once on the `/courses/import/canvas` page, the user can view a list of courses they have access to in Canvas.

screen shot 2016-07-06 at 4 11 15 pm

Once the user selects "Select Assignments to Import", they will view a list of assignments available for the selected course.

screen shot 2016-07-06 at 4 11 47 pm

To get this to work for your specific canvas install, you can visit http://umich.instructure.com, sign in, and create a temporary access token. You can then update the environment variable `ENV["CANVAS_ACCESS_TOKEN"]`.

Closes #2080


config/routes.rb
```diff
@@ -160,7 +160,14 @@
...
   resources :courses do
-    post :copy, on: :collection
+    collection do
+      post :copy
+      resource :import, only: [], controller: :import, module: :courses do
+        get :providers
+      end
+      get "/:provider/import" => "courses/import#courses", as: :provider_import
+      get "/:provider/import/:id/assignments" => "courses/import#assignments", as: :provider_import_assignments
+    end
```
acousticrobot: You've spent more time thinking about the routing of this than I have, but I'm just wondering out loud,

1) Is it strange at all to have both of these routes?
```
/courses/import/providers                                  
/courses/:provider/import
```

2) Will there ever be a need for this route?

```
courses/:course_id/import
```

jwright: 1. I don't think it's weird because `/courses/import/providers` lists the available providers and `/courses/:provider/import` is for one provider. I am open to changing it if you have a better suggestion. Does starting with `/import` make more sense? So `/import/providers` and `/import/:provider/courses`?

2. No because that is handled via `/courses/:provider/import/:id/blah` where `blah` is `assignments` but it can be `grades` in the future.

lib/lms_importer/course_importer.rb
```diff
@@ -0,0 +1,29 @@
...
+require_relative "course_importer/canvas_course_importer"
+
+module LMSImporter
+  class CourseImporter
+    include ActiveSupport::Inflector
+
+    attr_reader :provider
+
+    def initialize(provider, access_token)
+      klass = constantize("LMSImporter::#{classify("#{provider}CourseImporter")}")
```
acousticrobot: Works fine for :canvas, but will the provider ever be in snake_case?

```
[1]> provider = :ham_and_eggs
=> :ham_and_eggs
[2] > "LMSImporter::#{classify("#{provider}CourseImporter")}"
=> "LMSImporter::HamAndEggscourseimporter"
```

config/routes.rb
```diff
@@ -162,6 +162,11 @@
   resources :courses do
     collection do
       post :copy
+      resource :import, only: [], controller: :import, module: :courses do
+        get :providers
+      end
+      get "/:provider/import" => "courses/import#courses", as: :provider_import
+      get "/:provider/import/:id/assignments" => "courses/import#assignments", as: :provider_import_assignments
```
acousticrobot: You've thought about the routing more than I have at this point, but at first glance I'm wondering 

1) if it's strange to switch placement of `import` in these  two routes?
```
/courses/import/providers
/courses/:provider/import
```
2) would we ever need this route?
```
/courses/:id/import
```
3) Is this an unusual pattern, to have the `courses/` and course `:id` split by `/:provider/import/`? 

```
/courses/:provider/import/:id/assignments
```

lib/lms_importer/course_importer.rb
jwright: Great catch

jwright:
@acousticrobot What do you think about changing the routes to use `/import` as the verb first and then what is being imported? It's weird because we have `/users/import` already which is very restful. Using `/import` first is not a resource but a verb instead. It would eliminate your concerns above.

jwright:
@acousticrobot I really like `importers` instead of `import`. I will change it then.

jwright:
@acousticrobot There is an `/exports` route. Maybe `/imports`? (I like `/importers` better but I would like to make it consistent)

jwright:
👍

jwright:
@acousticrobot I sent this back to you for a quick review. The routes are now:

`/importers` -> Lists the available providers
`/importers/:importer_id/courses` -> Lists the available courses from the specific provider (`importer_id`)
`/importers/:importer_id/courses/:id/assignments` -> Lists the available assignments from the specific provider (`importer_id`) and course

jwright:
@chcholman This is ready now. I commented out the import course button.

PR #2202, opened 2016-07-11
Cleaned up badges table

chcholman:


spec/models/badge_spec.rb
```diff
@@ -19,6 +19,42 @@
       expect(subject).to_not be_valid
       expect(subject.errors[:course]).to include "can't be blank"
     end
+
+    it "is invalid without a visible state" do
```
jwright: I wonder if this is a good validation message. Will the user be able to decipher `Badge visible is not included in the list`?

I know that this does not appear on the UI but it may. Maybe a better message would be `Badge visible must be true or false`? We can easily change this via:

```
# badge.rb

validates_inclusion_of :visible, :can_earn_multiple_times, :visible_when_locked,
  :show_name_when_locked, :show_points_when_locked, :show_description_when_locked,
  in: [true, false], message: "must be true or false"
```

jwright:
@chcholman Looks great! I just had a question for you.

spec/models/badge_spec.rb
chcholman: That's awesome - I love the improved message.

PR #2205, opened 2016-07-12
2171 import canvas assignment

### LINKED ISSUE #2171: Import Canvas assignment

jwright:
Import the assignments from Canvas into GradeCraft.

From a professor's point of view, this is how they will be importing an assignment:

1. Visit `/importers` and select "Canvas"

screen shot 2016-07-12 at 4 18 37 pm

2. Select the course from Canvas which holds the assignment or assignments to import

screen shot 2016-07-12 at 4 18 47 pm

3. Select the assignment or assignments that are to be imported and select the `AssignmentType`

screen shot 2016-07-12 at 4 19 01 pm

4. View the results of the import

screen shot 2016-07-12 at 4 19 14 pm

When importing an assignment from Canvas, the following fields are imported into a new `Assignment` object:

* Canvas Assignment `name` -> GradeCraft Assignment `name`
* Canvas Assignment `description` -> GradeCraft Assignment `description`
* Canvas Assignment `due_at` -> GradeCraft Assignment `due_at`
* Canvas Assignment `points_possible` -> GradeCraft Assignment `full_points`
* Canvas Assignment `grading_type` == `pass_fail` -> GradeCraft Assignment `pass_fail`

The provider's name and assignment id are stored in a new table called `ImportedAssignments`. This will be useful when we need to sync assignments. I added this now in case professors start using this in production, we will be able to sync them in the future when we implement that feature.

Closes #2171

### DIFF
--- a/app/importers/assignment_importers/canvas_assignment_importer.rb
+++ b/app/importers/assignment_importers/canvas_assignment_importer.rb
@@ -19,6 +19,7 @@ class CanvasAssignmentImporter
         assignment.pass_fail = true if canvas_assignment["grading_type"] == "pass_fail"
 
         if assignment.save
+          link_imported canvas_assignment["id"], assignment
           successful << assignment
         else
           unsuccessful << { data: canvas_assignment,
@@ -28,5 +29,14 @@ class CanvasAssignmentImporter
     end
 
     self
+  end
+
+  private
+
+  def link_imported(provider_id, assignment)
+    imported = ImportedAssignment.find_or_initialize_by(provider: :canvas,
+                                                        provider_id: provider_id)
+    imported.assignment = assignment
+    imported.save
   end
 end
--- a/app/importers/assignment_importers/canvas_assignment_importer.rb
+++ b/app/importers/assignment_importers/canvas_assignment_importer.rb
@@ -33,9 +33,9 @@ class CanvasAssignmentImporter
 
   private
 
-  def link_imported(provider_id, assignment)
+  def link_imported(provider_resource_id, assignment)
     imported = ImportedAssignment.find_or_initialize_by(provider: :canvas,
-                                                        provider_id: provider_id)
+      provider_resource_id: provider_resource_id)
     imported.assignment = assignment
     imported.save
   end
--- a/app/importers/assignment_importers/canvas_assignment_importer.rb
+++ b/app/importers/assignment_importers/canvas_assignment_importer.rb
@@ -12,7 +12,12 @@ class CanvasAssignmentImporter
     unless assignments.nil?
       assignments.each do |canvas_assignment|
         assignment = course.assignments.build name: canvas_assignment["name"],
+          description: canvas_assignment["description"],
+          due_at: canvas_assignment["due_at"],
+          full_points: canvas_assignment["points_possible"],
           assignment_type_id: assignment_type_id
+        assignment.pass_fail = true if canvas_assignment["grading_type"] == "pass_fail"
+
         if assignment.save
           successful << assignment
         else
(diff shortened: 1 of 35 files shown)


app/importers/assignment_importers/canvas_assignment_importer.rb
```diff
@@ -0,0 +1,42 @@
...
+        end
+      end
+    end
+
+    self
+  end
+
+  private
+
+  def link_imported(provider_id, assignment)
```
acousticrobot: Nice way to keep track of Assignment origins! How come this is a `has_many` relationship on assignment?

db/migrate/20160712195437_create_imported_assignments.rb
```diff
@@ -0,0 +1,11 @@
+class CreateImportedAssignments < ActiveRecord::Migration
+  def change
+    create_table :imported_assignments do |t|
+      t.references :assignment, index: true, foreign_key: true
+      t.string :provider, null: false
+      t.string :provider_id, null: false
```
acousticrobot: `:provider_id` made me think this was and id associated with the provider and not the provider's assignment id.
I also don't know if we might actually have a provider model in the future, which would require this namespace for associations. 

Maybe we want to change this to `provider_assignment_id` or something more explicit? 

Alternatively,  you could call it `foreign_id` if you are planning on having a series of these models and want to standardize a single field name for all of them:  `ImportedAssignments.foreign_id`, `ImportedCourses.foreign_id`, etc.

Thoughts?

lib/active_lms/invalid_provider_error.rb
```diff
@@ -1,4 +1,4 @@
-module LMSImporter
+module ActiveLMS
```
acousticrobot: 🌴 😎

lib/active_lms/syllabus.rb
```diff
@@ -1,14 +1,14 @@
 require "active_support"
-require_relative "course_importer/canvas_course_importer"
+require_relative "syllabus/canvas_syllabus"
 
-module LMSImporter
-  class CourseImporter
+module ActiveLMS
+  class Syllabus
```
acousticrobot: 👍

app/importers/assignment_importers/canvas_assignment_importer.rb
jwright: @acousticrobot It should be a one-to-one relationship. One imported assignment (provider and imported id) per one assignment. I do a `find_or_initialize_by` here just in case. I wonder if we should just always create a new `ImportedAssignment` though. Thoughts?

db/migrate/20160712195437_create_imported_assignments.rb
jwright: I like `provider_assignment_id`. I will change it.

jwright:
@acousticrobot No. Those are one of those sporadic failures.

db/migrate/20160712195437_create_imported_assignments.rb
jwright: Actually, I think I like `provider_resource_id`. Objections?

acousticrobot: That sounds great to me!

app/models/assignment.rb
```diff
@@ -48,6 +48,7 @@ class Assignment < ActiveRecord::Base
   has_many :submissions, dependent: :destroy
 
   has_many :criterion_grades, dependent: :destroy
+  has_many :imported_assignments, dependent: :destroy
```
acousticrobot: I think `find_or_initialize_by` is fine in `link_imported`. This is the `has_many` I was questioning.

jwright: @acousticrobot  This is because an assignment may have multiple providers but it will probably only have one. I could change this to a `has_one` if you think that makes more sense. I don't think the above scenario would happen. What do you think?

acousticrobot: I think we should keep it as a `has_one` for now for clarity of intent, until we handle a situation where we are syncing between multiple providers.

jwright: 👍

PR #2223, opened 2016-07-16
Removed bullets for list items in assignment form, added descriptor t…

chcholman:
This PR: 

* fixes the styling of the list items in the unlock conditions assignment section
* Adds Pass/Fail as an unlock condition
* Adds course-level point earnings as an unlock condition

### DIFF
--- a/app/models/course_membership.rb
+++ b/app/models/course_membership.rb
@@ -25,6 +25,8 @@ class CourseMembership < ActiveRecord::Base
 
   validates :instructor_of_record, instructor_of_record: true
 
+  after_save :check_unlockables
+
   def assign_role_from_lti(auth_hash)
     return unless auth_hash["extra"] && auth_hash["extra"]["raw_info"] && auth_hash["extra"]["raw_info"]["roles"]
 
@@ -54,6 +56,15 @@ class CourseMembership < ActiveRecord::Base
 
   def staff?
     professor? || gsi? || admin?
+  end
+
+  def check_unlockables
+    if self.course.is_a_condition?
+      unlock_conditions = UnlockCondition.where(condition: self.course).each do |condition|
+        unlockable = condition.unlockable
+        unlockable.check_unlock_status(user)
+      end
+    end
   end
 
   private
--- a/app/models/course_membership.rb
+++ b/app/models/course_membership.rb
@@ -60,8 +60,7 @@ class CourseMembership < ActiveRecord::Base
 
   def check_unlockables
     if self.course.is_a_condition?
-      unlock_conditions = UnlockCondition.where(condition: self.course).each do |condition|
-        unlockable = condition.unlockable
+      self.course.unlock_conditions.map(&:unlockable).each do |unlockable|
         unlockable.check_unlock_status(user)
       end
     end
--- a/app/models/course_membership.rb
+++ b/app/models/course_membership.rb
@@ -60,7 +60,7 @@ class CourseMembership < ActiveRecord::Base
 
   def check_unlockables
     if self.course.is_a_condition?
-      self.course.unlock_conditions.map(&:unlockable).each do |unlockable|
+      self.course.unlock_keys.map(&:unlockable).each do |unlockable|
         unlockable.check_unlock_status(user)
       end
     end
--- a/app/models/unlock_condition.rb
+++ b/app/models/unlock_condition.rb
@@ -122,15 +122,12 @@ class UnlockCondition < ActiveRecord::Base
 
   def check_if_badge_earned_enough_times(student)
     badge_count = student.earned_badges_for_badge_count(condition_id)
-    return false unless badge_count >= condition_value
-    return true
+    badge_count >= condition_value
   end
 
   def check_if_badge_earned_enough_times_by_date(student)
     badge_count = student.earned_badges_for_badge_count(condition_id)
-    return false unless badge_count >= condition_value &&
-      student.earned_badges.where(badge_id: condition_id).last.created_at < condition_date
-    return true
+    badge_count >= condition_value && student.earned_badges.where(badge_id: condition_id).last.created_at < condition_date
   end
 
   def check_assignment_condition(student)
@@ -184,14 +181,11 @@ class UnlockCondition < ActiveRecord::Base
 
   def check_passed_condition(student)
     grade = student.grade_for_assignment_id(condition_id).first
-    return true if grade.pass_fail_status == "Pass"
-    return false
+    grade.pass_fail_status == "Pass"
   end
 
   def check_if_grade_earned_meets_condition_value(grade)
-    return false unless GradeProctor.new(grade).viewable? &&
-        grade.score >= condition_value
-    return true
+    GradeProctor.new(grade).viewable? && grade.score >= condition_value
   end
 
   def check_if_grade_earned_met_condition_date(grade)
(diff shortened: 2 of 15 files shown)


app/models/course_membership.rb
```diff
@@ -25,6 +25,8 @@ class CourseMembership < ActiveRecord::Base
 
   validates :instructor_of_record, instructor_of_record: true
 
+  after_save :check_unlockables
```
jwright: Any way we can remove this callback and move it to the controller or a service to perform this? I am not a fan of callbacks unless it will be a lot of work to separate this from the model.

app/models/course_membership.rb
```diff
@@ -56,6 +58,15 @@ def staff?
     professor? || gsi? || admin?
   end
 
+  def check_unlockables
+    if self.course.is_a_condition?
+      unlock_conditions = UnlockCondition.where(condition: self.course).each do |condition|
+        unlockable = condition.unlockable
+        unlockable.check_unlock_status(user)
+      end
+    end
```
jwright: This block seems to be doing more than it needs to (I think). For example, we are assigning `unlock_conditions` to an enumerator but I don't see where it is being used. In addition, I think we can clean this up a bit. (This is pseudo-code)

```
if self.course.is_a_condition?
  self.course.unlock_conditions.map(&:unlockable).each do |unlockable|
    unlockable.check_unlock_status(user)
  end
end
```

app/models/unlock_condition.rb
```diff
@@ -24,6 +24,8 @@ def is_complete?(student)
       check_badge_condition(student)
     elsif condition_type == "Assignment"
       check_assignment_condition(student)
+    elsif condition_type == "Course"
+      check_course_membership_condition(student)
```
jwright: I think this `if` block is becoming a bit unmanagable and I think we can either use a factory pattern or change this to something like:

```
method = "check_#{condition_type.underscore}_condition"
self.send method, student if self.respond_to? method
```

app/models/unlock_condition.rb
```diff
@@ -33,32 +35,33 @@ def is_complete_for_group?(group)
...
+    return false unless group.present?
+    group.students.each do |student|
+      if self.is_complete?(student)
+        unlocked_count += 1
       end
-      return unlocked_count
-    else
-      return false
     end
+    return unlocked_count
```
jwright: We are returning two different types from this model (`false` and the count) which may confuse the calling method. Should we return a 0 or a `nil` here instead?

For calculating the unlock count, this may be more expressive code:
```
unlocked_count += group.students.count { |student| self.is_complete?(student) }
```

app/models/unlock_condition.rb
```diff
@@ -96,48 +103,43 @@ def condition_state_past
...
 
   def check_if_badge_earned_enough_times(student)
     badge_count = student.earned_badges_for_badge_count(condition_id)
-    if badge_count >= condition_value
-      return true
-    else
-      return false
-    end
+    return false unless badge_count >= condition_value
+    return true
```
jwright: This is a little hard to read but I think this will provide the same result: `badge_count >= condition_value`.

app/models/unlock_condition.rb
```diff
@@ -96,48 +103,43 @@ def condition_state_past
...
   def check_if_badge_earned_enough_times_by_date(student)
     badge_count = student.earned_badges_for_badge_count(condition_id)
-    if badge_count >= condition_value &&
+    return false unless badge_count >= condition_value &&
       student.earned_badges.where(badge_id: condition_id).last.created_at < condition_date
-      return true
-    else
-      return false
-    end
+    return true
```
jwright: Same thing here. I think this can be reduced to `badge_count >= condition_value && student.earned_badges.where(badge_id: condition_id).last.created_at < condition_date`

app/models/unlock_condition.rb
```diff
@@ -96,48 +103,43 @@ def condition_state_past
...
 
   def check_assignment_condition(student)
     if condition_state == "Submitted"
       check_submission_condition(student)
     elsif condition_state == "Grade Earned"
       check_grade_earned_condition(student)
+    elsif condition_state == "Passed"
+      check_passed_condition(student)
     elsif condition_state == "Feedback Read"
       check_feedback_read_condition(student)
```
jwright: You could implement the same pattern that I mentioned above by creating the method name off of the `condition_state`.

app/models/unlock_condition.rb
```diff
@@ -186,13 +182,16 @@ def check_grade_earned_condition(student)
     end
   end
 
+  def check_passed_condition(student)
+    grade = student.grade_for_assignment_id(condition_id).first
+    return true if grade.pass_fail_status == "Pass"
+    return false
+  end
```
jwright: This can be simplified with `grade.pass_fail_status == "Pass"`

app/models/unlock_condition.rb
```diff
@@ -186,13 +182,16 @@ def check_grade_earned_condition(student)
...
   def check_if_grade_earned_meets_condition_value(grade)
-    if GradeProctor.new(grade).viewable? &&
+    return false unless GradeProctor.new(grade).viewable? &&
         grade.score >= condition_value
-      return true
-    else
-      return false
-    end
+    return true
   end
```
jwright: This can be simplified with `GradeProctor.new(grade).viewable? && grade.score >= condition_value`

app/models/unlock_condition.rb
```diff
@@ -202,30 +201,24 @@ def check_if_grade_earned_met_condition_date(grade)
...
-        return true
-      end
-    else
-      return false
-    end
+    return false unless grade.present? && grade.feedback_read?
+    return true unless condition_date?
+    return false unless grade.feedback_read_at < condition_date
+    return true
+  end
```
jwright: Can be simplified with `grade.feedback_read_at < condition_date`

app/models/unlock_condition.rb
```diff
@@ -202,30 +201,24 @@ def check_if_grade_earned_met_condition_date(grade)
...
+
+  def check_course_membership_condition(student)
+    course_membership = student.course_memberships.where(course_id: condition_id).first
+    return false unless course_membership.score >= condition_value
+    return true
```
jwright: Can be simplified with `course_membership.score >= condition_value`

app/models/unlock_condition.rb
```diff
@@ -202,30 +201,24 @@ def check_if_grade_earned_met_condition_date(grade)
...
   # Checking if the number of students who have completed the condition match
   # the size of the group, returning true if so.
   def check_condition_for_each_student(group)
     unlocked_count = count_unlocked_in_group(group)
-    if unlocked_count == group.students.count
-      return true
-    else
-      return false
-    end
+    return false unless unlocked_count == group.students.count
```
jwright: Can be simplified with `unlocked_count == group.students.count`

jwright:
@chcholman This looks like some great additions to the functionality!

I added some comments where the code could be more readable with some minor improvements. I think the sentence creation code that is part of the `UnlockCondition` could be pulled out eventually since it's not really part of the model for now but I think that may be too early of an optimization. I also do not see code coverage for some of the code that was added.

app/models/course_membership.rb
chcholman: Awesome, thank you.

app/models/unlock_condition.rb
chcholman: 👍

app/models/unlock_condition.rb
chcholman: 👍

app/models/unlock_condition.rb
chcholman: 👍

app/models/unlock_condition.rb
chcholman: 👍

app/models/unlock_condition.rb
chcholman: 👍

app/models/unlock_condition.rb
chcholman: 👍

app/models/unlock_condition.rb
chcholman: Fancy!

app/models/course_membership.rb
chcholman: 👀

app/models/unlock_condition.rb
chcholman: Fixed!

app/models/unlock_condition.rb
chcholman: 👍

Issue #2217, opened 2016-07-15
Can't add first Grade Scheme Elements item


chcholman:
When you try to add the first item to the grade scheme elements, it fails the validations and won't add.

chcholman:
- The first level added is assumed to be the ‘high point’, and both values need to be settable without validations against any other element (since they don’t exist)
- All additional levels should ‘respond’ to values in the adjacent levels, adjusting their high point in response to the above low point being altered, and reversed for low point and blow
- The level high and low point should be separated by at least 1 point, with the high value being higher than the low value
- Goal: we want to ensure that they add a level that starts with 0 and covers the full spectrum

zazaian:
@chcholman is there a default value we want to use for the 'max' value on the first-added grade scheme element? Maybe 100, or 1000 just for demonstration's sake? Currently the min value is zero.

chcholman:
@zazaian Actually let's remove the 0 as the min value default (the first element added is supposed to be the highest level, and so 0 would be confusing). I'm worried that entering any number in the max value field will prompt them to try to work around that as opposed to understanding that it can be changed.

PR #2236, opened 2016-07-21
Firm Up Grade Scheme Element Implementation

zazaian:
This PR addresses the issues and features brought up in #2217. Namely:

- Whenever the high range of a grade scheme element is changed, the low range of the previous element should be dynamically adjusted as well.

- Similarly, whenever the low range of a grade scheme element is changed, the high range of the next element should be dynamically adjusted.

- Submissions should be disabled when there are empty or invalid issues in the form.

- There are some native angular `ng-form` behaviors that we weren't taking advantage of that make it much easier to juggle min/max validations without creating custom validators.

- add some CSS that works with the native angular ng-form attributes

- add a warning message that shows only when the form is invalid.

I began to implement some of the methods on the controller through the original highRange and lowRange directives, but it was much more cumbersome to re-create the ng-change functionality that made this all work so smoothly.

Additionally it was impossible to digest/apply the min/max behaviors for both high and low ranges within the gradeSchemeElementsService without injecting the scope into every method, so it felt cleaner to leave these in the controller. There is further explanation of these choices in comments on the controller and the grade scheme directives files.

@acousticrobot please be so kind as to review this and let me know your thoughts when you get an opportunity.


app/assets/javascripts/angular/directives/gradeSchemeDirectives.js.coffee
```diff
@@ -0,0 +1,53 @@
+# I appreciate the virtues of a directive-oriented implementation but in this
```
acousticrobot: Yeah, I'd say go ahead and remove the Directives page, if we decide to go this route later we can pull this file up again but most likely the whole approach will be rethought. (More general comments to follow.)

app/assets/javascripts/angular/services/GradeSchemeElementsService.js.coffee
```diff
@@ -1,18 +1,18 @@
 @gradecraft.factory 'GradeSchemeElementsService', ['$http', ($http) ->
     elements = []
     deletedIds = []
-    totalPoints = null
+
+    # we're doing so much index juggling here that we should really create a
+    # factory for both the overall grade scheme elements collection, as well as
+    # for the individual grade scheme elements so we don't have to do all of
+    # this work inside of the service object.
```
acousticrobot: I'm not sure a factory would help you with the index juggling, I see you had to do something similar in the [Criterion Factory](https://github.com/UM-USElab/gradecraft-development/blob/master/app/assets/javascripts/angular/factories/Criterion.js.coffee#L49-L51).

app/assets/javascripts/angular/services/GradeSchemeElementsService.js.coffee
```diff
@@ -1,18 +1,18 @@
 @gradecraft.factory 'GradeSchemeElementsService', ['$http', ($http) ->
```
acousticrobot: I have an [open PR](https://github.com/UM-USElab/gradecraft-development/pull/2237) where I broke the models in the `PredictorService` into their own services but I left `GradeSchemeElements` in the `PredictorService` for now since I knew you were working on it and I didn't want to create merge conflict for you to deal with. One thing I like about the current state of this service is that it's written in a way that manages GSEs reusable for other pages and outside of the context of any controller $scope or call to `new`.

app/assets/javascripts/angular/services/GradeSchemeElementsService.js.coffee
acousticrobot: I didn't use any factories when implementing the Predictor because I found the existing factories too tightly coupled to the Rubrics to reuse. I might be missing some of the benefits, but my understanding is that Angular Services and Factories [accomplish the same thing](http://stackoverflow.com/questions/14324451/angular-service-vs-angular-factory).

app/assets/javascripts/angular/services/GradeSchemeElementsService.js.coffee
```diff
@@ -22,53 +22,42 @@
...
-        true
+    lowestPoints = (index) ->
+      nextElement = elements[index]
+      if nextElement && nextElement.highest_points > 0
+        nextElement.highest_points + 1
       else
-        false
+        null
 
     getGradeSchemeElements = ()->
```
acousticrobot: One change I would like to make (can be in a different PR) is for the `getGradeSchemeElements` method to use the `/api/grade_scheme_elements` route. I added `highest_points` to the JSON in my last PR for your page in anticipation. Then I can call inject this service into the Predictor with the rest.

If it returns pure JSON, we could move the `postGradeSchemeElements` endpoint into the `api/` namespace as well.

app/assets/javascripts/angular/directives/gradeSchemeDirectives.js.coffee
zazaian: Cool, I just put this here to show a kind of proof of concept and to try not to meddle too much with the initial implementation beyond what was really required. I could see how it would be useful to have a more universal low-range high-range directive if this were being used in other parts of the app, but because this is so context specific the utility of the directive, especially considering the additional work required, is probably limited.

app/assets/javascripts/angular/services/GradeSchemeElementsService.js.coffee
zazaian: In my view the difference between the utility of a factory and a service is more of a conceptual one than a functional one. They can accomplish almost exactly the same outcome depending on how they're implemented, but conceptually I'd suggest that factories should be more coupled with what we'd consider a more object-oriented model pattern, and a service would reflect some class that's responsible for orchestrating models, but isn't necessarily a model itself. Does that make sense? I guess I'm just reflecting here on the idea of interesting ways to bundle sets of behaviors inside of a single object so that, for example, when we know a given grade scheme element is at index 1, we can manage its properties, and the higher and lower elements from within that class, rather than having to pass the index around to multiple methods.

app/assets/javascripts/angular/services/GradeSchemeElementsService.js.coffee
zazaian: Cool, I'm happy with doing this however you'd like to. Maybe it makes the most sense to merge this PR in, and setup another branch to address this issue?

zazaian:
@acousticrobot I think removing the submit button here is probably a move in the right direction, though because there seem to be lots of moving pieces here I wonder how it would feel to be constantly updating your changes rather than just being able to edit the overall scheme to get all of the individual elements as you'd like them, and then submit them. This is probably a call for @chcholman to make based on her feedback for professors, as you suggested.

Either way this is still being managed entirely by the angular-form underpinnings, and the only difference really would be when the form and/or individual elements are saved. Otherwise, the form management by angular would be pretty much identical.

Notes on the requested fixes/changes:
- I've fixed the invalid sample, and additionally formatted the range numbers properly with underscores at the thousands and millions so they'd be rubocop-compliant.
- I'm not totally sure what you're asking for with the CSS change to the button visibility. Are you suggesting that all of the possible add level buttons should always be visible? I wonder if in this current format that isn't perhaps too intrusive and hurts the readability of the data. Maybe it would make more sense for the add buttons to be between the elements, but as smaller [ + add level ] buttons in their own column to the left of the grade scheme elements. With this approach the data wouldn't have to be interrupted by the buttons, and they could be visible at all times.
- I'll take a look at the +/- functionality and see if we can work around that easily.

app/assets/javascripts/angular/services/GradeSchemeElementsService.js.coffee
acousticrobot: Outside of Angular I would totally agree with you. It feels as wrong storing state in something called a  "Service" to me as it does making API calls from a "Factory". Using Angular though, I think we're always going to end up violating our notions of what these pattens mean, or violating Angular patterns themselves.  

One limitation I've had with OOP approach, while it makes it easier to duplicate the back-end logic with more "self aware" models, is that it requires them, from what I've seen, to be managed from within the controller. I need to know, for instance, that the Assignments I'm working with in the `PredictorCtlr`, are the same ones available in the `studentPanel` directive, and completely decoupled from the scope.

I've had good results so far maintaining a more functional approach to the model management. Without needing to be as pure as a Flux/Redux pattern, by keeping the model state in arrays and close to the original JSON, I've found the services remain context agnostic and reusable on other pages. And it has the benefit that when it comes to making a POST/PUT request, they don't need to be manipulated back into JSON, resulting in more predictable parameters received in the ruby.

I get nervous loosing this clarity for the benefit of what a factory provides when I see currently that the `Badge` factory is dependent on the `EarnedBadge` factory and also requires `$timeout`, that `EarnedBadge` requires `$q`,  that in the `RubricCtrlr` badges are actually managed through the `CourseBadge` factory, and that it also needs to pass `$scope` into the `Criterion` factory. I might still be missing a pattern here, but I couldn't see a way to use these resources in the Predictor. 

On the converse, I like that the GradeSchemeElements service, as it currently is written for a completely different purpose, would only take a couple lines of code to completely integrate it into the Predictor Service.

app/assets/javascripts/angular/services/GradeSchemeElementsService.js.coffee
acousticrobot: Absolutely! Once we get this in, I can make a couple quick changes in another PR, and test to make sure both the predictor and GSE page still work with the service.

app/assets/javascripts/angular/services/GradeSchemeElementsService.js.coffee
zazaian: So I'm not really talking about making API calls from factories so much as storing sets of behaviors. When we have one GradeSchemeElement, for example, or even one HighRangeInput, there are going to be lots of contextual behaviors that that element has based on its place in the `$scope.elements` array. We've determined that it's impossible to manage some of those behaviors in services without injecting the `$scope` into the service, which seems to defeat the purpose of a Service at all, but additionally we've determined that it's cumbersome to re-build any of these behaviors through directives, as most of the functionality that we need is already being provided through existing Angular directives such as `ng-min`, `ng-max`, and `ng-change`, etc.

So with all of those things accounted for, there's really nowhere to put a 'highRangeMax' method that also takes advantage of these native angular directives except for something like a GradeSchemeElementFactory, which would have methods like `highRangeMax()`, `lowRangeMax()`, `highRangeMin()` and `lowRangeMin()`, but could return the appropriate results without having to pass the `$index` into each one as we're currently doing. There's no reason this would need to make API calls, though also I don't see why that's a responsibility that should lie entirely with services.

Additionally I'm cautious about trying to write code for the express purpose of re-usability in other contexts. I think it's great when there are more universal patterns that can serve multiple contexts, but when we have more context-specific problems to resolve I don't think that a lack of reusability is a compelling reason to de-couple a factory or other library from the immediate context. That said, a `GradeSchemeElementFactory` in this context would be just as reusable as a Service would be, it's just more coupled to the idea of a single GradeSchemeElement entity than it is to serving everything within the larger GradeSchemeElement functional context.

zazaian:
This is probably a gap in your experience with angular forms as you said, but there's a ton of power in `ng-form`, which might actually be the best feature in Angular. It's very intelligent and adds a lot of functionality in checking input dirtiness, validity, max, min, changed states, etc, and does all of this based on directive triggers set on the input elements themselves. This can all be re-modeled through a directive, but it's literally a re-implementation of native Angular functionality, and seems much denser to parse through while also being less full-featured.

I understand what you're saying about the new level button and the motion sickness, etc. I was just suggesting that maybe the levels should still be closer together for the sake of readability. Something like this might solve all of the problems that have been brought up:

grade_scheme_elements

So the limitation is just that when a `max` or `min` is set on an input in Angular, it's possible to type an invalid amount in, but the field and the form will be marked invalid (in this case the input background turns red). With the +/- buttons, however, if a max/min has been defined, the +/- will not increment/decrement below that amount.

zazaian:
Yes, I understand what you're expressing. In order to accomplish this we
need to remove the max constraint from the high range fields, and the min
constraint from the low range fields. I'll remove those constraints.

Also I was thinking about adding smart number but let's do that in another
PR.

zazaian:
Awesome, thanks for taking the time to review all of that so thoroughly.
I'd be happy to add new issues for the +new button style as well as the
smart number addition if that hasn't been done already.

PR #2247, opened 2016-07-27
Cleanup course table

chcholman:
This PR drops 8 columns from the Courses table, updates naming conventions to be as discussed in #tuesdaychats and adds not-null/default values to many fields. It also moves the basic group size constraints to be per assignment rather than at the course level. 

The groups work is partial - much to be done in the next two weeks to complete - but I believe maintains the current state of the features.


db/migrate/20160721204652_cleanup_courses.rb
```diff
@@ -0,0 +1,58 @@
...
+    remove_column :courses, :media_caption
+
+    change_column :courses, :name, :string, null: false
+    change_column :courses, :created_at, :datetime, null: false
+    change_column :courses, :updated_at, :datetime, null: false
+    rename_column :courses, :courseno, :course_number
+    change_column :courses, :course_number, :string, null: false
+
+    rename_column :courses, :user_term, :student_term
+    change_column :courses, :student_term, :string, default: "Student", null: false
```
acousticrobot: 👍 💯 great way to handle this!

chcholman: Thanks so much!

PR #2248, opened 2016-07-27
2239 remove default weight

acousticrobot:
This PR removes the `default_weight` calculations from a course (but leaves the field on the model for historical purposes). When an AssignmentType accepts weights, but a student has not placed any weights on it, all Grades for it's assignments have a zero score.

The UX remains for the most part the same. When weights are available, any AssignmentType that they can be added to indicates acceptance with a blue circle:

screen shot 2016-07-27 at 2 27 05 pm

To explain the calculation when no more weights are available, or `max_assignment_types_weighted` has been reached, a "no weights" indicator will show up in the summation.

On the Predictor:

screen shot 2016-07-27 at 2 27 23 pm

On the weights page,

screen shot 2016-07-27 at 2 27 51 pm

The recalculation of score on the grades looks good to me, but I would like it if someone else could test and confirm this, and that it displays correctly on all pages where grades are displayed in list form, since I don't know these pages well.

I'm also not sure if the display of zero'd out grades on the grades show pages is clear or not:

screen shot 2016-07-27 at 2 57 27 pm

### DIFF
--- a/app/assets/javascripts/angular/controllers/PredictorCtrl.js.coffee
+++ b/app/assets/javascripts/angular/controllers/PredictorCtrl.js.coffee
@@ -23,7 +23,7 @@
     PredictorService.termFor(article)
 
   $scope.unusedWeightsRange = ()->
-   PredictorService.unusedWeightsRange()
+    PredictorService.unusedWeightsRange()
 
   $scope.weightsAvailable = ()->
     PredictorService.weightsAvailable()
--- a/app/assets/javascripts/angular/controllers/PredictorCtrl.js.coffee
+++ b/app/assets/javascripts/angular/controllers/PredictorCtrl.js.coffee
@@ -22,6 +22,8 @@
   $scope.termFor = (article)->
     PredictorService.termFor(article)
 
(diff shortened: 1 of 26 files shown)


mariehooper:
Hi @acousticrobot ! This is looking awesome - love the 'no multipliers' addition! Would you mind making the multipliers look like this:
screen shot 2016-07-28 at 9 48 35 am

I think this will help students see exactly what math is occurring with these and understand the whole process a bit better.
- [ ] Show the additional multipliers added 'stacked' beneath the first and on the first coin show the actual number that represents. (ex: 1 multiplier shows a 1 and 3 multipliers shows a 3)
- [ ] To remove multiplier, hover on top coin and show 'x' icon that can be clicked to remove
- [ ] To add multiplier, change the outline circle to same yellow as coin with yellow '+'

chcholman:
@acousticrobot @mariehooper I love this so much!

mariehooper:
@acousticrobot Absolutely LOVE it!

One small thing - would you mind changing the light blue outline color of the empty circles to the yellow? Then it will be 💯

Issue #2203, opened 2016-07-11
Merge the Controller/Presenter/Mailers for all Export Workflows


zazaian:
Currently `SubmissionsExport` and `CourseAnalyticsExport` classes are using unique controllers, presenters and mailers to execute these workflows. Since they're so similar, however, these should probably be performed through a common `ExportsController` and at least a common base `Presenter::Exports::Base` class.

Additionally common model behaviors could be extracted into an `ExportsConcern`

PR #2250, opened 2016-07-29
Course Analytics Exports Jobs

zazaian:
The main thrust of this PR is that the course analytics export process is so resource-intensive that when triggered from a controller on Production the request would timeout. The best solution to this problem is to perform these exports, previously handled entirely by the `AnalyticsController#export` method, through a new `CourseAnalyticsExport` model with a corresponding `CourseAnalyticsExportPerformer` performing the build and notifying the user of success or failure via email.

Because we're currently doing work on `AnalyticsController#export` method on the #2118 branch, we've ported the entire legacy method into `CourseAnalyticsExportPerformer#build_the_export` for the time being. It's being tested just to ensure that it passes. A total refactor of this process will come in with #2118.

Additionally we're now also saving these exports on S3 and giving users access to them through the `SecureToken` process that we've already implemented in the `SubmissionsExport` model. Since there was already so much logic for handling exports via  `SubmissionsExport` anything that wasn't specific to that model has been extracted to a new `Export::Model` module in `/lib` which handles common export behaviors. This is now included in both the `SubmissionsExport` and `CourseAnalyticsExport` models.  Similarly, any common S3-specific logic has been extracted to `S3Manager::Resource`.

Because the process for `SubmissionsExport` and `CourseAnalyticsExport` creation is so similar, it's possible in the future that these should be handled through a common `ExportsController` that takes an `export_type` parameter rather than having redundant controller logic, but this will be the subject of another issue altogether.


app/controllers/analytics_controller.rb
```diff
@@ -182,113 +182,6 @@ def assignment_prediction_averages
     render json: MultiJson.dump(data)
   end
 
-  # TODO: fix this
```
jwright: 😄

app/controllers/course_analytics_exports_controller.rb
```diff
@@ -0,0 +1,60 @@
+# pull in the presenter that we need to bring this all together
+require "./app/presenters/course_analytics_exports/base"
+
+class CourseAnalyticsExportsController < ApplicationController
+  before_filter :ensure_staff?, except: :secure_download
```
jwright: Why do we skip this here? Is it because we don't want to require a staff member to sign in? Would they already be signed in?

app/controllers/course_analytics_exports_controller.rb
```diff
@@ -0,0 +1,60 @@
...
+  skip_before_filter :increment_page_views, only: :secure_download
+  skip_before_filter :course_scores, only: :secure_download
+
+  def create
+    if presenter.create_and_enqueue_export
+      flash[:success] = "Your #{presenter.resource_name} is being prepared. " \
+                        "You'll receive an email when it's complete."
+    else
+      flash[:alert] = "Your #{presenter.resource_name} failed to build. " \
+                      "An administrator has been contacted about the issue."
```
jwright: LIES!....or have they? :) I cannot see where someone get's notified if there is an issue or does this happen as part of the queuing and execution of background jobs?

app/mailers/course_analytics_exports_mailer.rb
```diff
@@ -0,0 +1,45 @@
...
+    @export = export
+    @course = export.course
+    @secure_token = token
+
+    send_mail status: "is ready"
+  end
+
+  protected
+
+  def send_mail(status:)
```
jwright: Why do we use named parameters here? I feel like named parameters are useful when there are many and we want to pass in a hash. Here, we can clear this up with just using `status` and the calling methods could just use: `send_mail "is ready"`. Thoughts?

app/models/course_analytics_export.rb
```diff
@@ -0,0 +1,60 @@
...
+  # give this resource additional methods that aren't s3-specific but that
+  # assist in the export process
+  #
+  include Export::Model
+
+  attr_accessible :course_id, :professor_id, :last_export_started_at,
+    :last_export_completed_at, :last_completed_step
+
+  belongs_to :course
+  belongs_to :professor, class_name: "User", foreign_key: "professor_id"
```
jwright: I feel like `owner` is a better name here since the `User` that created this can be a professor, gsi, or admin.

Thoughts?

app/models/course_analytics_export.rb
```diff
@@ -0,0 +1,60 @@
...
+    "exports/courses/#{course_id}/course_analytics_exports/" \
+      "#{object_key_date}/#{object_key_microseconds}"
+  end
+
+  def formatted_course_number
+    # create a url-safe course number to use for the export filename and for
+    # the root directory of the final export archive. be sure to replace
+    # forward-slashes with hyphens and replace ampersands with "and"
+    #
+    course_number = course.courseno.gsub(/\/+/,"-").gsub("&", "and")
```
jwright: What about all of the other possible uri-unsafe characters. I feel like we should use `URI::escape` here.

app/performers/course_analytics_export_performer.rb
```diff
@@ -0,0 +1,135 @@
...
+  def setup
+    @export = CourseAnalyticsExport.find attrs[:export_id]
+    @professor = @export.professor
+    @course = @export.course
+
+    export.update_export_started_time
+  end
+
+  # perform() attributes assigned to @attrs in the ResqueJob::Base class
+  def do_the_work
```
jwright: 👍

app/mailers/course_analytics_exports_mailer.rb
```diff
@@ -0,0 +1,45 @@
+class CourseAnalyticsExportsMailer < ApplicationMailer
+  layout "mailers/exports_layout"
+
+  attr_reader :professor, :course, :export, :status
+
+  # the SecureTokenHelper brings in the #secure_downloads_url method which we
+  # need for building the secure download method on success emails
+  add_template_helper(SecureTokenHelper)
+
+  def export_failure(export:)
```
jwright: For mailer method names, I think they should end in `_email`. So `export_failure_email`. Thoughts?

app/mailers/course_analytics_exports_mailer.rb
```diff
@@ -0,0 +1,45 @@
...
+    @professor = export.professor
+    @course = export.course
+
+    send_mail status: "failed to build"
+  end
+
+  def export_success(export:, token:)
```
jwright: For mailer method names, I think they should end in `_email`. So `export_success_email`. Thoughts?

app/controllers/course_analytics_exports_controller.rb
zazaian: We're skipping this because the point of the secure_download action is to be able to download resources with a link via SecureToken authentication. Requiring login in addition to access the action makes it impossible to download the requested resource from the link.

app/controllers/course_analytics_exports_controller.rb
zazaian: We'll get the failure data through loggly and NewRelic, and probably don't need to expressly send an email to the admin whenever one of these fails, though we can add that if it seems necessary.

app/mailers/course_analytics_exports_mailer.rb
zazaian: I like using keyword arguments here because passing an argument to `#send_mail` doesn't implicitly suggest that it should be taking a status. We could use un-named arguments here but it seems more confusing to me.

app/models/course_analytics_export.rb
zazaian: I think owner is okay, but professor was being used here to suggest that specifically that this isn't a student. Maybe there's a better way to name this that includes this suggestion? Maybe `faculty`?

app/mailers/course_analytics_exports_mailer.rb
zazaian: I'd be okay with this, except we're not currently doing this anywhere else else in our mailer suite so we'd have to change everything else in the name of consistency. Maybe let's add a separate ticket for this and/or make it a TuesdayChats topic to see how people feel.

app/models/course_analytics_export.rb
chcholman: Let's do owner - it won't always be someone on the instructional team (ie, `admin`)

zazaian: :+1: cool, I guess we'll understand that students won't have access to exports

zazaian:
@jwright build is passing with the changes from `Course#courseno` -> `Course#course_number`

app/controllers/course_analytics_exports_controller.rb
jwright: That is what I thought. Thanks!

app/controllers/course_analytics_exports_controller.rb
jwright: I think this is fine. Thank you.

app/mailers/course_analytics_exports_mailer.rb
jwright: I like that idea. I will add it to TuesdayChats 🎼

app/controllers/course_analytics_exports_controller.rb
zazaian: :+1:

app/controllers/course_analytics_exports_controller.rb
zazaian: :sunglasses:

app/mailers/course_analytics_exports_mailer.rb
zazaian: Cool, sounds good!

zazaian:
@jwright FYI I've made the `professor` -> `owner` changes and pushed them to `origin`.

app/presenters/course_analytics_exports/base.rb
```diff
@@ -0,0 +1,78 @@
...
+      def export_job
+        @export_job ||= ::CourseAnalyticsExportJob.new export_id: export.id
+      end
+
+      def export
+        @export ||= ::CourseAnalyticsExport.find params[:id]
+      end
+
+      # destroying an export should probably automatically delete it from s3
+      # as well. let's modify that behavior in a future PR.
```
jwright: Do we have an issue for this? If not, can you add one and link it here?

zazaian: Sorry, I just neglected to remove the comment here. This has been implemented in this PR.

db/schema.rb
```diff
@@ -214,6 +214,20 @@
     t.string   "media_caption",       limit: 255
   end
 
+  create_table "course_analytics_exports", force: :cascade do |t|
```
jwright: These have very similar fields to `submissions_export`. Do you think we should place these in an `exportable` concern with a common schema? If so, I am not suggesting it for this PR but perhaps in the future.

app/models/course_analytics_export.rb
zazaian: Any additional URI-unsafe characters are being removed in the next line by `Formatter::Filename#url_safe`.

db/schema.rb
zazaian: Yup, this is a course of action that I've discussed with @chcholman and will be addressed in a separate issue.

jwright:
@zazaian This is really, really good man. Great job!

I made several comments but I think they will be addressed in future PRs or through some discussion on Tuesday Chats.

The build is broken but once that is fixed, I think this can be passed on.

zazaian:
Awesome, @jwright ! Thanks for digging through it! I know it's a lot of code. I'm trying to keep these PR's smaller but the job/export workflow touches several parts of the system, which when tested balloons the size of the feature. Also there were lots of exports extractions to move toward the work that will be done whenever we get to #2203.

The broken build is coming in from a brakeman complaint in CodeClimate, "Potentially dangerous attribute available for mass assignment". The problem is that it's coming up regardless of whether the require statements in `CourseAnalyticsExport` are present, and when I remove them it complains about the class declaration. Any thoughts on what Brakeman might be choking on here?

zazaian:
@chcholman this is passing and ready to merge.

PR #2251, opened 2016-08-01
881 multi element leveling

chcholman:


app/views/students/_unlock_icons.haml
```diff
@@ -0,0 +1,11 @@
...
+  .unlock_conditions
+    %h3.bold This Level is Locked
+    .small.italic In order to unlock it you must:
+    - gse.unlock_conditions.each do |condition|
+      .condition= condition.requirements_description_sentence
+-# do we want to remind students of the conditions to unlock once they've completed them? 
+-# - else
+-#   .met_unlock_conditions
+-#     %h3.bold You have Unlocked this Level
+-#     .small.italic To fully achieve it, you must earn #{points gse.lowest_points} points
```
acousticrobot: Did you want to leave this in for now? I'm only asking because the wording "To fully achieve it" doesn't seem to fit with having already unlocked the level.

spec/controllers/grade_scheme_elements_controller_spec.rb
```diff
@@ -25,6 +25,22 @@
...
+    describe "GET edit" do
+      it "renders the edit grade scheme form" do
+        get :edit, id: @grade_scheme_element.id
+        expect(assigns(:grade_scheme_element)).to eq(@grade_scheme_element)
+        expect(response).to render_template(:edit)
+      end
+    end
+
+    describe "PUT update" do
+      it "updates the grade scheme element with unlock conditions" do
```
acousticrobot: I don't see unlock conditions actually tested in this spec

app/views/students/grading_scheme.html.haml
```diff
@@ -5,13 +5,15 @@
     = render "layouts/alerts"
 
     - @grade_scheme_elements.each do |element|
-      - if current_student.cached_score_for_course(current_course) < element.lowest_points
+      - if current_student.cached_score_for_course(current_course) < element.lowest_points || (current_student.cached_score_for_course(current_course) >= element.lowest_points && ! element.is_unlocked_for_student?(current_student))
         .progress.bar_magic.transparent.center
```
acousticrobot: Do we still have access to a presenter within this partial? If so, I would suggest moving this logic into a presenter, (but I wouldn't build a new one just for this.)

app/views/students/_unlock_icons.haml
chcholman: I'm going to take it out - we'll see if students find that confusing/request it back in.

spec/controllers/grade_scheme_elements_controller_spec.rb
chcholman: Good catch!

app/views/students/grading_scheme.html.haml
chcholman: We don't, that's why I hadn't moved it over. Definitely worth a refactor in the fall though!

Issue #2207, opened 2016-07-12
Import Canvas Grade


jwright:
As a professor
I should be able to select a single or multiple grades to import into GradeCraft
So that I can use GradeCraft alongside Canvas

PR #2252, opened 2016-08-02
2206 list canvas grades

### LINKED ISSUE #2206: List Canvas Grades
As a professor
I should be able to view what grades are available for a specific assignment
So that I can use GradeCraft alongside Canvas

jwright:
Allows a staff member to pull down submissions (grades) from Canvas for a single or multiple assignments to display the grades and the user information within a list. The option to import these will be completed with #2207.

This is what the flow looks like from the staff perspective:

1. On the assignments view, there is still the existing "Import Grades" button.

screen_shot_2016-08-02_at_1_29_59_pm

2. There are now 2 options on the grades import view, one for Canvas and the existing one for CVS.

screen shot 2016-08-02 at 1 30 14 pm

3. If they select Canvas, a list of courses from Canvas are displayed.

screen shot 2016-08-02 at 1 30 23 pm

4. Once they choose a course, a list of assignments from Canvas for that course are displayed. The user can select one or multiple assignments to choose the grades for.

screen shot 2016-08-02 at 1 30 36 pm

6. The grades are displayed next. An indicator on whether the user exists and is in the current course is displayed.

screen shot 2016-08-02 at 1 30 46 pm

Closes #2206


jwright:
@ksofa2 This may be of interest to you as well.

app/controllers/assignments/grades_controller.rb
```diff
@@ -50,7 +38,11 @@ def update_status
   def export
     assignment = current_course.assignments.find(params[:assignment_id])
     respond_to do |format|
-      format.csv { send_data GradeExporter.new.export_grades_with_detail(assignment, assignment.course.students), filename: "#{ assignment.name } Grades - #{ Date.today }.csv" }
+      format.csv do
```
zazaian: Cool, looks a lot better :+1:

app/controllers/grades/importers_controller.rb
```diff
@@ -0,0 +1,81 @@
+require "active_lms"
+
+class Grades::ImportersController < ApplicationController
+  before_filter :ensure_staff?
+
+  def assignments
+    @assignment = Assignment.find params[:assignment_id]
+    @provider = params[:importer_id]
```
zazaian: Maybe all of these instances of `@provider` should be `@provider_id` since it's just an id and not an actual resource?

app/controllers/grades/importers_controller.rb
```diff
@@ -0,0 +1,81 @@
+require "active_lms"
+
+class Grades::ImportersController < ApplicationController
```
zazaian: Does this maybe make more sense as `Importers::GradesController` than `Grades::ImportersController`? Seems like this would help keep all of the various importer controllers together as opposed to grouping them with their respective resources. It would also be more consistent with the `Presenters::Grade::Base` pattern that we're already using.

app/controllers/grades/importers_controller.rb
```diff
@@ -0,0 +1,81 @@
...
+  # GET /assignments/:assignment_id/grades/importers/:id
+  def show
+    @assignment = Assignment.find params[:assignment_id]
+    provider = params[:id]
+
+    render "#{provider}"
+  end
+
+  # POST /assignments/:assignment_id/grades/importers/:importer_id/upload
+  def upload
```
zazaian: This feels like a lot of logic for the controller. Maybe we should use a presenter for anything beyond direct instance variable fetching/assignment, and simple conditional logic for render/redirect?

app/controllers/grades/importers_controller.rb
```diff
@@ -0,0 +1,81 @@
...
+      grade_ids = @result.successful.map(&:id)
+      enqueue_multiple_grade_update_jobs(grade_ids)
+
+      render :import_results
+    end
+  end
+
+  private
+
+  # Schedule the `GradeUpdater` for all grades provided
```
zazaian: Maybe these would also be better in a presenter?

app/controllers/grades/importers_controller.rb
```diff
@@ -0,0 +1,81 @@
...
+  def enqueue_multiple_grade_update_jobs(grade_ids)
+    grade_ids.each { |id| GradeUpdaterJob.new(grade_id: id).enqueue }
+  end
+
+  def syllabus
+    @syllabus ||= ActiveLMS::Syllabus.new(@provider,
+                                          ENV["#{@provider.upcase}_ACCESS_TOKEN"])
+  end
```
zazaian: Maybe this syntax makes more sense, since Rubocop will complain about the > 2 space indentation?
```ruby
def syllabus
  @syllabus ||= ActiveLMS::Syllabus.new \
    @provider,
    ENV["#{@provider.upcase}_ACCESS_TOKEN"]
end
```

app/views/importers/_assignments_table.html.haml
```diff
@@ -0,0 +1,26 @@
+= form_tag path do
```
zazaian: :+1: looks good

app/controllers/grades/importers_controller.rb
```diff
@@ -0,0 +1,81 @@
+require "active_lms"
```
zazaian: Breaking this out into a dedicated controller makes a lot of sense, nice call. :+1:

app/helpers/partial_helper.rb
```diff
@@ -0,0 +1,10 @@
+module PartialHelper
+  def partial_with(partial, options={}, &block)
+    options.merge!({ partial: partial })
```
zazaian: probably don't need the curly braces on the `merge!`

app/helpers/partial_helper.rb
```diff
@@ -0,0 +1,10 @@
...
+    if block_given?
+      options[:locals] = {} if options[:locals].nil?
+      options[:locals].merge!({ content: block })
```
zazaian: samesies here, no curlies needed :curly_loop:

lib/active_lms/syllabus/canvas_syllabus.rb
```diff
@@ -46,6 +46,22 @@ def assignment(course_id, assignment_id)
       assignment
     end
 
+    def grades(course_id, assignment_ids)
+      grades = []
+      client.get_data("/courses/#{course_id}/students/submissions",
```
zazaian: This feels a little awkward since we're opening a block at the end of this very large `get_data` call. Maybe we can splat the arguments here as an array to keep the `get_data` call on one line?

spec/controllers/grades/importers_controller_spec.rb
```diff
@@ -0,0 +1,89 @@
...
+        expect(response.body).to \
+          include("First Name,Last Name,Email,Score,Feedback")
+      end
+    end
+
+    describe "POST upload" do
+      render_views
+
+      let(:file) { fixture_file "grades.csv", "text/csv" }
+
```
zazaian: So as the person who didn't write the code, without `context` blocks here it's really hard to figure out what the different scenarios are that are being tested. One of the nice things about using context blocks is that you can expressly declare what the failing cases should be, or at least in which scenarios failure should occur, instead of having to look at the setup inside of each example block to figure out what's different between the one that's failing an the one that's passing.

It's very difficult to parse here what the difference is here between the failing and passing cases without looking into the other examples, which suggests that the structure of the individual examples probably aren't expressive enough.

For example, in the "renders any errors that have occurred" example, I have to look to the example above it to figure out that the first one probably didn't fail because students were created.

Wrapping that example in a context block as I have below at least tells me why it's failing without having to dig through the code.

```ruby
context "no users exist for the import" do
  it "renders any errors that have occured" do
    post :upload, assignment_id: world.assignment.id, importer_id: :csv, file: file
    
    expect(response.body).to include "3 Grades Not Imported"
    expect(response.body).to include "Student not found in course"
  end
end
```

spec/controllers/grades/importers_controller_spec.rb
```diff
@@ -0,0 +1,89 @@
...
+      end
+
+      it "renders any errors that have occured" do
+        post :upload, assignment_id: world.assignment.id, importer_id: :csv, file: file
+
+        expect(response.body).to include "3 Grades Not Imported"
+        expect(response.body).to include "Student not found in course"
+      end
+
+      it "adds error and redirects without a file" do
```
zazaian: This is also kind of confusing. In the absence of a distinct context block to separate out the cause and effect of the spec state here, it sounds like the end-result is to add an error, and then to redirect-without-a-file, rather than to say that when there's no file present, it should add an error and redirect.

Again, this should probably be rewritten as:
```ruby
context "no file exists" do
  it "adds error and redirects" do
    # example...
  end
end
```

spec/helpers/lms_helper_spec.rb
```diff
@@ -0,0 +1,26 @@
+require "rails_spec_helper"
+
+describe LMSHelper do
+  describe "#lms_user_match?" do
+    let(:course) { create :course }
+    let(:email) { "jimmy@example.com" }
+    let(:membership) { create :student_course_membership, user: student, course: course }
+    let(:student) { create :user, email: email }
+
```
zazaian: This is also a little confusing that we have to implicitly cache a student or membership resource at the beginning of the example for it not to fail, but since there's so little code it's not really a problem. It seems like at the very least when the example is dependent on a caching behavior like this it should at least have a comment suggesting that this is happening.

Also if no context is being used failing examples should come after passing ones so that you don't have to look down the spec to figure out why these are failing. I've been trying to play along with the very-few-contexts-in-specs approach but in many of these cases unless everything is ordered and commented perfectly it seems like it really hurts the clarity of the examples.

zazaian:
Hey @jwright -- I think this work all looks pretty good. There are probably some workflow changes with the new `Grades::ImporterController` to be made, specifically with delegating some of that code to a presenter.

Additionally it seems like some of the spec examples are relying very heavily on understanding the workflow as-is in order to recognize what should and should not be failing in various examples. I think using contexts judiciously here would go a very long way to clarifying the language and some of the logic for how the examples are being grouped, especially for failure states. Further commenting on some of these examples would probably also go a long way to improving readability.

app/controllers/assignments/grades_controller.rb
jwright: I think this is a different issue as I just fixed a rubocop issue here.

app/controllers/grades/importers_controller.rb
jwright: Hmmm....I don't feel using `provider` here assumes "an object". I classify things like `:canvas` and `:blackboard` are lms providers and not `provider_id`s. I feel like we should leave this alone.

app/controllers/grades/importers_controller.rb
jwright: Ah, yes. I agree. I would like to group all of the importers together. I am unsure how to accomplish this and keep the routes in tact.

Do you agree with these routes?

| Verb | Url                                                                                                                                       
| ------| -----------------------------------------------------------------------------------
| GET   | /assignments/importers/:importer_id/courses
| GET   | /assignments/importers/:importer_id/courses/:id/assignments
| POST | /assignments/importers/:importer_id/courses/:id/assignments/import
| GET   | /assignments/importers
| GET   | /assignments/:assignment_id/grades/importers/:importer_id/download
| POST | /assignments/:assignment_id/grades/importers/:importer_id/upload
| GET   | /assignments/:assignment_id/grades/importers/:importer_id/courses
| GET   |/assignments/:assignment_id/grades/importers/:importer_id/courses/:id/assignment
| POST | /assignments/:assignment_id/grades/importers/:importer_id/courses/:id/grades
| GET   | /assignments/:assignment_id/grades/importers
| GET   | /assignments/:assignment_id/grades/importers/:id

app/controllers/grades/importers_controller.rb
jwright: This was logic that was existing. I just moved it to this controller. Should we clean this up in the PR or in another? I am open to either.

I feel like using a presenter here is overkill as it's another level of indirection and there is one if statement here. Thoughts?

app/controllers/grades/importers_controller.rb
jwright: I think an object enqueuing multiple updates would be a better step here than putting these in a presenter.

app/controllers/grades/importers_controller.rb
jwright: 👍

app/helpers/partial_helper.rb
jwright: 👍

spec/controllers/grades/importers_controller_spec.rb
jwright: 👍

jwright:
@zazaian I fixed most of the issues about (except for the controller one) which I have a question about.

app/controllers/grades/importers_controller.rb
zazaian: I'm sorry, I misread what was happening here. I think this looks fine, though why are we using `provider` instead of `importer`?

app/controllers/grades/importers_controller.rb
zazaian: I think that sounds like a good approach. In reviewing this further it seems like the additional logic enqueue_multiple logic should probably be handled by a dedicated job

spec/controllers/grades/importers_controller_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+        second_student = create(:user, username: "jimmy")
+        second_student.courses << world.course
+        ResqueSpec.reset!
+
+        post :upload, assignment_id: world.assignment.id, importer_id: :csv, file: file
+
+        expect(GradeUpdaterJob).to have_queue_size_of(2)
+      end
+
+      context "with students that are no a part of the current course" do
```
zazaian: Typo :capital_abcd:

app/controllers/grades/importers_controller.rb
zazaian: Okay, so I'm reviewing these and before I get into the routes specifics, I'd like to ask about the workflow a bit. Can you explain the utility of passing an `importer_id` in with each request?

app/controllers/grades/importers_controller.rb
jwright: The importing of LMS data takes in a provider using the `adapter` pattern. The reason we use `importer` here is because of routing taking in an `importer_id` for the `ImportersController`. We could change this to `provider_id` using the Rails routing `param` option.

app/controllers/grades/importers_controller.rb
jwright: The utility is for what type (provider) of importing we will be importing from. For example, today we have `:canvas` and `:csv` but we will soon have others like `blackboard`, etc.

app/controllers/grades/importers_controller.rb
jwright: Should we introduce that in this PR or add an issue?

zazaian: @jwright let's add a separate issue for that. Sounds fine to me if you're cool with that.

app/controllers/grades/importers_controller.rb
zazaian: `provider_id` probably makes more sense in this context since we're calling it provider in the controller as well. Do you agree?

app/controllers/grades/importers_controller.rb
zazaian: Let's add an issue for this in a separate PR as well.

app/controllers/grades/importers_controller.rb
zazaian: Okay, I think that sounds like a nice way to handle those different patterns. I guess I'm still just tripping up on the variable naming a bit. When I see the `_id` suffix I'm used to expecting an integer in Rails, so it's confusing to me that this route is taking a string. Maybe instead of `importer_id` we can use something like `provider_type`, `provider_name`, or just `provider`? Does that make sense? Sorry to nitpick on this, just trying to add some clarity here where possible.

zazaian:
@jwright couple of comments on your comments, specifically let's bounce those out-of-scope items to other issues, and the last item here I think is figuring out the most expressive naming for `importer_id`.

app/controllers/grades/importers_controller.rb
jwright: I agree. I will change this using the `param` option and probably just use `provider_id` instead of assigning it to `provider`

app/controllers/grades/importers_controller.rb
jwright: @zazaian I think the change above to `provider_id` will handle this as well. Thoughts?

app/controllers/grades/importers_controller.rb
jwright: Created #2266.

app/controllers/grades/importers_controller.rb
jwright: Created #2267

zazaian:
@jwright everything looks good to go. Nice work :+1:

PR #2259, opened 2016-08-04
2170 authenticate with canvas

### LINKED ISSUE #2170: Authenticate Canvas with OAuth

jwright:
Uses [omniauth-canvas](https://github.com/atomicjolt/omniauth-canvas) to authorize with Canvas over oauth and stores the user's credentials in a new `UserAuthentication` table. The credentials are then used to access the Canvas API instead of using the existing single auth token.

Since Canvas returns short-lived tokens (3 hours I believe), the ability to refresh the token with a `refresh_token` was added so it can be done behind the scenes without interaction from the user.

Since this is using oauth to test on the development side, we can't have the callback url be localhost. We have setup the callback url using [ngrok](ngrok.io) with `https://gradecraft-development.ngrok.io`. You can download and run ngrok with the following command: 

`ngrok http -subdomain=gradecraft-development 5000`

Closes #2170


jwright:
@jonathoy Here is the PR that includes storing the authorizations with omniauth if you want to peruse or ask questions.

zazaian:
Hey there @jwright, I'll take a look at this forthwith. Sorry for the delay.

app/controllers/importers_controller.rb
```diff
@@ -42,8 +43,30 @@ def assignments_import
...
+
+  def require_authorization_with(provider)
+    authorization = UserAuthorization.for(current_user, provider)
+
+    if authorization.nil?
+      session[:return_to] = importer_courses_path(provider)
+      redirect_to "/auth/#{provider}"
+    elsif authorization.expired?
+      config = ActiveLMS.configuration.providers[provider.to_sym]
+      authorization.refresh!({ client_id: config.client_id,
```
zazaian: Having to call a method on the client for every key in this hash feels like code smell to me. We should probably just have a method that accepts the config, and then allow the authorization to handle the refresh from the config. Something like this:

```ruby
authorization.refresh_with_config! config
```

app/controllers/importers_controller.rb
```diff
@@ -42,8 +43,30 @@ def assignments_import
...
+                               client_secret: config.client_secret,
+                               client_options: config.client_options })
+    end
+  end
+
+  def authorization(provider)
```
zazaian: Why do we have a method here to build an authorization, but then aren't using it in the `#require_authorization_with` method? `#authorization` should probably build and cache an authorization instance to an `@authorization` ivar on the controller so that we can re-use it in both `#require_authorization_with` and `#syllabus` where it's also being called.

app/controllers/authorizations_controller.rb
```diff
@@ -0,0 +1,9 @@
+class AuthorizationsController < ApplicationController
```
zazaian: What's the plan for this class? Will this just handle oauth authorizations? Are there other authorizations that will be handled here?

app/controllers/importers_controller.rb
```diff
@@ -42,8 +43,30 @@ def assignments_import
 
   private
 
+  def require_authorization
+    provider = params[:importer_id]
+    require_authorization_with provider
+  end
+
+  def require_authorization_with(provider)
+    authorization = UserAuthorization.for(current_user, provider)
```
zazaian: This seems like it's probably unnecessary because of the existing `#authorization` method on this controller. Let's just add caching there so this can be handled consistently between methods.

app/models/user_authorization.rb
```diff
@@ -0,0 +1,36 @@
+require "omniauth-canvas"
+
+class UserAuthorization < ActiveRecord::Base
+  belongs_to :user
+
+  def self.create_by_auth_hash(auth_hash, user)
```
zazaian: The variable declaration here feels kind of redundant since we have to pass the same variable names in as keys to the `#update_attributes` call seven lines down. Maybe this is a more succinct way of expressing the same idea:

```ruby
  def self.create_by_auth_hash(auth_hash, user)
    provider = auth_hash["provider"]
    credentials = credentials_attrs auth_hash["credentials"]

    authorization = self.find_or_initialize_by provider: provider, user_id: user.id
    authorization.update_attributes credentials
    authorization
  end

  protected

  def credentials_attr(credentials)
    {
      access_token: credentials["token"],
      refresh_token: credentials["refresh_token"],
      expires_at: Time.at credentials["expires_at"]
    }
  end
```

app/models/user_authorization.rb
```diff
@@ -0,0 +1,36 @@
+require "omniauth-canvas"
+
+class UserAuthorization < ActiveRecord::Base
```
zazaian: I feel like `/models` is a poor place for an authorization class, particularly if we're going to have any of these beyond `UserAuthorization`. I could imagine creating a class akin to `SecureTokenAuthorization` as well as authorization classes for additional scenarios, and would probably want to see these in the same place. Something like `/app/authorizations` feels like it makes more sense to me here. Thoughts? Maybe this is a good TuesdayChats topic?

app/models/user_authorization.rb
```diff
@@ -0,0 +1,36 @@
...
+
+  def expired?
+    self.expires_at < DateTime.now
+  end
+
+  def self.for(user, provider)
+    where(user_id: user.id, provider: provider).first
+  end
+
+  def refresh!(options={})
```
zazaian: I wonder if the utility of this method is being lost in the somewhat involved variable declaration. Maybe this would be a bit more expressive if we fetched the provider class and constructed token from separate methods:

```ruby
  def refresh_attrs(options)
    return false if self.refresh_token.blank?

    strategy = provider_class.new nil, options
    token = token(strategy).refresh!

    self.update_attributes \
      access_token: token.token,
      refresh_token: token.refresh_token,
      expires_at: DateTime.now + token.expires_in.to_i.seconds
    }
  end

  def provider_class
    OmniAuth::Strategies.const_get self.provider.to_s.camelize
  end

  def token(strategy)
    OAuth2::AccessToken.new strategy.client, self.access_token,
      { refresh_token: self.refresh_token }
  end
```

config/initializers/active_lms.rb
```diff
@@ -0,0 +1,11 @@
+require "active_lms"
```
zazaian: :+1: this looks really good

lib/active_lms/configuration.rb
```diff
@@ -0,0 +1,31 @@
+module ActiveLMS
```
zazaian: I think this is really nicely done, but any thoughts on moving `ProviderCollection` and `Provider` into separate files?

spec/controllers/authorizations_controller_spec.rb
```diff
@@ -0,0 +1,47 @@
+require "rails_spec_helper"
+
+describe AuthorizationsController do
```
zazaian: Nicely implemented :+1:

zazaian:
Hey @jwright -- I think this looks really nice. I had some suggestions mostly on the `UserAuthorization` class regarding method complexity and have made some suggestions on how we might alleviate those things. Also there are still some failing examples in the spec suite, and a mass assignment complaint from Brakeman.

app/controllers/importers_controller.rb
jwright: 👍 Great idea.

app/controllers/importers_controller.rb
jwright: I can and should replace line 52 with a call to this method.

I cannot cache the instance on `authorization` since it requires the `provider` parameter and it's used in a `before_filter`. I could use the `params[:importer_id]` but I would be duplicating the code and I think it couples the code in the various methods too much.

app/controllers/authorizations_controller.rb
jwright: This is the endpoint for omniauth oauth callbacks. It is specified in the routing here:

https://github.com/UM-USElab/gradecraft-development/pull/2259/files#diff-21497849d8f00507c9c8dcaf6288b136R263

This will handle all the oauth integrations as there will be more.

app/controllers/importers_controller.rb
jwright: I changed it to call the method but we can't cache the method with an ivar due to the parameter.

app/models/user_authorization.rb
jwright: 👍

app/models/user_authorization.rb
jwright: 👍

lib/active_lms/configuration.rb
jwright: I *think* I am ok with moving them. I hesitate because I think these are more "implementation" classes that will never be used outside of configuration and so I think I like them as "hidden" classes. Thoughts?

If I do move them, I would like to move them under `lib/active_lms/configuration` and namespace them the same. However, I could not be able to use the `Configuration` module since it's already being used as a class. They would be namespaced with the class (as they are now). Are you ok with this?

```
# lib/active_lms/configuration/provider
module ActiveLMS
  class Configuration
    class Provider
      # ...
    end
  end
end
```

jwright:
@zazaian Thanks man! I really loved your suggestions. I made most of them and I have comments with the rest. I am going to pass this over to you for another look. I will rebase and fix the build/brakeman issues.

jwright:
@zazaian I rebased and made some changes to have the changes fit into both the assignments import and the grade import.

I fixed the valid issues for codeclimate but there are a few remaining. It says that `User` got worse because I added another relationship which is needed. The issue it found for `UserAuthorization` is invalid as `protected_attributes` are not needed and if I add some just to satisfy codeclimate, they will need to change when the `strong_parameters` PR is merged.

Everything seems to be good, just passing it back to you to review the changes and the comments.

lib/active_lms/configuration.rb
zazaian: Hey @jwright: I think this is fine as is, I just wanted to ask. These classes are insignificant enough that it feels fine to leave them here.

app/models/user_authorization.rb
zazaian: Nice updates, looks good!

app/models/user_authorization.rb
zazaian: Awesome, I think this feels a lot more accessible. :+1:

app/controllers/authorizations_controller.rb
zazaian: :+1: cool, sounds good

zazaian:
Hey @jwright, I think these updates all look really good. Deal with the CodeClimate stuff as you need to and I think this is ready to merge.

jwright:
@chcholman This is ready for merging.

PR #2260, opened 2016-08-08
Refactor Legacy AnalyticsController#export Workflow

zazaian:
The thrust of this PR is to finally tackle the very-large analytics export process that previously had been performed in a single method, originally in `AnalyticsController#export`, and the most recently in `CourseAnalyticsExportPerformer#build_the_export`. This process has been broken out into a few separate pieces since there was quite a bit going on.

The first component is a new `CourseExportContext` class which lives in `/app/analytics_exports/export_contexts`. This class takes all of the various data we needed to query from mongo, ActiveRecord, and from the analytics aggregates work and condenses them down into a single class that can perform and cache the queries, and return them in a single array through an method called `#export_data`. The idea here is that this class should provide sufficient data for any kind of course-level export.

The only shortcoming of this Context class right now is that it makes no distinction between data queried from Mongo vs. ActiveRecord vs. analytics exports, so beyond commenting there's no express segregation of these data sources within the context. I've proposed that we tackle this issue when we refactor the analytics aggregates classes, as they'll likely give us a more coherent sense of what might be useful in terms of further organization of these context classes.

The other major component of this process was the task of actually assembling the export itself, which has been delegated to a new `Analytics::Export::Builder` class. This class takes the `export_data` hash from the given Context class, as well as the export classes (in `/app/analytics_exports`) that we'd like to include in the export archive that we've requested to download, and actually performs the process of filtering the context data through each export class we've given it, and building the final output of those export actions.

Finally to tie this all together we've added some methods to the `CourseAnalyticsExport` model to tell the builder which export classes should be included in the final archive. Also we've included a new `Analytics::Export::Buildable` module in that class to add some methods for building the archive directly from the model class. This allows us to take the previous 120+ line method from `AnalyticsController#export` and perform it like this:

```ruby
export = CourseAnalyticsExport.find 5
export.build_archive!
```


zazaian:
@acousticrobot please take a look at this when you get a moment. I know you had some experience adding features to the analytics suite and think this stuff might be less alien for you because of it.

zazaian:
Hey there @acousticrobot, thanks for taking a look. I've been trying to keep these comments here for the benefit of the reviewer, but perhaps I'll pare them down in the future and move more of them into the PR itself. Would that make sense?

PR #2274, opened 2016-08-10
Feature/1328 rubric upload

acousticrobot:
This PR adds the ability to upload files attached to Rubric Grades, similar to the existing file upload available on the normal grading form.

The rubric grading form is handled through angular, so the multi-file upload will not work as it does with a standard form. Instead we upload the files in advance of submission, on change to the file input field.

### DIFF
--- a/app/assets/javascripts/angular/directives/gradeFileUpload.js.coffee
+++ b/app/assets/javascripts/angular/directives/gradeFileUpload.js.coffee
@@ -8,7 +8,7 @@
       element.bind('change', ()->
         scope.$apply(()->
           model.assign(scope, element[0].files)
-          RubricService.postGradeFiles(RubricService.grade, element[0].files)
+          RubricService.postGradeFiles(element[0].files)
         );
       );
     };
--- a/app/assets/javascripts/angular/directives/gradeFileUpload.js.coffee
+++ b/app/assets/javascripts/angular/directives/gradeFileUpload.js.coffee
@@ -1,0 +1,15 @@
+# adapted from https://uncorkedstudios.com/blog/multipartformdata-file-upload-with-angularjs
+@gradecraft.directive('gradeFileUpload', ['$parse', 'RubricService', ($parse, RubricService)->
+  return {
+    restrict: 'A',
+    link: (scope, element, attrs)->
+      model = $parse(attrs.gradeFileUpload)
+
+      element.bind('change', ()->
+        scope.$apply(()->
+          model.assign(scope, element[0].files)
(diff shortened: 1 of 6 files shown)


zazaian:
Hey there @acousticrobot -- sorry, just noticed this. I'll take a look at it today.

app/assets/javascripts/angular/directives/gradeFileUpload.js.coffee
```diff
@@ -0,0 +1,15 @@
...
+@gradecraft.directive('gradeFileUpload', ['$parse', 'RubricService', ($parse, RubricService)->
+  return {
+    restrict: 'A',
+    link: (scope, element, attrs)->
+      model = $parse(attrs.gradeFileUpload)
+
+      element.bind('change', ()->
+        scope.$apply(()->
+          model.assign(scope, element[0].files)
+          RubricService.postGradeFiles(RubricService.grade, element[0].files)
```
zazaian: Do we need to pass `RubricService.grade` here into the RubricService if `grade` is an attribute of the service itself? Maybe this could just be `RubricService.postGradeFiles(element[0].files)`

app/controllers/api/grades/grade_files_controller.rb
```diff
@@ -0,0 +1,15 @@
+class API::Grades::GradeFilesController < ApplicationController
+
+  before_filter :ensure_staff?
+
+  # POST /api/grades/:grade_id/grade_files
+  def create
+    grade = Grade.find(params[:grade_id])
+
+    params[:grade_files].each do |f|
```
zazaian: This doesn't seem too heavy-handed to me, but when I see something like truncating the filename for each file I wonder whether it might be useful to add a service for creating this batch of `GradeFile` records. Thoughts?

zazaian:
Hey @acousticrobot: I think this looks really good. Just a couple of comments on syntax maneuvers. Nice work!

app/assets/javascripts/angular/directives/gradeFileUpload.js.coffee
acousticrobot: great catch thanks!

app/controllers/api/grades/grade_files_controller.rb
acousticrobot: I'm not sure if this limit is still necessary, at one time when this was on the Grade model we were getting errors from long file names.  I agree that this would be better handled elsewhere, possibly `/lib/formatter/filename.rb` but without knowing how to replicate the error I didn't want to refactor.

PR #2282, opened 2016-08-12
2278 assignment planner module

mariehooper:
This PR adds an assignment planner module to the instructor dashboard that shows assignments due in the next 7 days as well as the number of predictions and submissions for each assignment.

screen shot 2016-08-12 at 11 39 40 am

### DIFF
--- a/app/presenters/info/dashboard_course_planner_presenter.rb
+++ b/app/presenters/info/dashboard_course_planner_presenter.rb
@@ -25,7 +25,11 @@ class Info::DashboardCoursePlannerPresenter < Showtime::Presenter
   end
 
   def to_do?(assignment)
-    assignment.include_in_to_do? && assignment.visible_for_student?(student) && !GradeProctor.new(grade_for(assignment)).viewable?
+    if student
+      assignment.include_in_to_do? && assignment.visible_for_student?(student) && !GradeProctor.new(grade_for(assignment)).viewable?
+    else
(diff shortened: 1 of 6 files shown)


chcholman:
@mariehooper Two minor design thoughts that we may want to iterate on as the dashboard becomes complete: 

1) the labels currently look *almost* like buttons. We may want to ask instructors if they tried to click them
2) it gets really tall - we may want to constrain it once we know what the whole grid looks like.

app/presenters/info/dashboard_course_planner_presenter.rb
```diff
@@ -25,7 +25,11 @@ def grade_for(assignment)
   end
 
   def to_do?(assignment)
-    assignment.include_in_to_do? && assignment.visible_for_student?(student) && !GradeProctor.new(grade_for(assignment)).viewable?
+    if student
+      assignment.include_in_to_do? && assignment.visible_for_student?(student) && !GradeProctor.new(grade_for(assignment)).viewable?
+    else
+      assignment.include_in_to_do?
+    end
   end
```
chcholman: Just wanted to note that I think this is an example of how we want to use presenters - we need to show data to both roles, but what bounds that data is challenging and different between them. It wouldn't make sense in the model and it makes the views super messy. I think we want to to keep checking in on whether what we write in presenters fits that.

PR #2291, opened 2016-08-16
Update color palette

mariehooper:
This PR is a first pass at refining the GC color palette


mariehooper:
@chcholman - really love the changes here. I am updating the orange to just a couple shades brighter and want to look at the dark greys (still a bit on a brown side - especially on the predictor headers). I will make those changes and send to you for review!

PR #2293, opened 2016-08-16
1938 change to strong parameters

jwright:
This removes the need for the [protected_attributes](https://github.com/rails/protected_attributes) gem by converting the creation and updating of resources to use [strong parameters](http://edgeguides.rubyonrails.org/action_controller_overview.html#strong-parameters).

The removal of `protected_attributes` was needed since it only supports Rails 4. (see https://github.com/rails/protected_attributes/blob/master/protected_attributes.gemspec#L19)

**IMPORTANT**: This *can* be a scary merge due to the fact that attributes which are not specified within the strong parameters will simply not update the model if they do not exist which may hide some bugs. A lot of the update and creation actions are covered by controller specs but that does not mean that the specs are accurate in representing the true view parameters within the passed in parameters. Proceed with caution.

For #1938 but there is still a lot of work to do on the Rails 5 upgrade. This PR was broken out since it includes a lot of changes.


jwright:
@zazaian This is ready for your review my friend

zazaian:
Hey @jwright -- apologies, I saw this but wasn't quite ready to delve into it. I just finished up some stuff on my branch and will probably delve into this tomorrow morning if that's okay. Cool?

jwright:
@zazaian No rush man. I just wanted to make sure you have seen it. I forgot to tag you in it the first time.

zazaian:
Hey @jwright - I think this all looks fine. It's a pretty straightforward update though as you said any of these might up as a landmine of some sort. I think it's fine to merge but we should keep an eye out for any new system errors that might occur as a result. Nice work!

PR #2299, opened 2016-08-18
1242 preview through sorcery

acousticrobot:
This PR creates a "preview mode" where a faculty is able to view GC as a chosen student from a course.

It does not restrict any assignment submissions, as faculty already have the ability to submit on behalf of a student.

It does restrict:
  * changing the student profile (faculty can view this page, but not update)
  * changing the current course (students could have access to courses not available to faculty)
  * viewing predictions (faculty are not allowed to view student predictions, so they are all at zero)

@chcholman  One consideration: Should we restrict the faculty ability to predict on behalf of the student, since they cannot view this information on refresh? I have left this alone for now.

@jwright please review the code, particularly the management of the preview session.

@mariehooper, please review the styling. To view changes for review:
  * Login as faculty
  * Go to `/students` and review the "preview as this student" button
  * click the button and review the orange top-bar, student's "my account page", and the "exit student preview" link.


chcholman:
@acousticrobot We don't need to restrict their ability to make predictions...but I would like to restrict the ability to _save_ those predictions. So, they might predict 5 assignments and see the resulting grade, but if the student were to load the page, their predictions would still be in the same place. 

Thanks!

chcholman:
@acousticrobot Awesome, thank you!

mariehooper:
@acousticrobot - this looks great to me! Cait and I are updating the color palette so I may update the color of the top bar after that but otherwise 🙌  ✨

app/controllers/api/predicted_earned_badges_controller.rb
```diff
@@ -5,9 +5,12 @@ class API::PredictedEarnedBadgesController < ApplicationController
 
   # GET api/predicted_earned_badges
   def index
-    if current_user_is_student?
+    if current_user_is_student? && !current_user_in_preview_mode?
```
jwright: I like the word `impersonation` because it's not the `current_user` who is in preview mode, it's the current user in preview mode for someone else. What do you think about `current_user_impersonation`? I know it's a bigger change but I think wording is important here. Thoughts?

app/controllers/api/predicted_earned_badges_controller.rb
```diff
@@ -5,9 +5,12 @@ class API::PredictedEarnedBadgesController < ApplicationController
...
       @student = current_student
       @update_badges = true
+    elsif current_user_in_preview_mode?
```
jwright: This could be cleared up a bit:

```
if current_user_is_student?
  @student = current_student
  @update_badges = !current_user_in_preview_mode?
# ...
end
```

app/controllers/api/predicted_earned_challenges_controller.rb
```diff
@@ -5,9 +5,12 @@ class API::PredictedEarnedChallengesController < ApplicationController
 
   # GET api/predicted_earned_challenges
   def index
-    if current_user_is_student?
+    if current_user_is_student?  && !current_user_in_preview_mode?
       @student = current_student
       @update_challenges = true
+    elsif current_user_in_preview_mode?
```
jwright: Same clean up recommended above.

app/controllers/api/predicted_earned_grades_controller.rb
```diff
@@ -5,13 +5,20 @@ class API::PredictedEarnedGradesController < ApplicationController
 
   # GET api/predicted_earned_grades
   def index
-    if current_user_is_student?
+    # restrict predictions for professor viewing student in preview mode
+    if current_user_in_preview_mode?
+      user = User.find(session[:previewing_agent])
```
jwright: I would recommend moving this `session` variable to the `lib/current_scopes` and perhaps changing the key to something with `_id` at the end of it. Perhaps `impersonating_user_id`?

app/controllers/concerns/predictor_data.rb
```diff
@@ -15,7 +15,7 @@ def predictor_badges(student)
       :icon
     ).map do |badge|
       prediction = badge.find_or_create_predicted_earned_badge(student.id)
-      if current_user_is_student?
+      if current_user_is_student? && !current_user_in_preview_mode?
```
jwright: Since we are doing these checks a lot, maybe a shortcut method makes sense? I am not sure what to call it so that may be a sign that it's not needed.

app/controllers/user_sessions_controller.rb
```diff
@@ -45,6 +46,20 @@ def lti_create
     respond_with @user, location: dashboard_path
   end
 
+  def enter_student_preview
```
jwright: If we decide to change the name of the `current_user_in_preview_mode?` then I think we should change the name of these methods to `impersonate_student`, etc.

jwright: I like that these methods are in the `UserSessionsController` 👍

jwright: Do we need to ensure that the `student_id` is an id for an actual student in the current course?

app/controllers/user_sessions_controller.rb
```diff
@@ -45,6 +46,20 @@ def lti_create
...
+    student = User.find(params[:student_id])
+    session[:previewing_agent] = current_user.id
```
jwright: I think we should move this to where the other `current_`* methods are.

app/controllers/user_sessions_controller.rb
```diff
@@ -45,6 +46,20 @@ def lti_create
...
+    auto_login(student)
+    redirect_to root_url
+  end
+
+  def exit_student_preview
+    faculty = User.find(session[:previewing_agent])
```
jwright: I think we should move this to where the other `current_`* methods are.

app/helpers/application_helper.rb
```diff
@@ -9,6 +9,10 @@ module ApplicationHelper
     end
   end
 
+  def current_user_in_preview_mode?
```
jwright: I think we should move this to `lib/current_scopes.rb` or move the current scopes methods to this helper. I like the latter but that may be another PR.

app/views/layouts/navigation/_authenticated_nav.haml
```diff
@@ -1,4 +1,5 @@
-- if current_user_is_student?
+-# do not allow faculty to switch courses in preview mode!
+- if current_user_is_student? && !current_user_in_preview_mode?
```
jwright: We should probably add this check to the controller method as well where the courses are changed so people can't bypass the browser.

spec/controllers/api/predicted_earned_badges_controller_spec.rb
```diff
@@ -62,6 +62,27 @@
     end
   end
 
+  context "as faculty previewing as student" do
+    before do
+      login_user(world.student)
```
jwright: What about having a method in a spec helper called `impersonate_student(student, professor` and use that in the specs since we are duplicating this code in a couple of places? That way we could get rid of the 2 calls to `login_user` and `session=`.

spec/controllers/user_sessions_controller_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+    context "user is not logged in" do
+      it "does not record the course login event" do
+        allow(subject).to receive(:login) { nil }
+        expect(subject).to_not receive(:record_course_login_event)
+        post :create, user: student.attributes
+      end
+    end
+  end
+
+  describe "enter_student_preview" do
```
jwright: 👍

spec/controllers/user_sessions_controller_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+
+    it "logs in as student" do
+      login_user(professor)
+      get :enter_student_preview, student_id: student.id
+      expect(session[:user_id]).to eq(student.id.to_s)
+    end
+  end
+
+  describe "exit_student_preview" do
+    it "returns session to faculty" do
```
jwright: Since we split out the specs above (1 expectation for each spec), I think we should follow that pattern here. 1 spec for `it "removes the professor id from session"` and `it "logs back in as the professor"`.

jwright:
@acousticrobot This is awesome! I just had a few small comments about naming and where the methods live. It looks great and it is very well tested.

app/controllers/api/predicted_earned_badges_controller.rb
acousticrobot: Good point. I think I'm going to go with `impersonating_student?`, I'll look through the rest of the code to make sure that makes sense throughout.

app/controllers/api/predicted_earned_badges_controller.rb
acousticrobot: This is much better, thanks!

app/controllers/api/predicted_earned_grades_controller.rb
acousticrobot: 👍

app/controllers/concerns/predictor_data.rb
acousticrobot: I think it reads better now that it is `current_user_is_student? && !student_impersonation?`

app/controllers/user_sessions_controller.rb
acousticrobot: @ll 3 done ^^^

app/controllers/courses_controller.rb
```diff
@@ -3,6 +3,7 @@ class CoursesController < ApplicationController
 
   skip_before_filter :require_login, only: [:badges]
   before_filter :ensure_staff?, except: [:index, :badges]
+  before_filter :ensure_not_impersonating?, only: [:index]
```
jwright: 👍

spec/controllers/api/predicted_earned_grades_controller_spec.rb
```diff
@@ -52,6 +52,21 @@
     end
   end
 
+  context "as faculty previewing as student" do
+    before do
+      login_user(world.student)
```
jwright: Do we want to use `login_as_impersonating_agent` here?

spec/support/session_helper.rb
```diff
@@ -0,0 +1,6 @@
+module SessionHelper
```
jwright: Maybe `ImpersonationHelper`?

jwright:
@acousticrobot I just had a few comments but everything looks great. I think it's ready for merging after the comments are resolved and it's rebased with master.

PR #2300, opened 2016-08-18
refinements to predictor text input field

acousticrobot:
This PR finesses the predictor input field for a better UX.  

issues addressed:
  * Deleting all text returns "NaN" via SmartNumber
  * The slider only updates to the text input on blur, so the interaction feels unnatural.

I have added separate `on-change` and `on-blur` events to the slider input field. They both trigger the same update sequence, but the Ajax update only happens on blur.

This results in real time "idiot-checking": empty fields and negative numbers revert to a zero, numbers above the threshold revert to the max points.  

Please let me know if any particular change feels too aggressive to be handled in real-time, and we can probably tweak the code to wait until blur to make the change.

### DIFF
--- a/app/assets/javascripts/angular/directives/modelIntConverter.js.coffee
+++ b/app/assets/javascripts/angular/directives/modelIntConverter.js.coffee
@@ -14,6 +14,7 @@
 
       ngModelController.$formatters.push((data)->
         #convert data to string for view
+        return "" if isNaN(data)
         return data.toString()
       )
   }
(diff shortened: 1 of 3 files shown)


mariehooper:
@acousticrobot The NaN fix is great! Not getting that at all anymore. One small thing is just when you click into the side panel to 'click away' from the field, the number doesnt update. Is there a way to fix that? (lower priority)

mariehooper:
I just tried to replicate it again and couldn't get it to happen so looks like it works how we want. One issue I just saw is with the "Standard Edit with Select/Quick Grade with Select [No Grades]" assignment card - it can't be slid and doesn't have a predicted number.

screen shot 2016-08-19 at 8 34 58 am

mariehooper:
@acousticrobot - just checked it. Looks great. That issue is gone.

PR #2302, opened 2016-08-19
Add not null values

chcholman:
This is the next PR to come out of our conversation around cleaning up the database schema in May. It includes dropping two unused tables `tasks` and `student_academic_histories`, removing unused fields where possible, and adding not_null values to appropriate fields. All fields where not_null has been added have been checked against the production db to confirm there are no current offending objects.


jwright:
@chcholman I am unsure why these changes were made. Can you give me some background or context around this PR?

app/models/grade.rb
```diff
@@ -10,8 +10,7 @@ class Grade < ActiveRecord::Base
   belongs_to :student, class_name: "User", touch: true
   belongs_to :team, touch: true
   belongs_to :submission
-  belongs_to :task, touch: true # Optional
-  belongs_to :group, polymorphic: true, touch: true # Optional
```
jwright: 👍  Awesome to see this unnecessary polymorphic `group` go away

db/migrate/20160815135612_remove_character_limits.rb
```diff
@@ -0,0 +1,121 @@
...
+    change_column :submission_files, :filename, :string, limit: nil
+    change_column :submission_files, :filepath, :string, limit: nil
+    change_column :submission_files, :file, :string, limit: nil
+    
+    change_column :submissions, :feedback, :string, limit: nil
+    change_column :submissions, :comment, :string, limit: nil
+    change_column :submissions, :link, :string, limit: nil
+    change_column :submissions, :assignment_type, :string, limit: nil
+    
+    drop_table :tasks
```
jwright: It's odd that this is in the `remove_character_limits` migration. Would it be easy/worth it to move it to it's own migration?

db/migrate/20160815135612_remove_character_limits.rb
```diff
@@ -0,0 +1,121 @@
...
+    
+    change_column :teams, :name, :string, limit: nil
+    change_column :teams, :banner, :string, limit: nil
+    
+    drop_table :themes
```
jwright: Same comment as above

spec/controllers/criteria_controller_spec.rb
```diff
@@ -17,8 +17,11 @@
 
     describe "POST create" do
       it "creates a new criterion" do
-        params = attributes_for(:criterion)
-        expect{ post :create, criterion: params }.to change(Criterion,:count).by(1)
+        @assignment = create(:assignment, course: @course)
```
jwright: Do these (`@assignment` and `@rubric`) need to be instance variables?

jwright:
@chcholman This looks like a *great* cleanup!

I added a few minor comments but no major things.

spec/controllers/criteria_controller_spec.rb
chcholman: Nope!

PR #2317, opened 2016-08-22
Student url cleanup

acousticrobot:
This PR changes the student urls to use more resourceful routing. If no resource exists (ex. "predictor") this route has been moved to the info controller, to match the location of the `/dashboard`

| old route | new route | controller |
| ------ | ------ | ----- |  ----- |
| /predictor | /predictor | info#predictor |
| /grading_scheme | /grade_scheme_elements | grade_scheme_elements#index |
| /syllabus | /assignments | assignments#index |
| /my_team | /teams | teams#index |

I moved the views and presenters to corresponding files, but left the presentation materials unchanged, pending review of these pages for update.

As a general pattern, the index.haml views split presentation between `_student_index` and `_instructor_index` partials to keep the previous presentation intact. this could warrant further review to determine if these views could be dried up or made to further resemble each other.


chcholman:
@acousticrobot This is looking great! Ready to merge, but I would like to make sure the student preview work comes in very rapidly after. Is that ready/possible?

PR #2319, opened 2016-08-22
load first assignment in predictor side panel on init

### LINKED ISSUE #2315: Default to first assignment in GP
[intermediate fix] When you load the grade predictor the sidebar should automatically display the first assignment. 

The next step will be to: 

1) show the tutorial initially
2) record if they've moved past the tutorial
3) auto-display the last assignment they looked at on re-load

acousticrobot:
This PR loads the first assignment from the array into the student side panel. Since the predictor takes some time to load all assets, I have kept the info "hint" in place, until the assignments have loaded.

I noticed that something recently changed on master which has broken assignment types position ordering -- with the sample data "Grade Settings" is no longer the first type to show on the `/syllabus` or `/predictor`. This may through you for a loop during review of this PR, since the first assignment is in this AT. I will submit a separate issue to address this.

screen shot 2016-08-22 at 4 12 10 pm

closes #2315


mariehooper:
@acousticrobot - this looks great to me! I like the hint staying during the load too.

PR #2339, opened 2016-08-25
redirect without flash notice when impersonating

acousticrobot:
@chcholman please test this change and let me know if preview mode still works as expected, and verify that the login notice no longer appears. thanks!


chcholman:
Looks great, thanks @acousticrobot!

PR #2350, opened 2016-08-27
Prediction stats

chcholman:


app/models/user.rb
```diff
@@ -267,6 +267,17 @@ def scores_for_course(course)
...
     }
   end
+  
+  # Checking to see if a student has any positive predictions for a course
+  def predictions_for_course?(course)
+    course.assignments.each do |assignment|
+      if predicted_earned_grades.where(assignment_id: assignment.id).present? && predicted_earned_grades.where(assignment_id: assignment.id).first.predicted_points > 0
+        return true
+      else 
+        return false
```
acousticrobot: I think this is going to always return for the first assignment it checks, and not iterate through them all. The `return false` should be outside of the loop.

However, there are a couple scopes on PEGs that could help. I think this could be changed to:
```
def predictions_for_course?(course)
  predicted_earned_grades.for_course(course).predicted_to_be_done.present?
end
```
Please note that in testing this I found the `for_course` scope to be broken on PEGs, and there is a PR that addresses this: #2355

PR #2354, opened 2016-08-29
Cleanup of CoursesController

acousticrobot:
This PR:    

    * moves change course route onto courses controller
    * adds a course proctor and abilities
    * moves the proctors into the app/proctors directory
    * associate flagged users with course for successful delete
    * changes non-crud course routes to assume current course


app/controllers/courses_controller.rb
```diff
@@ -103,7 +60,24 @@ def create
     end
   end
 
+  def copy
+    duplicated = @course.copy(params[:copy_type], {})
```
jwright: I feel like we should at least authorize this action with a `authorize! :read, @course`. We may want to use a `authorize! :create, Course` as well?

app/controllers/courses_controller.rb
```diff
@@ -118,17 +92,60 @@ def update
   end
 
   def destroy
+    authorize! :destroy, @course
     @name = @course.name
     @course.destroy
 
     respond_to do |format|
```
jwright: Do we need a `respond_to` here? I think we can drop it.

app/controllers/courses_controller.rb
```diff
@@ -118,17 +92,60 @@ def update
...
-        notice: "Course #{@name} successfully deleted"
+        redirect_to courses_url, notice: "Course #{@name} successfully deleted"
       end
     end
   end
 
+  # Switch between enrolled courses
+  def change
+    if course = current_user.courses.where(id: params[:id]).first
+      unless session[:course_id] == course.id
```
jwright: Do we want an `authorize! :read, course` here?

app/proctors/course_proctor.rb
```diff
@@ -0,0 +1,25 @@
...
+class CourseProctor
+
+  attr_reader :course
+
+  def initialize(course)
+    @course = course
+  end
+
+  def viewable?(user)
+    return false if course.nil?
```
jwright: Do we need to check for a nil user here as well?

app/proctors/course_proctor.rb
```diff
@@ -0,0 +1,25 @@
...
+    user.courses.include? @course
+  end
+
+  def updatable?(user)
+    return false if course.nil?
```
jwright: Do we need to check for a nil user here as well?

app/proctors/course_proctor.rb
```diff
@@ -0,0 +1,25 @@
...
+    user.is_professor?(course) || user.is_admin?(course)
+  end
+
+  def destroyable?(user)
+    return false if course.nil?
```
jwright: Do we need to check for a nil user here as well?

config/routes.rb
```diff
@@ -170,27 +170,27 @@
...
-      resources :earned_badges do 
+    get :change, on: :member
+    resources :badges do
+      resources :earned_badges do
         get :confirm_earned, on: :member
       end
     end
   end
+
+  controller :courses do
```
jwright: Why not use `resources` here? Is it because we don't need a `course` id?

spec/active_record_spec_helper.rb
```diff
@@ -27,7 +27,7 @@
 require "sorcery"
 require "yaml"
 require "./lib/display_helpers"
-require "./lib/grade_proctor"
+require "./app/proctors/grade_proctor"
```
jwright: 👍

spec/proctors/condition_sets/submission_file_condition_set_spec.rb
```diff
@@ -1,6 +1,6 @@
 require "proctor"
 require_relative "../../../app/proctors/submission_file_proctor"
-require_relative "../../../app/proctors/condition_sets/submission_file_condition_set"
+require_relative "../../../app/proctors/submission_file_proctor/submission_file_condition_set"
```
jwright: I know this was already here and I am not advocating a change here. I just wanted to note that we should standardize on using `require_relative` or `require` with an app path (like https://github.com/UM-USElab/gradecraft-development/pull/2354/files#diff-3d758429ae20871fda0f63f4621e3a49R1).

I vote for using `require` when the `require_relative` has to move up more than a directory.

jwright:
@acousticrobot This looks great! Thank you for moving the proctors over to app.

I just had a few questions and comments. Nothing big.

app/controllers/courses_controller.rb
acousticrobot: 👍

app/controllers/courses_controller.rb
acousticrobot: I'll add the :create permissions to the proctor and cancan

app/controllers/courses_controller.rb
acousticrobot: added

app/proctors/course_proctor.rb
acousticrobot: good idea, added.

config/routes.rb
acousticrobot: Yes, I've split out the routes w/ course id into `resources :courses do...`, and w/o into `controller :courses do...`

spec/proctors/condition_sets/submission_file_condition_set_spec.rb
acousticrobot: I agree. I remember you asked Mike about it at the time, but I can't remember the reason why he decided to  use `require_relative`.

PR #2374, opened 2016-08-31
1846 - Fix released grades bug

### LINKED ISSUE #1846: Released Grades may not be displayed for students
Released grades do have a possibility for not being displayed for students when the following flow happens:

1. Assignment is marked as release is necessary
1. Instructor updates the score but does not release it
1. Instructor releases the grade

This does not display for the student because there is a `remove` instruction in the `HistoryFilter` to remove `status` changes. Since the raw score is changed when the grade is not viewable to the student, the student never sees the history version where the raw score was updated but the status does not allow for releasing

jwright:
Modify the grade history module to display the grade change version if the grade was released after it was updated.

This displays the version information if the current grade is viewable to the student and the changeset includes a `raw_points` change or a `feedback` change.

Fixes #1846

### DIFF
--- a/app/presenters/submissions/grade_history.rb
+++ b/app/presenters/submissions/grade_history.rb
@@ -27,12 +27,24 @@ module Submissions::GradeHistory
       end
       .rename("SubmissionFile" => "Attachment")
       .include do |history|
+        viewable = true
+
         if history.changeset["object"] == "Grade" && only_student_visible_grades
-          grade = history.version.reify
-          GradeProctor.new(grade).viewable?
-        else
-          true
+          version = history.version.reify
+          viewable = GradeProctor.new(version).viewable?
+
+          unless viewable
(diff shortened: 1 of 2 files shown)


app/presenters/submissions/grade_history.rb
```diff
@@ -20,20 +20,49 @@ def submission_grade_filtered_history(submission, grade, only_student_visible_gr
...
+
+          if !viewable && history_item.version.event == "update"
+            last_raw_points_change = last_change?(history, history_item, "Grade",
+                                                  "raw_points")
+
+            last_feedback_change = last_change?(history, history_item, "Grade",
+                                                  "feedback")
+
+            viewable = (last_raw_points_change || last_feedback_change) &&
+                        GradeProctor.new(grade).viewable?
```
acousticrobot: I like your distillery!

image

PR #2376, opened 2016-09-01
Strip leading and trailing whitespace

### LINKED ISSUE #2097: Spaces after names/emails breaks import
If I import a CSV and there is a space after a student's name or their email address, the whole import breaks.

jonathoy:
When importing users via csv, strip leading and trailing whitespace prior to processing

Fixes #2097


jonathoy:
I added a new fixture file and moved some of the tests into a shared_example group so that they can be performed easily on both the file with erroneous spacing and the one without. I'm curious if there any thoughts or objections to this approach.

chcholman:
@jonathoy Yes, let's update the tests to remove the shared examples. Thank you!

PR #2377, opened 2016-09-01
1519 submit next

acousticrobot:
This PR adds a "Submit and Grade Next" button on the Grade edit form, to allow Professors to cycle through the students needing grading rather than constantly return to the Assignment index page.

screen shot 2016-09-01 at 3 50 42 pm

The logic is a follows: 

If there are submissions for the assignment, "Next" will be the next student, ordered alphabetically, with a submission for the assignment without a grade(status: "Graded").

If the assignment does not accept submissions, "Next" will be the next student, ordered alphabetically, without a grade(status: "Graded")

Both cases return to the index page on the last student.

Running through the students alphabetically and not looping ensures that clicking "S&G Next" when submitting an "In Progress" grade does not return to the current grade edit page.

When editing a "Graded" grade with the form, the button does not appear.

### DIFF
--- a/app/controllers/grades_controller.rb
+++ b/app/controllers/grades_controller.rb
@@ -37,7 +37,18 @@ class GradesController < ApplicationController
         GradeUpdaterJob.new(grade_id: grade.id).enqueue
       end
 
-      path = session[:return_to] || assignment_path(grade.assignment)
+      if params[:redirect_to_grade_next].present?
+        next_submission = grade.assignment.submissions.ungraded.first
+        if next_submission.present?
+          grade = Grade.find_or_create next_submission.assignment.id, next_submission.student.id
+          path =  edit_grade_path grade
+        else
+          path = assignment_path(grade.assignment)
+        end
+      else
+        path = session[:return_to] || assignment_path(grade.assignment)
+      end
+
       redirect_to path,
         notice: "#{grade.student.name}'s #{grade.assignment.name} was successfully updated"
     else # failure
--- a/app/controllers/grades_controller.rb
+++ b/app/controllers/grades_controller.rb
@@ -180,12 +180,7 @@ class GradesController < ApplicationController
   end
 
   def path_for_next_grade(grade)
-    if grade.assignment.accepts_submissions?
-      next_submission = grade.assignment.submissions.ungraded.first
-      next_student = next_submission.student if next_submission.present?
-    else
-      next_student = grade.assignment.next_ungraded_student(grade.student)
-    end
+    next_student = grade.assignment.next_ungraded_student(grade.student)
 
     if next_student.present?
       return edit_grade_path(
--- a/app/controllers/grades_controller.rb
+++ b/app/controllers/grades_controller.rb
@@ -184,7 +184,7 @@ class GradesController < ApplicationController
       next_submission = grade.assignment.submissions.ungraded.first
       next_student = next_submission.student if next_submission.present?
     else
-      next_student = grade.assignment.ungraded_students.first
+      next_student = grade.assignment.next_ungraded_student(grade.student)
     end
 
     if next_student.present?
--- a/app/controllers/grades_controller.rb
+++ b/app/controllers/grades_controller.rb
@@ -37,18 +37,11 @@ class GradesController < ApplicationController
         GradeUpdaterJob.new(grade_id: grade.id).enqueue
       end
 
-      if params[:redirect_to_grade_next].present?
-        next_submission = grade.assignment.submissions.ungraded.first
-        if next_submission.present?
-          grade = Grade.find_or_create next_submission.assignment.id, next_submission.student.id
-          path =  edit_grade_path grade
-        else
-          path = assignment_path(grade.assignment)
-        end
+      if params[:redirect_to_next_grade].present?
+        path = path_for_next_grade grade
       else
-        path = session[:return_to] || assignment_path(grade.assignment)
+        path = assignment_path(grade.assignment)
       end
-
       redirect_to path,
         notice: "#{grade.student.name}'s #{grade.assignment.name} was successfully updated"
     else # failure
@@ -184,5 +177,22 @@ class GradesController < ApplicationController
 
   def rubric_criteria_with_levels
     @rubric_criteria_with_levels ||= @rubric.criteria.ordered.includes(:levels)
+  end
+
+  def path_for_next_grade(grade)
+    if grade.assignment.accepts_submissions?
+      next_submission = grade.assignment.submissions.ungraded.first
+      next_student = next_submission.student if next_submission.present?
+    else
+      next_student = grade.assignment.ungraded_students.first
+    end
+
+    if next_student.present?
+      return edit_grade_path(
+        Grade.find_or_create(grade.assignment.id, next_student.id)
+      )
+    else
+      return assignment_path(grade.assignment)
+    end
   end
 end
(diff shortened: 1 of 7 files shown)


app/controllers/grades_controller.rb
```diff
@@ -174,4 +178,16 @@ def score_recalculator(student)
   def rubric_criteria_with_levels
     @rubric_criteria_with_levels ||= @rubric.criteria.ordered.includes(:levels)
   end
+
+  def path_for_next_grade(grade)
+    next_student = grade.assignment.next_ungraded_student(grade.student)
+
+    if next_student.present?
```
jwright: Rubocop is complaining about this by saying we should use an if statement. I think this is fine code but I am not sure if we want to add a comment to turn that off.

app/models/concerns/gradable.rb
```diff
@@ -64,4 +64,23 @@ def median
...
+    course.students.order_by_name -
+      (User.find(grades.graded.pluck(:student_id)) - User.find(ids_to_include))
+  end
+
+  def ungraded_students_with_submissions(ids_to_include=[])
+    ungraded_students(ids_to_include) & User.find(submissions.pluck(:student_id))
+  end
+
+  def next_ungraded_student(student)
+    if accepts_submissions?
```
jwright: This check couples this concern with the `Assignment` class. I know that this concern is only included in the `Assignment` class but it may prohibit future extractions.

I think this is ok for now but I just wanted to call it out. It's too early to make adjustments for premature extractions.

app/models/concerns/gradable.rb
```diff
@@ -64,4 +64,23 @@ def median
...
+      us = ungraded_students_with_submissions([student.id])
```
jwright: This local variable is really not descriptive enough. I think we should change this to `ungraded_students` and use `self` when calling the method.

app/models/concerns/gradable.rb
```diff
@@ -64,4 +64,23 @@ def median
...
+    else
+      us = ungraded_students([student.id])
+    end
+    i = us.map(&:id).index(student.id)
+    i && i < us.length - 1 ? us[i + 1] : nil
+  end
```
jwright: There is a lot going on in this method. I am concerned about performance. Have you looked at how many database calls this makes? If it makes a lot, we may want to brainstorm on how to make this more performant.

app/views/grades/_standard_edit.html.haml
```diff
@@ -55,6 +55,8 @@
           .submit-buttons
             %ul
               %li= submit_tag "#{grade.persisted? && grade.is_graded? ? 'Update' : 'Submit'} Grade", class: "button"
+              - unless grade.is_graded?
```
jwright: What about graded but not released? Do we care about that status?

In addition, I added a check in PR #2380 that checks to see if a grade has been graded based on the presence of the `graded_by_id`. Should we use that here or do we only care about the status?

app/views/grades/_standard_edit.html.haml
```diff
@@ -55,6 +55,8 @@
...
+                %li= submit_tag "Submit and Grade Next", name: "redirect_to_next_grade", class: "button"
```
jwright: 👍 Love the use of `name` here and in the controller check.

spec/controllers/grades_controller_spec.rb
```diff
@@ -129,10 +129,39 @@
         expect(@grade.reload.score).to eq(nil)
       end
 
-      it "returns to session if present" do
+      it "returns to assignment show page on simple submit" do
         session[:return_to] = login_path
```
jwright: This can be removed now.

spec/controllers/grades_controller_spec.rb
```diff
@@ -129,10 +129,39 @@
...
         put :update, { id: @grade.id, grade: { raw_points: 12345 }}
-        expect(response).to redirect_to(login_path)
+        expect(response).to redirect_to(assignment_path(@grade.assignment))
+      end
+
+      context "when redirect_to_next_grade is included in params" do
+        before do
+          @next_student = create(
+            :student_course_membership, course: @assignment.course,
+            user: create(:user,last_name: "Zzz")).user
```
jwright: Space after comma /

spec/controllers/grades_controller_spec.rb
```diff
@@ -129,10 +129,39 @@
...
 
-      it "returns to session if present" do
+      it "returns to assignment show page on simple submit" do
         session[:return_to] = login_path
         put :update, { id: @grade.id, grade: { raw_points: 12345 }}
-        expect(response).to redirect_to(login_path)
+        expect(response).to redirect_to(assignment_path(@grade.assignment))
+      end
+
+      context "when redirect_to_next_grade is included in params" do
```
jwright: This context is too programmery for BDD. I would vote for something like 

```
context "when the professor chooses to grade the next student" do
```

spec/controllers/grades_controller_spec.rb
```diff
@@ -129,10 +129,39 @@
...
+        before do
+          @next_student = create(
```
jwright: Can we use a `let` here instead of an instance variable?

spec/models/assignment_spec.rb
```diff
@@ -822,20 +822,80 @@
     end
   end
 
-  describe "#soon?" do
-    it "is not soon if there is no due date" do
-      subject.due_at = nil
-      expect(subject).to_not be_soon
+  describe "ungraded_students" do
```
jwright: I  ❤ this spec file.

jwright:
@acousticrobot This looks really great. Awesome functionality for the professors.

I just had some small comments but my main concern was the performance of selecting the next student.

app/models/concerns/gradable.rb
acousticrobot: Yes, this concerned me too, but since the methods dealing with submissions are all on the bloated assignment class, I decided to put it in here with like code.  I agree we should keep an eye on this for the future; once we have something else that is gradable, we can decide if all gradable models should include submission logic, or if this should be handled in another way.

app/models/concerns/gradable.rb
acousticrobot: Sounds good, originally this method passed 80 characters on several lines with a longer var name, but it fits now.

app/models/concerns/gradable.rb
acousticrobot: I would be open to a better way of doing this, I couldn't think of one that didn't compromise the requested functionality. I'll keep it as is for now so professors can use it sooner rather than later, and since this query would be made infrequently and never more than once per request. But I would like to revisit it with an eye to alternatives.

app/views/grades/_standard_edit.html.haml
acousticrobot: What I want to do here is to remove the "next" button when a released grade is changed, since this seemed like a situation that would revolve around a single student, and would make the "next_ungraded_student" even more complicated. I am leaving the next button for "In Progress" grades so that a professor confirming a list of InP grades could still run through them and release them in a single workflow.

spec/controllers/grades_controller_spec.rb
acousticrobot: 👍

spec/controllers/grades_controller_spec.rb
acousticrobot: This entire file is instance variables, so I stuck with the pattern. I think it could and should all be changed to `let` though.

PR #2379, opened 2016-09-02
1532 late submissions

### LINKED ISSUE #1531: Late Submissions
As a student
When I update a submission that is after the assignment is due
Then the submission should be marked as late
So that it communicates it is late to the instructor and student

jonathoy:
This change persists whether a submission is late. If someone creates or updates a submission past the due_at date on the assignment, a confirmation will be presented to the user asking if they want to proceed.

The status of the submission, specifically whether it is late or not, will appear still on the show submission page for students as well as on the grading page for instructors.

Fixes #1531

### DIFF
--- a/app/controllers/submissions_controller.rb
+++ b/app/controllers/submissions_controller.rb
@@ -1,6 +1,7 @@
 class SubmissionsController < ApplicationController
   before_filter :ensure_staff?, only: [:show, :destroy]
   before_filter :save_referer, only: [:new, :edit]
+  after_action :check_and_set_late_status, only: [:create, :update]
 
   def show
     presenter = Submissions::ShowPresenter.new(presenter_attrs_with_id)
@@ -15,21 +16,21 @@ class SubmissionsController < ApplicationController
 
   def create
     assignment = current_course.assignments.find(params[:assignment_id])
-    submission = assignment.submissions.new(submission_params.merge(submitted_at: DateTime.now))
-    if submission.save
+    @submission = assignment.submissions.new(submission_params.merge(submitted_at: DateTime.now))
+    if @submission.save
       redirect_to = (session.delete(:return_to) || assignment_path(assignment))
       if current_user_is_student?
-        NotificationMailer.successful_submission(submission.id).deliver_now if assignment.is_individual?
+        NotificationMailer.successful_submission(@submission.id).deliver_now if assignment.is_individual?
         redirect_to = assignment_path(assignment, anchor: "tab3")
       end
       # rubocop:disable AndOr
       redirect_to redirect_to, notice: "#{assignment.name} was successfully submitted." and return
     end
     render :new, Submissions::NewPresenter.build(assignment_id: params[:assignment_id],
-                                              submission: submission,
-                                              student: submission.student,
+                                              submission: @submission,
+                                              student: @submission.student,
                                               course: current_course,
-                                              group_id: submission.group_id,
+                                              group_id: @submission.group_id,
                                               view_context: view_context)
   end
 
@@ -41,15 +42,15 @@ class SubmissionsController < ApplicationController
 
   def update
     assignment = current_course.assignments.find(params[:assignment_id])
-    submission = assignment.submissions.find(params[:id])
+    @submission = assignment.submissions.find(params[:id])
 
     respond_to do |format|
-      if submission.update_attributes(submission_params.merge(submitted_at: DateTime.now))
-        path = assignment.has_groups? ? { group_id: submission.group_id } :
-          { student_id: submission.student_id }
-        redirect_to = assignment_submission_path(assignment, submission, path)
+      if @submission.update_attributes(submission_params.merge(submitted_at: DateTime.now))
+        path = assignment.has_groups? ? { group_id: @submission.group_id } :
+          { student_id: @submission.student_id }
+        redirect_to = assignment_submission_path(assignment, @submission, path)
         if current_user_is_student?
-          NotificationMailer.updated_submission(submission.id).deliver_now if assignment.is_individual?
+          NotificationMailer.updated_submission(@submission.id).deliver_now if assignment.is_individual?
           redirect_to = assignment_path(assignment, anchor: "tab3")
         end
         format.html { redirect_to redirect_to, notice: "Your submission for #{assignment.name} was successfully updated." }
@@ -59,11 +60,11 @@ class SubmissionsController < ApplicationController
           render :edit, Submissions::EditPresenter.build(id: params[:id],
                                                       assignment_id: params[:assignment_id],
                                                      course: current_course,
-                                                     group_id: submission.group_id,
-                                                     submission: submission,
+                                                     group_id: @submission.group_id,
+                                                     submission: @submission,
                                                      view_context: view_context)
         end
-        format.json { render json: submission.errors, status: :unprocessable_entity }
+        format.json { render json: @submission.errors, status: :unprocessable_entity }
       end
     end
   end
@@ -75,6 +76,10 @@ class SubmissionsController < ApplicationController
   end
 
   private
+
+  def check_and_set_late_status
+    @submission.check_and_set_late_status if current_user_is_student?
+  end
 
   def presenter_attrs_with_id
     base_presenter_attrs.merge id: params[:id]
--- a/app/controllers/submissions_controller.rb
+++ b/app/controllers/submissions_controller.rb
@@ -17,7 +17,7 @@ class SubmissionsController < ApplicationController
     assignment = current_course.assignments.find(params[:assignment_id])
     submission = assignment.submissions.new(submission_params.merge(submitted_at: DateTime.now))
     if submission.save
-      check_and_set_late_status submission
+      submission.check_and_set_late_status!
       redirect_to = (session.delete(:return_to) || assignment_path(assignment))
       if current_user_is_student?
         NotificationMailer.successful_submission(submission.id).deliver_now if assignment.is_individual?
@@ -46,7 +46,7 @@ class SubmissionsController < ApplicationController
 
     respond_to do |format|
       if submission.update_attributes(submission_params.merge(submitted_at: DateTime.now))
-        check_and_set_late_status submission
+        submission.check_and_set_late_status!
         path = assignment.has_groups? ? { group_id: submission.group_id } :
           { student_id: submission.student_id }
         redirect_to = assignment_submission_path(assignment, submission, path)
@@ -77,10 +77,6 @@ class SubmissionsController < ApplicationController
   end
 
   private
-
-  def check_and_set_late_status(submission)
-    submission.check_and_set_late_status!
-  end
 
   def presenter_attrs_with_id
     base_presenter_attrs.merge id: params[:id]
--- a/app/controllers/submissions_controller.rb
+++ b/app/controllers/submissions_controller.rb
@@ -79,7 +79,7 @@ class SubmissionsController < ApplicationController
   private
 
   def check_and_set_late_status(submission)
-    submission.check_and_set_late_status if current_user_is_student?
+    submission.check_and_set_late_status!
   end
 
   def presenter_attrs_with_id
--- a/app/controllers/submissions_controller.rb
+++ b/app/controllers/submissions_controller.rb
@@ -1,7 +1,6 @@
 class SubmissionsController < ApplicationController
   before_filter :ensure_staff?, only: [:show, :destroy]
   before_filter :save_referer, only: [:new, :edit]
-  after_action :check_and_set_late_status, only: [:create, :update]
 
   def show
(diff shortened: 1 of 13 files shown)


app/views/submissions/_form.html.haml
```diff
@@ -37,5 +37,7 @@
 
   .submit-buttons
     %ul
-      %li= f.button :submit, "#{presenter.submission.persisted? ? (presenter.submission.will_be_resubmitted? ? "Resubmit" : "Update Submission") : "Submit #{term_for :assignment}"}"
+      %li= f.button :submit,
+        "#{presenter.submission.persisted? ? (presenter.submission.will_be_resubmitted? ? "Resubmit" : "Update Submission") : "Submit #{term_for :assignment}"}",
+        data: { confirm: ("This submission will be late. Continue?" if presenter.submission_will_be_late?) }
```
acousticrobot: Nice use of the presenter here!

app/controllers/submissions_controller.rb
```diff
@@ -15,21 +16,21 @@ def new
 
   def create
     assignment = current_course.assignments.find(params[:assignment_id])
-    submission = assignment.submissions.new(submission_params.merge(submitted_at: DateTime.now))
-    if submission.save
+    @submission = assignment.submissions.new(submission_params.merge(submitted_at: DateTime.now))
```
jwright: Since we are not using the `submission` directly in the view here, we should not change this to an instance variable. This eliminates some confusion by the developer to show that we are not using it as an instance variable and we get to it through the presenter.

app/controllers/submissions_controller.rb
```diff
@@ -41,15 +42,15 @@ def edit
 
   def update
     assignment = current_course.assignments.find(params[:assignment_id])
-    submission = assignment.submissions.find(params[:id])
+    @submission = assignment.submissions.find(params[:id])
```
jwright: Same instance variable comment as above.

app/controllers/submissions_controller.rb
```diff
@@ -1,6 +1,7 @@
 class SubmissionsController < ApplicationController
   before_filter :ensure_staff?, only: [:show, :destroy]
   before_filter :save_referer, only: [:new, :edit]
+  after_action :check_and_set_late_status, only: [:create, :update]
```
jwright: I feel like this adds unnecessary complexity in the controller. For one, we now need to make it an instance variable to share state in the correct controllers. I feel like placing this after the `save` or `update_attributes` respectively would be easier to follow and reduce complexity.

app/models/submission.rb
```diff
@@ -86,8 +86,11 @@ def name
   end
 
   # Checking to see if a submission was turned in late
-  def late?
-    created_at > self.assignment.due_at if self.assignment.due_at.present?
+  # Set while skipping validations and callbacks
+  def check_and_set_late_status
+    if self.assignment.due_at.present?
+      self.update_column(:late, submitted_at > self.assignment.due_at)
```
jwright: Why do we need to store this state in the database if we can just check the attributes and have a `late?` method return the result?

app/presenters/submissions/new_presenter.rb
```diff
@@ -13,5 +13,9 @@ def student
     def title
       "Submit #{assignment.name} (#{view_context.points assignment.full_points} #{"point".pluralize(assignment.full_points)})"
     end
+
+    def submission_will_be_late?
+      assignment.due_at.present? ? DateTime.now > assignment.due_at : false
```
jwright: This can be simplified as `assignment.due_at.present? && assignment.due_at < DateTime.now`

db/schema.rb
```diff
@@ -341,6 +341,7 @@
     t.integer  "assignment_id"
     t.integer  "level_id"
     t.boolean  "student_visible", default: false, null: false
+    t.boolean  "shared",          default: false, null: false
```
jwright: Where did this come from? I don't see a migration.

app/models/submission.rb
```diff
@@ -86,8 +86,11 @@ def name
   end
 
   # Checking to see if a submission was turned in late
-  def late?
-    created_at > self.assignment.due_at if self.assignment.due_at.present?
+  # Set while skipping validations and callbacks
+  def check_and_set_late_status
```
jwright: This should probably be named with a bang (e.g. `check_and_set_late_status!`) since it has side affects

spec/controllers/submissions_controller_spec.rb
```diff
@@ -101,6 +101,12 @@
         post :create, assignment_id: @assignment.id, submission: params
         expect(response).to render_template :new
       end
+
+      it "calls check_and_set_late_status afterwards" do
```
jwright: 👍 

This BDD is a little too programmery. Perhaps `it "checks if the submission is late"`?

spec/controllers/submissions_controller_spec.rb
```diff
@@ -112,6 +118,13 @@
         expect(response).to redirect_to(assignment_submission_path(@assignment, @submission, student_id: @student.id))
         expect(@submission.reload.text_comment).to eq("Ausgezeichnet")
       end
+
+      it "calls check_and_set_late_status afterwards" do
+        params = attributes_for(:submission)
+        params[:text_comment] = "Gesundheit"
```
jwright: Why do we need this in the spec? It may cause confusion as to it being relevant to the outcome.

spec/controllers/submissions_controller_spec.rb
```diff
@@ -112,6 +118,13 @@
         expect(response).to redirect_to(assignment_submission_path(@assignment, @submission, student_id: @student.id))
         expect(@submission.reload.text_comment).to eq("Ausgezeichnet")
       end
+
+      it "calls check_and_set_late_status afterwards" do
```
jwright: Same programmery BDD comment as above.

spec/controllers/submissions_controller_spec.rb
```diff
@@ -152,6 +165,13 @@
         submission = Submission.unscoped.last
         expect(submission.submitted_at).to be > current_time
       end
+
+      it "calls check_and_set_late_status afterwards" do
```
jwright: The `after_action` is not set to work on `destroy`. Is this passing?

spec/models/submission_spec.rb
```diff
@@ -3,6 +3,7 @@
 require "toolkits/sanitization_toolkit"
 require "support/uni_mock/rails"
 require "formatter"
+require "byebug"
```
jwright: Do we need this require?

spec/models/submission_spec.rb
```diff
@@ -309,17 +310,23 @@
...
-      assignment = create(:assignment, due_at: Date.today - 1)
-      submission = create(:submission, assignment: assignment)
-      expect(submission.late?).to eq(true)
-    end
-
-    it "returns false if the submission was created before the due date" do
-      assignment = create(:assignment, due_at: Date.today + 1)
-      submission = create(:submission, assignment: assignment)
-      expect(submission.late?).to eq(false)
+  describe "#check_and_set_late_status" do
```
jwright: 👍

spec/presenters/submissions/edit_presenter_spec.rb
```diff
@@ -108,4 +108,31 @@
       expect(subject.title).to eq "Editing My Submission for Big assignment"
     end
   end
+
+  describe "#submission_will_be_late?" do
```
jwright: Really well written specs here!

spec/presenters/submissions/new_presenter_spec.rb
```diff
@@ -1,6 +1,7 @@
 require "active_support/inflector"
```
jwright: We can remove this require since we are requiring all of rails now.

spec/presenters/submissions/new_presenter_spec.rb
```diff
@@ -1,6 +1,7 @@
...
-require "spec_helper"
+require "rails_spec_helper"
 require "./app/presenters/submissions/new_presenter"
```
jwright: We can remove this require since we are requiring all of rails now.

spec/presenters/submissions/new_presenter_spec.rb
```diff
@@ -1,6 +1,7 @@
...
+require 'date'
```
jwright: This require is not needed.

jwright:
@jonathoy I added several comments. The testing is great!

I am wondering why we need to store the `late` flag in the database if we can just check `submitted_at > self.assignment.due_at` within a `late?` method.

db/schema.rb
jonathoy: There is a migration included in commit 60f2a7a592620fa4886ebeeba932f6d096681216

spec/controllers/submissions_controller_spec.rb
jonathoy: This test lives within the `describe "POST create"`

spec/models/submission_spec.rb
jonathoy: My mistake on this one

jonathoy:
Thanks for the feedback guys! I think that I wasn't all that clear on the conventions for using instance variables since I saw various implementations going on elsewhere in the app. I went ahead and removed the after_action filter and made explicit calls in create and update to check and set the lateness of a submission.

Just to clarify some of the questions regarding this PR, @chcholman wanted to persist whether a submission is late due to some datetime complexities. She also mentioned that for now, we should update the attribute regardless of whether the user is a student or not.

db/schema.rb
jwright: @jonathoy Sorry. I am referring to the new `shared` column for `earned_badges`. There is no migration for that.

jwright:
@chcholman Can you expand on the datetime complexities?

I see a need for the `late` flag being stored if we need to query on it from a database point of view but I am not seeing the need for a datetime issue unless the `submitted_at` or the `assignment.due_at` can change but we still want it marked as late. Thoughts?

app/controllers/submissions_controller.rb
```diff
@@ -17,6 +17,7 @@ def create
     assignment = current_course.assignments.find(params[:assignment_id])
     submission = assignment.submissions.new(submission_params.merge(submitted_at: DateTime.now))
     if submission.save
+      check_and_set_late_status submission
```
jwright: I think the separate private method here adds some unnecessary complexity since the method contains one line anyway and you pass in the submission.

```
submission. check_and_set_late_status!
```

app/controllers/submissions_controller.rb
```diff
@@ -45,6 +46,7 @@ def update
 
     respond_to do |format|
       if submission.update_attributes(submission_params.merge(submitted_at: DateTime.now))
+        check_and_set_late_status submission
```
jwright: Same as above.

app/models/submission.rb
```diff
@@ -86,8 +86,11 @@ def name
   end
 
   # Checking to see if a submission was turned in late
-  def late?
-    created_at > self.assignment.due_at if self.assignment.due_at.present?
+  # Set while skipping validations and callbacks
+  def check_and_set_late_status!
+    if self.assignment.due_at.present?
```
jwright: We may want to change this to use a guard clause:

```
def check_and_set_late_status!
  return false if self.assignment.due_at.nil?
  self.update_column(:late, submitted_at > self.assignment.due_at)
end
```

app/presenters/submissions/new_presenter.rb
```diff
@@ -13,5 +13,9 @@ def student
     def title
       "Submit #{assignment.name} (#{view_context.points assignment.full_points} #{"point".pluralize(assignment.full_points)})"
     end
+
+    def submission_will_be_late?
```
jwright: This method should be moved to the base `Submissions::Presenter` to remove the duplication.

jwright:
@jonathoy Really great improvements! I added a few additional small things. This is really taking shape.

I have a question out for @chcholman and the brand needs to be rebased.

chcholman:
@jwright Sorry for missing this! There's a couple of things I'm thinking about: 

* If we import submissions from another system, then the `created_at` and `updated_at` dates would be for GradeCraft but not reflective of _late_ persay. 

* There are scenarios where we're going to want to visualize all late assignments, or students who have turned in late assignments, and this seems like it makes that much more efficient as a query. 

* There's a future feature that involves personalized due dates - we might each have our own start times and due times for the same assignment. This can also be queried for the calculation, but might involve 300 separate references to determine the ontime/late statuses for a single assignment in a large class.

jwright:
@chcholman I think this solution is great for those scenarios above. 

I think this was a little pre-mature since this PR does not solve those issues. An issue can arise where those features are never implemented but we now have this unnecessary (so far) code to support.

I think we should not scrap it since it's built but maybe we want to up the priority of the features you mentioned above? Not sure about the current priority.

I am passing this back to you for merging! Nice work @jonathoy

PR #2380, opened 2016-09-02
1970 record grade modifier

### LINKED ISSUE #1970: Record the Grade Modifier
As an administrator
I would like to know who modified a grade
So that I can see if it's a student (student logged) or a faculty member

jwright:
This replaces the need for a `Grade#instructor_modified` column by just using the `graded_by_id` instead. This eliminates the need for two columns to be updated when a `Grade` is graded.

A method `grade_modified?` was added to `Grade` to be used for checking the existence of `graded_by_id` which replaces the `instructor_modified?` check.

Closes #1970

### DIFF
--- a/app/controllers/info_controller.rb
+++ b/app/controllers/info_controller.rb
@@ -37,7 +37,7 @@ class InfoController < ApplicationController
   # submissions in the system
   def grading_status
     @title = "Grading Status"
-    grades = current_course.grades.instructor_modified
+    grades = current_course.grades.grade_modified
     submissions = current_course.submissions.includes(:assignment, :grade, :student, :group, :submission_files)
     @ungraded_submissions_by_assignment = submissions.ungraded.group_by(&:assignment)
     @resubmissions_by_assignment = submissions.resubmitted.group_by(&:assignment)
--- a/app/exporters/grade_exporter.rb
+++ b/app/exporters/grade_exporter.rb
@@ -18,7 +18,8 @@ class GradeExporter
       csv << headers + detail_headers
       students.each do |student|
         grade = student.grade_for_assignment(assignment)
-        grade = NullGrade.new if grade.nil? || !(grade.instructor_modified? || grade.graded_or_released?)
+        grade = NullGrade.new if grade.nil? || !(grade.grade_modified? ||
+                                                 grade.graded_or_released?)
         submission = student.submission_for_assignment(assignment)
         csv << [student.first_name, student.last_name,
                 student.email,
--- a/db/samples/assignments.rb
+++ b/db/samples/assignments.rb
@@ -47,7 +47,6 @@
   # only used if :grades is true:
   grade_attributes: {
     raw_points: -> { rand(5000) },
-    instructor_modified: false,
     status: nil,
     feedback: nil,
     excluded_from_course_score: false
@@ -122,7 +121,6 @@ uncertainty! - Douglas Adams"
   grades: true,
   grade_attributes: {
     status: "Graded",
-    instructor_modified: true
   }
 }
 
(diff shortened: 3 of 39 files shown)


db/samples/assignments.rb
```diff
@@ -701,7 +688,6 @@
   },
   grades: true,
   grade_attributes: {
-    instructor_modified: false,
     raw_points: -> { nil },
```
acousticrobot: I'm not sure what we were testing with these sample grades where `instructor_modified: false`.  I'm going to go it and cleanup them up because I think it may have all been from when predictions were stored on grades, as a way to make sure the samples had grades available to save predictions onto.  From what you saw during this refactor, is there any situation that would be useful for testing by having a sample grade with a nil `graded_by_id`?

app/controllers/info_controller.rb
```diff
@@ -39,7 +39,7 @@ def earned_badges
   # submissions in the system
   def grading_status
     @title = "Grading Status"
-    grades = current_course.grades.instructor_modified
+    grades = current_course.grades.grade_modified
```
acousticrobot: Could this just be `grades = current_course.grades` now?

app/exporters/grade_exporter.rb
```diff
@@ -18,7 +18,8 @@ def export_grades_with_detail(assignment, students, options={})
       csv << headers + detail_headers
       students.each do |student|
         grade = student.grade_for_assignment(assignment)
-        grade = NullGrade.new if grade.nil? || !(grade.instructor_modified? || grade.graded_or_released?)
+        grade = NullGrade.new if grade.nil? || !(grade.grade_modified? ||
+                                                 grade.graded_or_released?)
```
acousticrobot: Could this be simplified to `grade = NullGrade.new if grade.nil? || !grade.grade_modified?`? I'm wondering what condition this was testing for before and is it still relevant, for instance self logged grades which will now be true for `grade_modified?`. This may be a question for @chcholman.

db/samples/assignments.rb
jwright: I guess this brings up the question of how do we decipher between grades and predicted grades? Were we using `instructor_modified` to tell that before?

app/controllers/info_controller.rb
jwright: If they are all going to have a `graded_by_id` we can. It hinders on the question I have above.

app/exporters/grade_exporter.rb
jwright: Well, a grade can be graded by someone but still have an `In Progress` status.

jwright:
We are treating `graded_by_id` as the same as the previous `graded_by_id` and `instructor_modified`. Are there situations where we want to check that the `graded_by_id` is actually an instructor? What are the use cases for this?

chcholman:
@acousticrobot Is is indeed all legacy, and should be removed - unless it would be useful in any way to address the workflow we were navigating yesterday.

jwright:
@acousticrobot Sounds great.

I have a few further questions now.

1. Does the predictor no longer create all the grades?
2. If it does not create the grades, does it now use a `NilGrade`? 

I am confused by the `grade.nil?` comment. Are you saying that a `Grade` can come from the database but still return true for some reason for `nil?`?

jwright:
@acousticrobot Ok. I am going to review the predictor a bit more but I am pretty convinced that we no longer need a `grade_modified?` predicate method nor the `grade_modified` scope since I believe all grades will have a `graded_by_id`

jwright:
I believe this story needs further definition/requirements and so I am putting this on hold until further discussion is had around this. I think this PR in it's current state introduces new issues and does not solve the actual problem that #1970 is addressing.

jwright:
@acousticrobot @chcholman Per our discussion on 9/13, there will no longer be any grades that do not have a `graded_by_id` and therefore, we can remove any code that treats grades as ungraded because they all will be. Is that accurate? Let me know and I will remove those places that @acousticrobot pointed out in the conversation.

jwright:
Blocked due to missing requirements on how grades should be created overall

PR #2389, opened 2016-09-07
2172 refresh canvas assignments

### LINKED ISSUE #2172: Refresh Canvas assignments
As a professor
I should be able to update my GradeCraft Assignment that was imported from Canvas
So I can capture any changes that occurred on Canvas

jwright:
This allows any imported assignment to be "synced" with it's provider so that changes that occurred on the provider (e.g. canvas) are updated with the assignment in GradeCraft.

A few additional table changes to `imported_assignments` were needed.

1. A `provider_data` was added to specify additional data needed to retrieve the assignment on the canvas site. `course_id` is needed to retrieve a single assignment and so that is stored.
2. A `last_imported_at` timestamp was added to mark the date/time an assignment was last imported. This will be used when "syncing" the data back and forth.

Closes #2172

### DIFF
--- a/app/importers/assignment_importers/canvas_assignment_importer.rb
+++ b/app/importers/assignment_importers/canvas_assignment_importer.rb
@@ -45,6 +45,7 @@ class CanvasAssignmentImporter
       provider_resource_id: provider_resource_id)
     imported.assignment = assignment
     imported.provider_data = provider_data
+    imported.last_imported_at = DateTime.now
     imported.save
   end
 end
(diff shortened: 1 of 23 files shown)


app/importers/assignment_importers/canvas_assignment_importer.rb
```diff
@@ -31,12 +29,23 @@ def import(course, assignment_type_id)
     self
   end
 
+  def self.import_row(assignment, canvas_assignment)
```
acousticrobot: 👍

PR #2396, opened 2016-09-07
predictor sliders with relaxed snapping

### LINKED ISSUE #2382: Typing numbers in the grade predictor fails if levels are present
This happens only before you've set enough points to actually reach a level. Once you have (for instance by dragging the slider), then you can type. Until then, everything is zeroed out as you type.

acousticrobot:
### The Problem:

fail

There appeared to be a bug on production where a level slider would not progress past zero unless the slider was first pulled over to the right.

We found this particular assignment had a level set at zero. With the live snapping in place on the slider input, it would drag the slider back down to zero before enough digits could be entered to pass the threshold for this level value.

I found this was also the behavior on a level with a threshold -- the threshold would in effect snap a value back to zero before it could be fully entered.

### The Solution: 

This PR relaxes the snapping checks that are in place during a live update to an input field.  Min and Max snapping still occurs, but level and threshold checks are only run once the user clicks out of the field.

pass

Please check that this interaction feels intuitive, and still provides enough feedback to the user while typing. 

closes #2382

### DIFF
--- a/app/assets/javascripts/angular/directives/predictor/slider.js.coffee
+++ b/app/assets/javascripts/angular/directives/predictor/slider.js.coffee
@@ -127,17 +127,20 @@
         else
           return points
 
-      scope.snap = (points)->
+      # Only snaps that we want to fire while live updates
+      # are coming from the input field
+      scope.liveSnap = (points)->
+        points = scope.pointsSnappedToRange(points)
+        scope.updateArticle(points)
+        scope.updateSlider(points)
+
+      scope.snapAndSave = (points)->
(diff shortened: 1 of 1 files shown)


mariehooper:
@acousticrobot If I click into the text field and then click away on an assignment that has levels, it doesn't automatically snap to the next closest level. However, it doesn't snap back to 0 which sounds like it was the original issue. Is that correct?

mariehooper:
I gotcha! It looks good to me considering the level of complication of this interaction!

Issue #2174, opened 2016-06-28
Sync Canvas assignments


ksofa2:
As an instructor
I should be able to have an assignment automatically update Canvas when the GradeCraft details change and vice versa
So that I do not have to worry about making sure both systems are in sync

jwright:
Depends on #2173

PR #2398, opened 2016-09-08
2173 update canvas assignments

### LINKED ISSUE #2173: Update Canvas assignments
As a professor
I should be able to update my Canvas Assignment that was imported from Canvas with GradeCraft data
So I can sync any changes with Canvas

jwright:
This PR allows an assignment to be updated on Canvas from GradeCraft. This is in preperation of the syncing task (#2174). Since we store when an assignment was last updated and when an assignment was last imported, we can now sync any changes between GradeCraft and Canvas.

The buttons on the `assignment#show` view now have the ability to update from Canvas or update to Canvas.

screen shot 2016-09-08 at 10 05 16 am

I am not crazy about this interface. I am starting to think that we need a new tab for imported assignments. This would allow the user to view when the last time an assignment was imported and the last time it was exported to Canvas. They would also be able to turn on/off sync when #2174 was completed. I'd love some thoughts around this.

Closes #2173

### DIFF
--- a/app/controllers/assignments/importers_controller.rb
+++ b/app/controllers/assignments/importers_controller.rb
@@ -63,6 +63,13 @@ class Assignments::ImportersController < ApplicationController
     redirect_to assignment_path(assignment)
   end
 
+  # POST /assignments/importers/:importer_provider_id/assignments/:id/update
+  def update_assignment
+    assignment = Assignment.find(params[:id])
+
+    redirect_to assignment_path(assignment)
+  end
+
   private
 
   def syllabus
--- a/app/controllers/assignments/importers_controller.rb
+++ b/app/controllers/assignments/importers_controller.rb
@@ -65,7 +65,17 @@ class Assignments::ImportersController < ApplicationController
 
   # POST /assignments/importers/:importer_provider_id/assignments/:id/update
   def update_assignment
+    provider_name = params[:importer_provider_id]
     assignment = Assignment.find(params[:id])
+
+    result = Services::ImportsLMSAssignments.update provider_name,
+      authorization(provider_name).access_token, assignment
+
+    if result.success?
+      flash[:notice] = "You have successfully updated #{assignment.name} on #{provider_name.capitalize}"
+    else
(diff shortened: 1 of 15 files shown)


chcholman:
@mariehooper and I just talked about this - we're in agreement on creating a "Canvas Sync" tab, and removing the two buttons from the nav, with the tab content display the info Jamie described. 

It would be ideal to add a 'preview changes' step in both the import and export process (this is universal - we should add this to the grade import process as well). Let's talk about that part more tomorrow!

chcholman:
@jwright @mariehooper Is going to take a crack at the info design needed here and will share that tomorrow!

mariehooper:
@jwright  - after much discussion, here is the workflow that @chcholman and I have landed on for this piece.

**Course Setup - occurs in form under ‘Course Setup’ heading (these pages will be paginated later under one link)**
- Authenticate: show Canvas integration card with button and explanation
- Identify course to maintain persistent connection to
- Be able to unlink if necessary

**Importing grades**
- If the GC assignment has a Canvas id, it should give you the list of grades to import
- If it doesn’t, it should give you a list of assignments to select ONE from to import, and then store that connection
- There needs to be a way to unlink the assignment between GC and Canvas

**After assignment is linked, show table with the below modifications:**
- Remove fields:
   - Drop course, drop email, drop graded at
   - Retitle Assignment to “Integration + assignment”
   - Add feedback for both
   - Add and identify GC Score
- New request:
    If someone marks a non-gradecraft match as to be imported, create an account for them

Assignments page needs a new tab, 'Canvas Sync'
screen shot 2016-09-19 at 3 01 12 pm

jwright:
@chcholman I am wondering if this should be a new PR/issue and we merge this one? There are a lot of changes with the workflow here that affects assignments and grades so I wonder if we create a new issue and PR. Thoughts?

chcholman:
@jwright Totally agree! Merging now.

jwright:
#2453 is the new issue to fulfill the requirements above

PR #2421, opened 2016-09-09
1865 student rubric redesign

mariehooper:
This PR restyles the student facing rubric feedback to be presented in a tab-panel design that addresses feedback received in the Spring semester usability testing. The mobile experience is improved by placing the focus on levels earned per criteria. Also included in this PR is a slider to hide the "x students earned this level" per feedback from students in usability testing.

This PR covers student rubric grade earned pages, instructor viewing student's rubric grade and instructor viewing empty rubric shell on rubric graded assignment. 

## Next Steps:

1. The mobile experience will be further improved with the addition of a slider to allow access to all levels of each criterion, not just the one you earned.
2. The instructor editing experience will be improved and updated to match the look and feel of the student rubric design.
3. Deprecated styles will be removed and styles for student rubric will be further condensed with redesign of instructor rubric editing

## Before
screen shot 2016-09-09 at 8 49 03 am

## After
screen shot 2016-09-09 at 8 43 39 am


app/assets/javascripts/assignments.js
```diff
@@ -165,3 +165,37 @@
...
+});
+
+// for student rubric feedback on mobile
+function rubricScreenSize() {
+  if ($('.level-tab').css('display') === 'none') {
+    $('.level-tab.earned').trigger('click');
+  }
+}
+rubricScreenSize();
+$(window).resize(rubricScreenSize);
```
ksofa2: Can this be done in CSS?

app/assets/javascripts/assignments.js
```diff
@@ -165,3 +165,37 @@
...
+$('#class-analytics-toggle').change(function(){
+  if ($(this).prop("checked")) {
+    $('.graded-students').show();
+  } else {
+    $('.graded-students').hide();
+  }
+});
+
+// for student rubric feedback on mobile
+function rubricScreenSize() {
```
ksofa2: It would be helpful to have a comment here about the reasons for this function and its behavior.

app/assets/javascripts/assignments.js
mariehooper: As far as I am aware, there is no way to simulate a click event when the screen is resized than with js :) this was required due to the fact that someone could easily have clicked a tab other than the level they earned before resizing the screen.

app/views/rubrics/components/_rubric_table.haml
```diff
@@ -1,16 +1,13 @@
...
-      - if include_grade_info
-        %td.comments.heading
-          Comments
-
-  %tbody
+  #student-rubric-feedback
     = render partial: "rubrics/components/criteria", locals: { rubric: rubric, student: student, include_grade_info: include_grade_info }
+- else
+  #rubric-shell
+    = render partial: "rubrics/components/rubric_shell", locals: { rubric: rubric, student: student, include_grade_info: include_grade_info }
```
mariehooper: I did this because the html is pretty different for these two and it relieved a bit of the logic required that was cluttering up `components/criteria`

app/assets/javascripts/assignments.js
ksofa2: I was suggesting avoiding the need to simulate a click, maybe something with media queries, but that might be more complex than necessary.

You mentioned that a next step will be to give access to the non-earned levels on mobile, that may eliminate the need for taking action on resizing.

app/views/rubrics/components/_rubric_table.haml
ksofa2: I see. I think the partials' names are throwing me off. `criteria` is used to display the rubric when there is a grade, and `rubric_shell` when there is no grade?

Would it make sense to call `rubric_shell` something like `criteria_without_grade`?

mariehooper: Definitely - will make that change!

ksofa2: Cool. Another way to rename (this might be less confusing):

`criteria` -> `criteria_graded`
`rubric_shell` -> `criteria`

This would be a slight reversal of your original names, but I think less confusing than what I just proposed.

mariehooper:
Thanks for the feedback @ksofa2 - please check out the incorporated changes.

app/assets/stylesheets/_rubrics.sass
```diff
@@ -327,6 +327,192 @@ button.meets-expectations-status.no-expectation
...
+  display: inline-block
+  width: 40px
+  height: 24px
+
+/* Hide default HTML checkbox */
+.switch input 
+  display: none
+
+/* The slider */
+.slider
```
acousticrobot: On the predictor, we have binary `switch` for pass/fail - similar to what you have here I believe, and a `slider` which handles a range of values.  If we could it would be great to standardize on the naming and styles between these two uses of a switch and keep them separate from sliders.  It would be even better if we could control your new switch with an Angular directive instead of jquery, either modifying or building new based on the existing one: `/angular/directives/predictor/switch.js.coffee`

ksofa2: I'm OK with keeping this as jquery for now. I agree it makes sense to align with the naming used in the pass-fail switch.

`.switch` --> `.rubric-switch-socket`
`.switch-slider` --> `.rubric-switch-handle`

(Is that right?)

I think this element should be available across the application, and probably be the same element that the predictor uses (so my suggestion to use the .rubric- prefix is mainly to avoid confusion until we do that). I agree that it would be nice to have consistency and use Angular for this, but don't think it has to happen in this PR.

mariehooper: @ksofa2 done.

db/schema.rb
```diff
@@ -305,10 +305,10 @@
     t.string   "name"
```
chcholman: This file should be removed from the commit (I believe it was added extraneously).

mariehooper:
@chcholmam correct, it shouldn't be there. I had weirdness rebasing. How can I remove those changes?

chcholman:
@mariehooper try: `git checkout origin/master -- db/schema.rb`

Issue #2181, opened 2016-06-30
Improve responsiveness of ribbon/context menu


mariehooper:
instructor side menu with options - convert to dropdown, cull down list of options, etc

PR #2431, opened 2016-09-15
2181-context-menu-improvements

mariehooper:
This PR is intended to improve the consistency and responsiveness of the context_menu (ribbon menu on instructor side) by eliminating several options and moving to only use this menu for actions you can take on a page instead of navigation. On the Analytics tab, a select menu was included in order to function as navigation.

Closes issue #2181


app/assets/stylesheets/_context_menu.sass
```diff
@@ -9,39 +9,36 @@
...
       margin-right: -4px
       text-transform: capitalize
-      padding: .5rem .4rem
       font-size: .7rem
       display: inline-block
       border-right: 1px solid darken($color-blue-3, 20%)
       color: $color-blue-2
       border-radius: 0 !important
+      @media (max-width: 1200px)
+        font-size: .6rem
```
chcholman: Feels like we should become consistent about using variables or specified widths to constrain size - my preference is for variables, how about yours?

app/views/analytics/_in_page_nav.haml
```diff
@@ -0,0 +1,5 @@
+%select#analytics-switcher
+  = analytics_switcher_option(analytics_students_path, 'Analytics', 'student')
```
chcholman: Just noting that I'm curious if we'll end up abstracting this to be any page switcher

PR #2447, opened 2016-09-20
remove create grades dependency for predictor

acousticrobot:
This PR removes the need for grades to be created for predictor display. I would like to remove the need for Null objects entirely (handle json without nested objects from within the front end) but while we have them this is the easiest way to accomplish this.

### DIFF
--- a/app/serializers/predicted_assignment_serializer.rb
+++ b/app/serializers/predicted_assignment_serializer.rb
@@ -15,7 +15,9 @@ class PredictedAssignmentSerializer < SimpleDelegator
   def grade
     if @grade.nil?
       if student.present?
-        grade = Grade.find_or_create(assignment.id, student.id)
+        grade = Grade.where(
+          assignment_id: assignment.id,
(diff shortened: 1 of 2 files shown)


jwright:
@acousticrobot This looks great but there are several other areas where `Grade.find_or_create` is called. Can we remove those instances as well?

https://github.com/UM-USElab/gradecraft-development/blob/59d54585a041fe11a0d7ceeae2070adf707eeddd/app/controllers/assignments/students_controller.rb#L6

https://github.com/UM-USElab/gradecraft-development/blob/455f0e686e5e37f737b74563068311765a30728c/app/services/creates_grade/builds_grade.rb#L13

https://github.com/UM-USElab/gradecraft-development/blob/7eaa255e0b16627fed7ddcde49ecf1acae662bbd/app/controllers/api/grades_controller.rb#L7

https://github.com/UM-USElab/gradecraft-development/blob/59d54585a041fe11a0d7ceeae2070adf707eeddd/app/controllers/grades_controller.rb#L186

PR #2449, opened 2016-09-20
Title refactor

chcholman:
This PR removes all titles and breadcrumbs from the individual views and centralizes them to single files. 

Still to do: 
- [x] Integrate with object names 
- [x] incorporate custom_terms


db/schema.rb
```diff
@@ -658,6 +677,13 @@
     t.integer  "average_score",         default: 0,     null: false
   end
 
+  create_table "themes", force: :cascade do |t|
+    t.string   "name",       limit: 255
+    t.string   "filename",   limit: 255
+    t.datetime "created_at",             null: false
+    t.datetime "updated_at",             null: false
+  end
+
```
acousticrobot: Was this intended to be part of this pull request? We aren't using the  "tasks" or "themes" tables anywhere in this PR, and I don't see the migrations that added them.

app/helpers/application_helper.rb
```diff
@@ -38,14 +38,13 @@ def body_class
...
-    if @title.nil?
-      base_title
-    else
-      "#{@title}"
-    end
+    # look up translation key based on controller path, action name and .title
+    # this works identical to the built-in lazy lookup
+    title = t("#{ controller_path.tr('/', '.') }.#{ action_name }.title", default: "#{current_course.name}")
+    # evaluate any variables in the title
+    # we may need to put some safety checks on this!
```
acousticrobot: I think we should address the best way to default to `current_course.name` without evaluating it before merge. Maybe something like:

```
    title = t("#{ controller_path.tr('/', '.') }.#{ action_name }.title", default: "")
    # evaluate any variables in the title, default to name of course
    title.empty? ? current_course.name : eval %{ "#{ title }" }
```

db/schema.rb
chcholman: Nope, not intended at all. Resetting it to master. Thank you!

app/helpers/application_helper.rb
chcholman: I'm actually much happier having the default be an empty string - the whole course name in this context is strange (but great for the other place we do this, as the page title for the browser).

PR #2467, opened 2016-09-23
1995 copy rubric

acousticrobot:
This PR adds the ability to copy a rubric onto an assignment.  This options is only available if the assignment does not already have a rubric.

Before merge, we should probably clean up the view `/rubrics/index_for_copy.haml`
  * Address the h1 and h2 content to fit in line with our current styles and titles.
  * View the information provided in the select list -- this is a subset of the assignments index table and could possibly be refactored to reuse partials if we want them to remain similar.


app/views/assignments/_buttons.html.haml
```diff
@@ -30,10 +30,17 @@
...
+          %a{:href => design_assignment_rubrics_path(presenter.assignment) }
+            = glyph(:sliders)
             Edit Rubric
-          - else
+      - else
+        %li.hide-for-small
+          %a{:href => design_assignment_rubrics_path(presenter.assignment) }
+            = glyph(:sliders)
             Create Rubric
+        %li.hide-for-small
```
chcholman: Let's remove this from the navigation, and add it at the top of the page to create a new rubric (and then hide it if the rubric exists already)

PR #2473, opened 2016-09-26
2441 autoprefixer gem

mariehooper:
This PR adds [Autoprefixer Rails](https://github.com/ai/autoprefixer-rails) to add vendor prefixes to CSS in post-production. This allows us to remove vendor prefixes from the Sass files and be assured that we are up to date with most recent standards.


chcholman:
Fantastic cleanup, so nice to see this stuff come out!

PR #2491, opened 2016-09-27
Student-awarded badges

ksofa2:
## Status
**READY**

## Migrations
YES

## Description
Add a field to Badge to indicate whether a badge is student-awardable.

## Related PRs
N/A

## Todos
N/A

## Deploy Notes
The `awarded_by_id` column allows NULL in migration in the this PR. This can be changed after the rake task is run with updates the awarded_by_id.

## Steps to Test or Reproduce
N/A

## Impacted Areas in Application
Badge awarding.


app/controllers/earned_badges_controller.rb
```diff
@@ -23,12 +23,15 @@ def confirm_earned
...
       render nothing: true, status: 200
-    else 
+    else
       redirect_to root_path
     end
   end
 
   def new
+    if current_user_is_student?
+      return redirect_to root_url unless @badge.student_awardable
```
ksofa2: This redirect pattern is used elsewhere to handle unauthorized access, I'm curious if there are other ways to handle this.

app/controllers/earned_badges_controller.rb
```diff
@@ -38,9 +41,16 @@ def edit
...
+    end
+
     result = Services::CreatesEarnedBadge.award earned_badge_params
 
     if result.success?
+      result.earned_badge.awarded_by = current_user
+      result.earned_badge.save
+
```
ksofa2: While this is clean and understandable, I wonder if there's a good way to do this through `Services::CreatesEarnedBadge.award` or if that's appropriate.

ksofa2:
(Looks like CodeClimate doesn't like the TODO.)

ksofa2:
- [ ]  Update `mass_edit` to set awarded_by

app/controllers/earned_badges_controller.rb
```diff
@@ -111,9 +123,9 @@ def destroy
   private
 
   def earned_badge_params
-    params.require(:earned_badge).permit :points, :feedback, :student_id, :badge_id,
+    params.require(:earned_badge).permit(:points, :feedback, :student_id, :badge_id,
       :submission_id, :course_id, :assignment_id, :level_id, :criterion_id, :grade_id,
-      :student_visible, :_destroy
+      :student_visible, :_destroy).merge(awarded_by: current_user)
```
ksofa2: I didn't have any luck with setting current_user directly in params, but found this approach with `.merge` (http://stackoverflow.com/questions/16530532/rails-4-insert-attribute-into-params).

Drawbacks to this technique? (I like that it's happening in `earned_badge_params`.)

app/controllers/earned_badges_controller.rb
ksofa2: The main difficulty I see with doing this with `Badge` is that we'll need to figure out the "is self awarding" piece (that you note in your comment on the `create` method below).

I think that will be more straightforward on an `EarnedBadge` proctor because we can compare `user.id` with `earned_badge.student_id`. Does that make sense?

acousticrobot: I think if you find as you work through it that creating an `EarnedBadge` proctor makes more sense you should go for it.  It's still related directly to a resource, so I don't see it as breaking any established patterns. 

The only disadvantage I see is you will have to build the EarnedBadge, and then check permissions to create with CanCan before you save it.

app/models/abilities/earned_badge_ability.rb
```diff
@@ -0,0 +1,10 @@
+module EarnedBadgeAbility
+  def define_earned_badge_abilities(user)
+    can :create, EarnedBadge do |earned_badge|
+      user.is_staff?(earned_badge.course) || (
+        earned_badge.badge.student_awardable? &&
+        earned_badge.student != user
+      )
+    end
+  end
+end
```
ksofa2: I'm liking this approach. Should I create a proctor, or is it OK to do it like this in the ability. (It looks like both patterns are used.)

app/controllers/earned_badges_controller.rb
```diff
@@ -111,9 +119,9 @@ def destroy
   private
 
   def earned_badge_params
-    params.require(:earned_badge).permit :points, :feedback, :student_id, :badge_id,
+    params.require(:earned_badge).permit(:points, :feedback, :student_id, :badge_id,
       :submission_id, :course_id, :assignment_id, :level_id, :criterion_id, :grade_id,
-      :student_visible, :_destroy
+      :student_visible, :_destroy).merge(awarded_by: current_user, course: current_course)
```
ksofa2: What do you think of this technique to get `current_user` into `params`? I like that it this is happening in `earned_badge_params`.

acousticrobot: I like this too -- I'm not aware of any drawbacks to this approach.

app/models/abilities/earned_badge_ability.rb
acousticrobot: I think it is fine to limit it to CanCan, especially since we only need the `create` check at this time. We could always delegate to a Proctor later.

app/controllers/earned_badges_controller.rb
```diff
@@ -23,21 +23,29 @@ def confirm_earned
...
+      @students = current_course.students
+    end
   end
 
   def edit
     @students = current_course.students
   end
 
   def create
+    authorize! :create, EarnedBadge.new(earned_badge_params)
```
acousticrobot: Here we are building the the earned badge twice -- once just to check the permissions, and then the second time because the existing service is expecting params to build with.  Have you tried updating `/services/creates_earned_badge/creates_earned_badge.rb`? to check the permission between build and save? It would have the advantage of then checking permissions during all uses of this service.

app/services/creates_earned_badge/creates_earned_badge.rb
```diff
@@ -8,6 +8,15 @@ class CreatesEarnedBadge
...
         context.earned_badge = EarnedBadge.new context.attributes
+
+        unless context.earned_badge.awarded_by.is_staff?(context.earned_badge.course) || (
+          context.earned_badge.badge.student_awardable? &&
+          context.earned_badge.student != context.earned_badge.awarded_by
+        )
+          message = "Permission denied"
+          context.fail_with_rollback! message
+        end
+
```
ksofa2: I think I'm changing my mind on this. I've taken the PR back while I ponder it.

ksofa2:
OK. made some changes that I think give me what I wanted to do here.

ksofa2:
I've redone things to use an ability and a proctor to check permissions. I was initially trying to avoid this complexity, but I now think it's worth it.

app/proctors/earned_badge_proctor.rb
```diff
@@ -0,0 +1,20 @@
...
+
+  attr_reader :earned_badge
+
+  def initialize(earned_badge)
+    @earned_badge = earned_badge
+  end
+
+  def creatable?(user)
+    return false unless earned_badge.course.users.include? user
+    return false unless earned_badge.student.nil? || earned_badge.course.students.include?(earned_badge.student)
```
acousticrobot: Could you tell me what this accomplishes? I can see why we would want to check that the student is in the course, but if we are also allowing the student to be nil, I'm not sure if this check is effective.

app/services/creates_earned_badge/creates_earned_badge.rb
```diff
@@ -8,6 +10,12 @@ class CreatesEarnedBadge
 
       executed do |context|
         context.earned_badge = EarnedBadge.new context.attributes
+
+        unless EarnedBadgeProctor.new(context.earned_badge).creatable? context.earned_badge.awarded_by
+          message = "Permission denied"
+          context.fail_with_rollback! message
```
acousticrobot: I think you can put the message string in this line without needing to assign it to a variable. You can also include an error code.

ksofa2: Quite right, I'll make this change. (I was following the pattern below; I'll change that too.)

app/proctors/earned_badge_proctor.rb
```diff
@@ -0,0 +1,25 @@
...
+      )
+    )
+  end
+
+  def creatable?(user)
+    newable?(user) &&
+    earned_badge.course.students.include?(earned_badge.student) &&
+    earned_badge.student != user
+  end
+end
```
ksofa2: I'm liking this approach with two abilities instead of one. What do you think? (I don't really like the name `newable`. Got any suggestions?)

acousticrobot: 👍   This is a great separation and makes for good reading sense. Following the FactoryGirl pattern, I would suggest calling it `buildable` and then calling `authorize! :build, @earned_badge` in the controller.

ksofa2: Thanks, I'll update that.

PR #2495, opened 2016-09-29
2445 instructor rubric

mariehooper:
# Not Ready for Merging
This PR refines the style of the instructor rubric design and rubric grade edit views to match the updated student design.

## Considerations
We are using color to represent many different states of Criteria and Levels in rubrics. 
### Level States
- Saved : light grey -> counterintuitive since *unsaved* criterion is this color
- Unsaved: dark grey
- Meets Expectations: green -> proposing we shift this to an indicator other than background color
- Selected Meets Expectations Level: green
- Selected below or no meets expectations level: green
- Earned: green

### Criterion States
- Saved: dark blue -> this makes sense since in all other rubric views, criteria is dark blue
- Unsaved: light grey -> this is fine if we want to show unsaved state, but I wonder if users are realizing that this is what this color change means.

## Proposed changes
- Alter the way 'Meets Expectations' threshold is represented on both rubric design and grade edit view. Rather than using a green background-color to show these levels, let's add a marker to each of the levels that meets expectations so that all selected and earned levels can be green without confusion.
- get rid of the saved vs unsaved states in the rubric design view since users may not actually realize what is happening...?

### DIFF
--- a/app/assets/javascripts/angular/controllers/GradeRubricCtrl.js.coffee
+++ b/app/assets/javascripts/angular/controllers/GradeRubricCtrl.js.coffee
@@ -20,10 +20,19 @@
     $scope.services()
 
   $scope.services = () ->
+    # because getting Criteria requires badges from $scope
+    # we need to wait until the badges are created before
+    # making this call.
+    queCriteriaAfterBadges = ()->
+      if (RubricService.badgesAvailable() == false)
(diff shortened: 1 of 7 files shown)


chcholman:
@mariehooper I am so excited about this rework. Will you:

- [ ] Move the criteria summary to a modal
- [ ] Get the criteria blue box to take up the full width of the box
- [ ] Remove the saved state coloring

Issue #2472, opened 2016-09-26
Attachments to rubric grades should be removable


chcholman:
Following the same pattern as the standard grade edit form, attachments to the rubric grade form should be removable by the instructor.

PR #2496, opened 2016-09-29
2471 rubric attachment

### LINKED ISSUE #2471: Attachments not displaying if previously attached in rubric grade edit form

acousticrobot:
This PR delegates the attached file presentation to Angular when grading with a Rubric. This way files that have been previously uploaded with also appear in the list with new attachments.

It includes the grade files in the json response from `/api/assignments/id/students/id/grade`

It adds a better JSON success response from `api/grades/grade_files_controller.rb#create`. This file still doesn't handle a failure response, I would like to know if there is a good way to handle this when multiple resources are created. 

It doesn't handle removing files ( feature request #2472), or attachments for groups, this will be added in the next PRs.

closes #2471


chcholman:
@acousticrobot This is looking great! One question - the spinner seems to spin constantly once you attach something locally. Is that an artifact of being local, or the functionality you expect for production as well? Should the user stay on the page until the file is noted as uploaded, or is submitting the grade safe to do while it's still spinning?

chcholman:
@acousticrobot Totally agree with the 1) pathway.

PR #2508, opened 2016-10-03
Downcasing the email on import

chcholman:
## Status
**READY**

## Migrations
NO

## Description
This PR downcases emails on import and user creation so that matching them across contexts is simplified. 

## Related PRs
n/a

## Todos
- [ ] Tests
- [ ] Documentation

## Deploy Notes
None

## Steps to Test or Reproduce
1. Create a new user with the email address "TEST@gradecraft.com"
2. Confirm that the stored email address is "test@gradecraft.com"
--
1. Import a CSV of user with the email addresses "TESTING@gradecraft.com" 
2. Confirm that the stored email address is "testing@gradecraft.com"

## Impacted Areas in Application

* Import workflow
* User creation

### DIFF
--- a/app/services/creates_or_updates_user/creates_or_updates_user.rb
+++ b/app/services/creates_or_updates_user/creates_or_updates_user.rb
@@ -10,7 +10,8 @@ module Services
 
       executed do |context|
         attributes = context[:attributes]
-        email = attributes[:email]
+        entered_email = attributes[:email]
+        email = entered_email.downcase
 
         if email.blank?
           context.fail! "Email can't be blank", 422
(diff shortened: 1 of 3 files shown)


app/services/creates_or_updates_user/creates_or_updates_user.rb
```diff
@@ -10,7 +10,8 @@ class CreatesOrUpdatesUser
 
       executed do |context|
         attributes = context[:attributes]
-        email = attributes[:email]
+        entered_email = attributes[:email]
+        email = entered_email.downcase
```
acousticrobot: If we aren't using `entered_email` for any other purpose, I think we can chain these two together: `email = attributes[:email].downcase`.

jwright: We may want to ensure that `nil` is not caught here.

```
email = attributes[:email].to_s.downcase
```

PR #2514, opened 2016-10-03
copy level badges not badges for level

### LINKED ISSUE #2503: Copying a Rubric duplicates Course Badges
As an instructor, if I copy a rubric between assignments, it duplicates the badges currently in the course (it should be duplicating level badges, but not the master badge).

acousticrobot:
The copy method for Levels was duplicating Badges, when it should have only duplicated the LevelBadges, referencing the same Badge.

This error was occurring when a Rubric was copied to another Assignment.  I don't believe this behavior is desired in any other situation when a Level is copied, but I would like confirmation since it was a preexisting behavior.

closes #2503


jwright:
@acousticrobot Looks great. I have one small suggesting about the spec for `LevelBadge#copy`. 

I can't comment on the preexisting behavior.

PR #2517, opened 2016-10-04
1492 quick grade for groups

### LINKED ISSUE #1492: Quick grade for groups
Instructors should be able to grade all groups for an assignment at the same time in list form the same way they can for individual assignments

jonathoy:
Allows a professor to quick grade assignments with groups.

screen shot 2016-10-04 at 4 31 58 pm

When you go to the quick grade page, you will be presented with a list of groups for that assignment. Upon assigning the group grade, all students within that group will receive the specified grade.

Fixes #1492


jonathoy:
@acousticrobot Perhaps you might be interested in taking a look at this PR, if and when you get the chance.

app/views/assignments/grades/components/_group_table_body.haml
```diff
@@ -0,0 +1,28 @@
+- @grades_by_group.each_with_index do |gbg, i|
+  %tr
+    %td{:style => "text-align: right"}= link_to gbg[:group].name, group_path(gbg[:group])
+    %td.center
+      = f.simple_fields_for "grades_by_group[]", i do |gf|
+        = gf.hidden_field :graded_by_id, value: current_user.id
```
acousticrobot: I know you are copying this pattern from the `grades/mass_edit.html.haml` but it seems to me that the best practice would be to set `graded_by_id` in the back end. This might be a case for why it is better to explicitly send this id into the `BuildsGrade` service action as it's own parameter. This way we can:

  * Ensure that is is always set because it will be required by the service itself.
  * Manage this from the controller actions and `current_user`, rather than permitting non-secure forms from setting this parameter. 

The same is probably true with `instructor_modified` as well, but less important since this field is slated for removal.

@chcholman @jwright I wonder if you have an opinion about this, especially in the context of our discussions re: #2380

app/controllers/assignments/grades_controller.rb
```diff
@@ -73,39 +76,53 @@ def index
...
     @assignment = current_course.assignments.find(params[:assignment_id])
-    if @assignment.update_attributes(assignment_params)
+
+    success = false
+    if @assignment.has_groups?
+      assign_graded_at_date params[:assignment][:grades_by_group] if params[:assignment][:grades_by_group].present?
+      result = Services::CreatesManyGroupGrades.create @assignment.id, assignment_group_grades_params
+      success = result.success?
+    else
+      assign_graded_at_date params[:assignment][:grades_attributes] if params[:assignment][:grades_attributes].present?
```
acousticrobot: Would it make sense to handle the `else` with a service as well, perhaps `CreatesManyGrades`, now that you have done all this refactoring of the service?

app/controllers/assignments/grades_controller.rb
```diff
@@ -183,4 +206,23 @@ def mass_update_grade_ids
       memo
     end
   end
+
+  def assign_graded_at_date(params)
```
acousticrobot: Could this be handled within the service, again to standardize the process of generating grades?

app/controllers/assignments/grades_controller.rb
```diff
@@ -183,4 +206,23 @@ def mass_update_grade_ids
...
+    params.each do |index, value|
+      value.merge!(graded_at: DateTime.now)
+    end
+  end
+
+  def redirect_on_mass_update_success
+    if !params[:team_id].blank?
```
acousticrobot: I understand you may have wanted to break these out since if the controller action method was getting too long.  We found it a little unmanageable in the past through to break out private methods handling success and failure for a single route.

app/services/group_services/verifies_group.rb
```diff
@@ -0,0 +1,19 @@
+module Services
+  module Actions
+    class VerifiesGroup
```
acousticrobot: 👍 Excellent refactor here!

app/views/assignments/grades/components/_individual_table_body.haml
```diff
@@ -0,0 +1,29 @@
+- @grades.each_with_index do |grade, i|
+  %tr
+    %td{:style => "text-align: right"}= link_to grade.student.name, student_path(grade.student)
+    %td.center
+      = f.simple_fields_for :grades, grade do |gf|
+        = gf.hidden_field :graded_by_id, :value => current_user.id
```
acousticrobot: same comment as above re: `graded_by_id`

app/controllers/api/criterion_grades_controller.rb
```diff
@@ -47,7 +48,7 @@ def update
   end
 
   # PUT api/assignments/:assignment_id/students/:student_id/criteria/:id/update_fields
-  def update_fields
+  def update_field
     cg = CriterionGrade.find_or_create(params[:assignment_id], params[:id], params[:student_id])
```
jonathoy: I wonder if this was an unintended change that somehow slipped through. Thanks for catching this.

app/controllers/assignments/grades_controller.rb
jonathoy: @acousticrobot If I put this logic back into mass_update, I get something like this:

```
if success
  if !params[:team_id].blank?
    redirect_to assignment_path(@assignment, team_id: params[:team_id])
  else
    respond_with @assignment
  end
else
  redirect_to mass_edit_assignment_grades_path(@assignment, team_id: params[:team_id]),
        notice: "Oops! There was an error while saving the grades!"
end
```

which RuboCop does not like. How would you personally clean this up?

jonathoy:
I went ahead and refactored the services to address the concerns that @acousticrobot had. To sum up, I changed the individual quick grading to use the grading services and moved the graded_at and graded_by_id to `BuildsGrade`.

I left `instructor_modified` for now since it seems that it will be going away soon. 

Per our discussion, I also removed a broken Rspec test that was no longer really relevant to the state of the code in this PR.

jonathoy:
@jwright I know that there were some concerns that you had when we reviewed this PR yesterday, so I wanted to assign this back to you. With all the refactoring, it ended up being a rather large changeset.

app/controllers/api/criterion_grades_controller.rb
jonathoy: I changed the method name back to update_fields in 44e072540af91bd649eb3cc60903c7f63c23d325 as it was an unintended name change.

spec/services/group_services/iterates_assignment_groups_to_create_grades_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+describe Services::Actions::IteratesAssignmentGroupsToCreateGrades do
+  let(:assignment) { create(:assignment) }
+  let(:professor) { create(:user) }
+  let(:grades_by_group_params) { { grades_by_group: {
+    "0" => { "instructor_modified" => "true", "raw_points" => "10", "status" => "graded" },
+    "1" => { "instructor_modified" => "true", "raw_points" => "20", "status" => "graded" } } }
+  }
+  let(:context) { { assignment_id: assignment.id, grades_by_group_params: grades_by_group_params, graded_by_id: professor.id } }
+
+    it "expects an grades_by_group_params" do
```
jwright: Some of the spacing is off here.

app/views/assignments/grades/components/_group_table_body.haml
```diff
@@ -0,0 +1,27 @@
+- presenter.grades_by_group.each_with_index do |gbg, i|
+  %tr
+    %td{:style => "text-align: right"}= link_to gbg[:group].name, group_path(gbg[:group])
+    %td.center
+      = f.simple_fields_for "grades_by_group[]", i do |gf|
+        - if !gbg[:grade].status
+          = gf.hidden_field :status, value: "Graded"
+        = gf.hidden_field :instructor_modified, value: true
```
jonathoy: Should I go ahead and remove this field for update or leave it for now?

jonathoy:
@jwright I made a few changes to address some of the suggestions that you had for code cleanup. For the record, I think you made valid comments so I don't hate you.

A couple areas that I haven't addressed:

1. You mentioned that there was a lot of logic in `_group_table_body.haml` and `_individual_table_body.haml`. I cleaned up `_group` a bit by moving more into the presenter, but I'm wondering if maybe I should revisit refactoring the `_individual` partial and the mass grading logic on an individual basis in a separate PR since this changeset is becoming a bit large and the view is untouched from the way it was before. From what I can tell, there is some redundant logic in that view anyway and I would like for it to go into a presenter as well.

2. I also wasn't entirely sure how to best handle the setting of the grade `status` and other attributes such as `instructor_modified`. I agree with you in that they probably shouldn't be on the front-end. Would they be best put in the controller, in the refactored service somewhere, or should we wait to make changes on this?

jwright:
@jonathoy Thank you!

1. Yes, let's definitely do this in a separate PR since this one is getting large.

2. Yes, in the controller would be fine. You can do something like:

```
grade_params.merge(instructor_modified: true, status: "Graded")
```

app/controllers/assignments/groups/grades_controller.rb
```diff
@@ -0,0 +1,45 @@
...
+  params.require(:assignment).permit grades_by_group: [:graded_by_id, :graded_at,
+                                                         :instructor_modified, :raw_points,
+                                                         :status, :pass_fail_status, :group_id]
+end
+
+def find_assignment
+  @assignment = current_course.assignments.find(params[:assignment_id])
+end
+
+def filter_params_with_raw_points!(attribute_name)
```
jwright: Can we rename this to `filter_params_with_no_grades!` like in the other controller? This will indicate that there is duplicate code in a few controllers and we can identify it better. I think the method name `filter_params_with_no_grades!` is better.

app/services/creates_grade/assert_result_from_many_outcomes.rb
```diff
@@ -0,0 +1,16 @@
...
+module Services
+  module Actions
+    class AssertResultFromManyOutcomes
+      extend LightService::Action
+
+      expects :unsuccessful
+
+      executed do |context|
+        unsuccessful = context[:unsuccessful]
+        context.fail!("There were #{"grade".pluralize(unsuccessful.length)} that were not updated") unless unsuccessful.empty?
```
jwright: The word "were" can be called with `pluralize` as well to determine if "was" or "were" should be used.

app/services/creates_many_group_grades.rb
```diff
@@ -0,0 +1,15 @@
+require "light-service"
+require_relative "group_services/iterates_assignment_groups_to_create_grades"
```
jwright: I think this should be moved to shared. Thoughts?

jonathoy:
@chcholman At your convenience, this should be ready for testing. Changes were made that extend to group quick grading as well as rubric and individual quick grading. Let me know if you encounter any issues.

chcholman:
@jonathoy I've gone through and tested each mode of grade awarding - this looks great to me! Will you confirm with @jwright that this is merge-ready since his review says it's still requesting changes?

jonathoy:
@chcholman I missed the comment about moving one of the services over to shared, and just made the commit to move it over. @jwright If you want to double-check that the changes are acceptable that works for me, otherwise they should take care of the comments in the previous review.

jwright:
@jonathoy @chcholman Everything looks great!

PR #2528, opened 2016-10-05
passing team id through submit and grade next

acousticrobot:
This PR adds an additional layer to the "Grade for next Student" workflow, to enhance the GSI experience.  If a student is chosen for grading from the assignment show page with the page filtered down to a single team, the "submit and grade next" should bring up the next student *in that team*.  This requires passing team id in the params through the process.

### DIFF
--- a/app/models/concerns/gradable.rb
+++ b/app/models/concerns/gradable.rb
@@ -65,22 +65,27 @@ module Gradable
     predicted_earned_grades.predicted_to_be_done.count
   end
 
-  def ungraded_students(ids_to_include=[])
-    course.students.order_by_name -
-      (User.find(grades.graded.pluck(:student_id)) - User.find(ids_to_include))
-  end
-
-  def ungraded_students_with_submissions(ids_to_include=[])
-    ungraded_students(ids_to_include) & User.find(submissions.pluck(:student_id))
-  end
-
-  def next_ungraded_student(student)
-    if accepts_submissions?
-      ungraded_students = ungraded_students_with_submissions([student.id])
+  def ungraded_students(ids_to_include=[], team=nil)
+    if team
+      course.students_by_team(team).order_by_name -
+        (User.find(grades.graded.pluck(:student_id)) - User.find(ids_to_include))
     else
-      ungraded_students = ungraded_students([student.id])
+      course.students.order_by_name -
+        (User.find(grades.graded.pluck(:student_id)) - User.find(ids_to_include))
     end
-    i = ungraded_students.map(&:id).index(student.id)
-    i && i < ungraded_students.length - 1 ? ungraded_students[i + 1] : nil
+  end
+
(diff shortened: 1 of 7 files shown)


app/models/concerns/gradable.rb
```diff
@@ -65,22 +65,27 @@ def predicted_count
     predicted_earned_grades.predicted_to_be_done.count
   end
 
-  def ungraded_students(ids_to_include=[])
-    course.students.order_by_name -
-      (User.find(grades.graded.pluck(:student_id)) - User.find(ids_to_include))
+  def ungraded_students(ids_to_include=[], team=nil)
+    if team
+      course.students_by_team(team).order_by_name -
+        (User.find(grades.graded.pluck(:student_id)) - User.find(ids_to_include))
```
jwright: I would vote for moving this small amount of duplication to a method.

spec/controllers/grades_controller_spec.rb
```diff
@@ -1,47 +1,41 @@
 require "rails_spec_helper"
 require "grade_proctor"
 
-describe GradesController do
+describe GradesController , focus: true do
```
jwright: `focus` flag should be removed

spec/controllers/grades_controller_spec.rb
```diff
@@ -1,47 +1,41 @@
...
   include PredictorEventJobsToolkit
 
-  before(:all) do
-    @course = create(:course)
-    @assignment = create(:assignment, course: @course)
-    @student = create(:user)
-    @student.courses << @course
-  end
+  let(:course) { create :course }
```
jwright: 👏

spec/controllers/grades_controller_spec.rb
```diff
@@ -1,47 +1,41 @@
...
-    before (:each) do
-      @grade = create(:grade, student: @student, assignment: @assignment, course: @course)
-      login_user(@professor)
+    before do
+      login_user(professor)
     end
 
     after(:each) do
-      @grade.delete
+      grade.delete
```
jwright: Since this is now in a `let`, I don't believe we need this.

spec/controllers/grades_controller_spec.rb
```diff
@@ -198,86 +203,84 @@
...
-      @grade = create :grade, student: @student, assignment: @assignment,
-        course: @course
-      login_user(@student)
-      allow(controller).to receive(:current_student).and_return(@student)
+      login_user(student)
+      allow(controller).to receive(:current_student).and_return(student)
     end
-    after(:each) { @grade.delete }
+
+    after(:each) { grade.delete }
```
jwright: I don't think this is required since it is using a `let`

acousticrobot: 🎆

Issue #2523, opened 2016-10-05
Make criterion details bar full width


chcholman:
The criterion bar should take up the full width of the outer container

![screen shot 2016-10-05 at 12 13 49 pm](https://cloud.githubusercontent.com/assets/234276/19121657/3ccb6264-8af5-11e6-97dc-6d23fa7cfeaa.png)

PR #2533, opened 2016-10-06
make criterion bars full width

mariehooper:
This PR makes the criterion bars full width on the rubric design interface and updates the level heading file in order to only show levels earned when grade has been released.

Closes issue #2523


chcholman:
@mariehooper Will you remove the red state on the new criterion?

PR #2536, opened 2016-10-06
1705 criterion grades review

acousticrobot:
NOT READY FOR MERGE

This PR revives the lost "Criterion Grades Review" page, renames it "Grades Review", and modifies it to work with both standard and rubric graded assignments.  The page has no public linking, but it available for any assignment at `/assignments/:id/grades_review`.  This page still uses a legacy rubric view structure, which I have quickly stripped down to what I think is the important information.

I am passing it along for review so we can determine:
  * where the link button on the assignment page should be
  * what grade and rubric information do we want to display, and what should be left out for the sake of reviewing a list of ~300 grades.  
  * If some of the current class tags and partials dealing with rubric display can be modified to be used on this page as well.

I am happy to pass this work along to others or to take it back with some guidance regarding these quesitons.

### DIFF
--- a/app/assets/stylesheets/pages/_rubrics.sass
+++ b/app/assets/stylesheets/pages/_rubrics.sass
@@ -336,7 +336,7 @@ h4.notice
     @media (max-width: $media-small-max)
       display: none
 
-  .level-tab
+  .level-tab, .criterion-tab, .comments-tab
     list-style-type: none
     display: inline-block
     flex: 1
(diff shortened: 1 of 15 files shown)


chcholman:
@acousticrobot Passing this back to you with: 

* The information in place for both the standard and rubric grades review
* Implemented reuse of various grade components

I think the link to access this page should be in the new dropdown menu that Marie is currently working on/should be merged later this week.

PR #2538, opened 2016-10-06
remove legacy student preview routes

### LINKED ISSUE #2401: remove legacy routes
Now that faculty have a preview mode, all controllers in `/app/controllers/api/students/` are probably unnecessary and can be removed.

  * remove controllers
  * remove routes
  * check to see if `uriPrefix` method can be removed and all Angular routes that use it cleaned up.

acousticrobot:
Now that we have the `impersonating_agent` logic to preview all pages as a student, we no longer need the code and api routes that supported the `student/:id/predictor` or the `view_student_context` logic.

This PR removes this legacy code. The only `api/student/` controller remaining is `badges`, which is still used to index awarded badges for a student on the grading page.

closes #2401


app/views/badges/index.html.haml
```diff
@@ -7,7 +7,7 @@
   .pageContent
     = render "layouts/alerts"
 
-    %table.badge-index-table.second-row-header{:class => ("student-index" if presenter.view_student_context?)}
```
jwright: It looks like we can remove this class as well from the styles. I don't see it being used anywhere else.

PR #2543, opened 2016-10-07
remove the extreme redness on new criterion

### LINKED ISSUE #2535: Remove color alert on new criterion items
Currently we color the name and points red when new criterion are created. We should remove this color cue when the initial item is created. We should add an alert in the instance that they don't enter anything and can't save the data.

acousticrobot:
* removes red background on new criterion
  * adds a prompt below -- "Criterion Name" and "Max Points" are required fields -- that disappears as soon as these fields are filled.

screen shot 2016-10-07 at 10 52 04 am

Closes #2535


chcholman:
@acousticrobot Thanks for this! It's a much more elegant solution.

Issue #2241, opened 2016-07-26
Remove high level on the Grade Scheme Element


chcholman:
reframe the interface to just show the 'minimum points threshold' to earn each

PR #2545, opened 2016-10-07
Remove high level on the grade scheme element

ksofa2:
This PR begins work on #2241.

It removes the highest_points parts of the GSE interface, and adds a method to calculate highest_points before saving changes to the grading scheme.

So these changes are compatible with the existing GradeSchemeElement model. Next step will be to begin to update the parts of the application that rely on highest_points, and then eliminate highest_points from the model.

### DIFF
--- a/app/assets/javascripts/angular/controllers/GradeSchemeElementsCtrl.js.coffee
+++ b/app/assets/javascripts/angular/controllers/GradeSchemeElementsCtrl.js.coffee
@@ -7,13 +7,20 @@
 
     $scope.validateElements = (index)->
       for element, i in $scope.grade_scheme_elements
-        $scope.gradeSchemeForm["lowest_points_#{i}"].$setValidity('conflict', true)
+        $scope.gradeSchemeForm["lowest_points_#{i}"].$setValidity('directConflict', true)
+        $scope.gradeSchemeForm["lowest_points_#{i}"].$setValidity('nearConflict', true)
         for otherElement, j in $scope.grade_scheme_elements
           continue if i == j
 
-          pointRange = [(element.lowest_points - 1)..(element.lowest_points + 1)]
-          if otherElement.lowest_points in pointRange
-            $scope.gradeSchemeForm["lowest_points_#{i}"].$setValidity('conflict', false)
-            $scope.gradeSchemeForm["lowest_points_#{j}"].$setValidity('conflict', false)
+          if element.lowest_points == otherElement.lowest_points
+            # Invalid because it is in direct conflict with another level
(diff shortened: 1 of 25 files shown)


chcholman:
@ksofa2 Thanks so much for this!

A few UX concerns:

- When I add a new level, it appears below the one I clicked on - I think the button should probably indicate that by saying 'Add a level below'
- That new level is has the same value as the level you created it off of. I think we should probably set the item as blank to begin with? 
- If you set a level value to higher than the value of the one above it, it fails to validate/submit the data, but there's no cue in the interface for the user that this has happened. It seems like we should indicate this with color or an alert?

ksofa2:
Thanks @chcholman, good points. I thought I'd handled that last case, must have dropped it at some point. I'd also like to beef up the handling of errors from the server on submit, and I think I'll update this PR to do that.

chcholman:
@ksofa2 Awesome!

ksofa2:
I've added client-side validation that disables the submit button when input is invalid.

I've expanded the validation to account for the case when the thresholds are within 1 point of each other (this is necessary because the model still has both `highest_points` and `lowest_points`.

mariehooper:
@ksofa2 - This is looking good! Just a couple items - I would love if there were error messages that told me what I was doing wrong if I was unable to submit for some reason. Also - in that same vein - when I am lacking a 'zero' level, I see that it's created for me when I hit Save, but I would like this to show me an alert instead so that I am forced to make a new level that has a bottom of 0 and name it.

ksofa2:
Good suggestions, thanks @mariehooper.

@chcholman I'm taking this back to implement Marie's suggestions.

chcholman:
@ksofa2 Awesome!

ksofa2:
I've added error message to explain the validation errors. I'm putting the error message in a `title` on an icon. Open to ideas about other ways to implement this.

Also added an alert when 0-level is missing.

mariehooper:
@ksofa2 - this looks good to me! I made one update to make sure the font was rendering properly on the add level button.

ksofa2:
@mariehooper Thanks!

chcholman:
@ksofa2 This looks great! One tiny change - rather than using the title attribute, can you use the hover state that we use to describe other icons? Example here: https://github.com/UM-USElab/gradecraft-development/blob/master/app/views/assignments/_index_icons.haml#L4 

Thank you!

ksofa2:
@chcholman Thanks for the pointer, I'll get that into this PR soon.

chcholman:
@ksofa2 Awesome, thank you!

PR #2548, opened 2016-10-10
Titles and tabs

chcholman:


mariehooper:
@chcholman - check this out and see what you think. I think we can further refine how we are specifying menu items in the context menu... but our consistency here is much improved. I left the animation on the tabs (see what you think. It started growing on me)

PR #2557, opened 2016-10-11
1938 rails 5 upgrade

### LINKED ISSUE #1938: Rails 5

jwright:
Upgrades the current Rails version 4.2.3 to Rails 5.0.0!

There were a lot of gem dependencies that needed to be changed or upgraded to support Rails moving from Rack 1 to Rack 2.

**Please pull down locally and ensure that `bundle install` works and the server can be started. Also ensure there are no obvious stylesheet issues when running the app.**

## Notable Changes

1. [mailcatcher](https://github.com/sj26/mailcatcher) was removed. This was because it could not be started in the `Procfile` as there it depends on an old version of `ActiveSupport`. It seems like it will not be updated because it is supposed to be run seperately. (https://github.com/sj26/mailcatcher/issues/303).
1. There is a depreciation warning with controller specs when the keyword `params` is not used. This means that all of the following request calls were changed from something like this: `get :show, id: model.id` to this: `get :show, params: { id: model.id }`.
1. Changed controller callbacks to use `_action` instead of `_filter`. This includes `before_filter` and `after_filter`.
1. `render nothing: true` is depreciated and replaced with `render head: :ok, body: nil`.
1. `redirect_to :back` is depreciated and replaced with `redirect_back`.
1. `or` is a new keyword in `ActiveRecord` and replaced instances where a `WHERE` clause was manually written to support `OR`.
1. Capybara no longer finds hidden nodes in integration specs. If a node is hidden (i.e. `<div class="blah" style="display: none;">...</div>`), then `page.find(".blah")` will not return anything. This needs to change to something like `page.find(".blah", visible: false)`.
1.  The [mongoid.yml.sample](https://github.com/UM-USElab/gradecraft-development/blob/master/config/mongoid.yml.sample) was changed to use newer yml keywords. For mongoid, it should refer to `clients` instead of `sessions`. The sample was changed but the "live" version (`config/mongoid.yml`) on development and production should change. (Added to [deployment notes](#deployment-notes).
1. Several "array like" methods were removed from `ActiveRecord::Association` such as `<<` so these instances were updated to use newer constructs or convert the collections to arrays first.
1. The application config files were updated to replace older methods on config with newer methods.
1. This guide was used in the upgrade which spells out several more changes: http://edgeguides.rubyonrails.org/upgrading_ruby_on_rails.html

## Deployment Notes

1. Change `config/mongoid.yml` to use the `clients` keyword instead of the `sessions` keyword.

Closes #1938


jwright:
@acousticrobot I cannot get the failing test to pass for a performer. It passes locally but does not pass on the CI server. Do we remove the spec?

jwright:
@acousticrobot Thank you for your changes!

I will take a gander further regarding the error and timebox a fix. I want to ensure that the background jobs are actually working as when running the app.

@chcholman I will pass this to you but we may want to do a few things before we merge this in.

1. Have all the developers pull this down and ensure they can run the specs and run the server.
2. Tag a version of the application before this is merged.

I will let you know when I look at the error.

jwright:
@chcholman This is ready for merging Cait. I fixed the failing spec issue.

PR #2564, opened 2016-10-13
2527 limit delete all grades

### LINKED ISSUE #2527: limit "delete all grades" to team if filtered by team
The Assignment show page lists all grades for all students for that assignment, and includes a "delete all grades" button.  If the grader has filtered the list down to a team by using the "Select Team" dropdown selector, and clicks the delete all button, it should only delete grades for that team.

jonathoy:
### Status
Ready

### Description
When viewing grades for an assignment, there is a button that allows the user to "Delete All Grades". The delete action was previously deleting all grades for the assignment, regardless of whether the user filtered down students by team or not.

This change allows for the deletion of all grades on the assignment if no team is selected, as well as the deletion of grades for a subset of students belonging to a team.

### Deploy Notes
N/A

### Migrations
NO

Ensure grade deletions for an assignment work as intended.

Closes #2527

### DIFF
--- a/spec/controllers/assignments/grades_controller_spec.rb
+++ b/spec/controllers/assignments/grades_controller_spec.rb
@@ -1,52 +1,45 @@
 require "rails_spec_helper"
 
 describe Assignments::GradesController do
-  before(:all) do
-    @course = create(:course)
-    @assignment = create(:assignment, course: @course)
-    @student = create(:user)
-    @student.courses << @course
-  end
-  before(:each) do
-    @grade = create :grade, student: @student, assignment: @assignment,
-      course: @course
-  end
-  after(:each) { @grade.delete }
+  let(:course) { create(:course) }
+  let(:assignment) { create(:assignment, course: course) }
+  let(:student) { create(:user) }
+  let!(:student_course_membership) { create(:student_course_membership, course: course, user: student) }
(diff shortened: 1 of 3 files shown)


jonathoy:
@acousticrobot I removed all the instance variables and replaced with `let`

spec/controllers/assignments/grades_controller_spec.rb
```diff
@@ -68,197 +61,209 @@
...
+        expect(post :self_log, assignment_id: assignment.id ).to \
           redirect_to(:root)
       end
     end
 
     describe "DELETE delete_all" do
-      it "deletes all the grades for the assignment" do
-        delete :delete_all, assignment_id: @assignment.id
-        expect(@assignment.reload.grades).to be_empty
+      context "when there is no team id" do
```
acousticrobot: I would add another grade here for these specs, so that we are verifying that all grades are deleted without a team, and that only the team grade is deleted with team params.

Something like

```
      let!(:other_student) { create(:student_course_membership, course: course).user }
      let!(:other_grade) { create(:grade, assignment: assignment, course: course, student: other_student) }
```

jonathoy:
Good catch, thanks. I believe this should be ready to merge now if there are no additional suggestions.

PR #2565, opened 2016-10-13
2549 assignment index page

mariehooper:
### Status
**READY**

### Description
This PR restyles the assignment index page for students to make it more usable and understandable per feedback we received in several rounds of usability testing.

### Todos
- [x] Tests
- [x] Documentation

* Student Assignment Index

Closes issue #2565

### DIFF
--- a/app/assets/stylesheets/components/_collapse.sass
+++ b/app/assets/stylesheets/components/_collapse.sass
@@ -1,11 +1,11 @@
 @import "../utilities/variables"
 
 /* Add the sprite to the assignment expand */
-h3.collapse
+.collapse
   margin: 0
 
-h3.collapse.open a i.fa.fa-angle-double-right.fa-fw
+.collapse.open a i.fa.fa-angle-double-right.fa-fw
   transform: rotate(90deg)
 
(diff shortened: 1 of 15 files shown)


chcholman:
@mariehooper Absolutely love the new design, it's so much more clear what's going on. I fixed one bug (not introduced by you, but it became apparent when testing) around pass/fail grading. Nice job updating the tests!

A few pieces of design feedback: 
* I love the green check for things that have been done, but I'm not sure it's enough to call out the different state. Would you take a look at what it looks like with a light color background?
* I think the assignment container should be full width on desktop rather than centered
* I like the styling of the dark blue line beneath the assignment type name, but it took me a bit to see the connection between the title and the assignments within that category. How would we make that feel more cohesive? 
* What do you think about changing the hover styling from an underline to a background color? 
* Is the margin beneath the open assignment type section purposeful? It feels slightly weird given that there's no margin when it's in the closed state.

mariehooper:
@chcholman - see what you think.. should be good to go

chcholman:
@mariehooper Absolutely love it, beautiful job on this redesign.

Issue #1743, opened 2016-03-09
Unpredictable assignments if requirements have passed


chcholman:
Some assignments need to be unlocked by others. If an earlier assignment is now unavailable (and zeroed out in the predictor), then the things that require it to be unlocked should also be zero.

In this state, the card should be greyed out, and not allow an interaction.

acousticrobot:
blocked until #2552 is resolved

PR #2578, opened 2016-10-17
2552 assignment json

### LINKED ISSUE #2552: better assignment json
I removed the `find_or_create Grades` dependency for the predictor by passing in a NullGrade. This works, but causes an issue on the outlier predictor widgets for closed assignments that have a submission. They should still be open for prediction but a NullGrade will return nil for submissions.

The solution will be to either patch the logic to work in this particular case, or to refactor the json with an includes section and handle this format properly in the Angular.

This is also blocking #1743.

acousticrobot:
### Status
**READY**

### Description

The main objective of this PR is to format JSON for Assignments, Challenges, and Badges without needing to create prediction models if they don't exist.

Currently grades and predictions are nested within the Assignments, Challenges and Badges JSON. This requires `find_or_create` to be run on all predictions for a student, and then all predictions to be run through an `update` method.

This PR uses the [JSON API](http://jsonapi.org/format/) standard with a `relationships` and `included` section in the JSON response. If predictions or models aren't supplied, the NullObject that was originally  managed by ruby classes is now handled through a simple js object literal attached within `assignments`, `badges`, and `challenges`.  This allows queries (for instance `assignment.grade.points` or `assignment.prediction.predicted_points`) to still occur without a need for additional presence checks (`assignment.grade.points if assignment.grade?` etc.)

New predictions are now managed through `create` routes which are more resourceful, and both `create` and `update` routes now use strong params.

This refactor has a net reduction in almost 1,000 lines of code:
  * We no longer need to use Null Objects so they have been removed
  * We no longer need `find_or_create` methods on Prediction models
  * We no longer need serializers for prediction models since the simplified json can be build within the jbuilder views.

### Related PRs

  * unblocks #1743
  * closes #2552 
  * add TODO #2579 

### Migrations
NO

#### As a student you should still be able to:
   * Make new predictions on Assignments, Challenges, and Badges
   * Update predictions on these models
   * Verify no new Prediction models are added to the database by loading the Predictor page.  

#### As a professor you should still be able to:
  * View the generic predictor
  * View a student's predictor when in preview mode

#### Exports

  * Exporting grades should still work (see below)

* All JSON routes used by the predictor
* Removes all references to Null Objects 
* Removes all Predictor serializers and does the work within the JSON view
* CRUD operations for `PredictedEarnedGrades`, `PredictedEarnedChallenges`, and `PredictedEarnedBadges`

#### A note on Exports

Exporting grades has been updated since it used NullObjects.  I don't believe this was necessary; the call to `student.grade_for_assignment(assignment)` returns a `Grade.new` if no grade is present. This works just as beautifully as a NullGrade, because all of it's fields are valid but return default values.


app/assets/javascripts/angular/helpers/serviceHelpers.js.coffee
```diff
@@ -54,7 +55,7 @@ angular.module('helpers').factory('GradeCraftAPI', ()->
     _.each(options.include, (included)->
       return if !response.included || !item.relationships || !item.relationships[included]
       child =  _.find(response.included,
-        {id: item.relationships[included].data.id,
+        {id: item.relationships[included].data.id.toString(),
```
jwright: Why do we convert the id to a string here? For new models (e.g. using `Grade.new`) the id will be `null` and calling `toString()` on that will result in an error.

app/assets/javascripts/angular/helpers/serviceHelpers.js.coffee
```diff
@@ -90,6 +91,29 @@ angular.module('helpers').factory('GradeCraftAPI', ()->
         modelArray.push(item.attributes)
     )
 
+
+  # shared methods for predictions
+
+  createPrediction = (article, url, requestParams)->
```
jwright: I really love these methods but I feel like they should not be part of this class. I feel they should be part of a `PredictionService` class that may have this `ServiceHelper` injected in it because these methods feel too specific for a generic `ServiceHelper` class.

jwright: "These methods" being `createPrediction` and `updatePrediction`. They can turn into `PredictionService#create` and `PredictionService#update`.

app/assets/javascripts/angular/services/AssignmentService.js.coffee
```diff
@@ -31,26 +31,32 @@
...
+      _.each(assignments, (assignment)->
+        # add null prediction and grades when JSON contains none
+        assignment.prediction = { predicted_points: 0 } if !assignment.prediction
+        assignment.grade = { score: null, final_points: null, is_excluded: false } if !assignment.grade
+      )
+
+      GradeCraftAPI.setTermFor("assignment", response.meta.term_for_assignment)
+      GradeCraftAPI.setTermFor("pass", response.meta.term_for_pass)
+      GradeCraftAPI.setTermFor("fail", response.meta.term_for_fail)
+      update.predicted_earned_grades = response.meta.update_predictions
```
jwright: The data for `response.meta.update_predictions` seems like a weird name here and it is a bit odd to read this assignment. It seems like we are assigning a boolean to an array due to the naming. Does it make sense here to name the data `predicted_earned_grades` on the right side of this assignment?

app/assets/javascripts/angular/services/ChallengeService.js.coffee
```diff
@@ -33,24 +33,31 @@
...
+      return unless response.meta.include_in_predictor
+      GradeCraftAPI.loadMany(challenges,response, {"include" : ['prediction','grade']})
+      _.each(challenges, (challenge)->
+        # add null prediction and grades when JSON contains none
+        challenge.prediction = { predicted_points: 0 } if !challenge.prediction
+        challenge.grade = { score: null, final_points: null } if !challenge.grade
+      )
+
+      GradeCraftAPI.setTermFor("challenges", response.meta.term_for_challenges)
+      update.challenges = response.meta.update_predictions
```
jwright: Same comment as https://github.com/UM-USElab/gradecraft-development/pull/2578/files#diff-2aa1ff5e2c88e52d7e5800212a82c98cR45

app/controllers/api/assignments_controller.rb
```diff
@@ -0,0 +1,20 @@
+class API::AssignmentsController < ApplicationController
+  before_filter :ensure_student?, only: [:update]
+
+  # GET api/assignments
+  def index
+    @assignments = current_course.assignments
+
+    return unless current_user_is_student?
```
jwright: What does returning `nil` here do to the controller? Should we return `head: :ok` or a different status?

app/controllers/api/assignments_controller.rb
```diff
@@ -0,0 +1,20 @@
...
+    @student = current_student
+    @update_predictions = !student_impersonation?
```
jwright: I was confused by this name in the json returned. Perhaps we should call it `should_update_predictions` or some other name to show that this is a boolean value?

app/assets/javascripts/angular/services/AssignmentService.js.coffee
jwright: I see that the right side is an actual boolean value. Perhaps the right side name is the issue.

app/controllers/api/badges_controller.rb
```diff
@@ -14,17 +14,13 @@ def index
       :visible,
       :visible_when_locked)
 
-    if current_user_is_student?
-      @student = current_student
-      @update_predictions = !student_impersonation?
+    return unless current_user_is_student?
```
jwright: What does returning nil here do to the controller? Should we return head: :ok or a different status?

app/controllers/api/challenges_controller.rb
```diff
@@ -0,0 +1,34 @@
+class API::ChallengesController < ApplicationController
+  before_filter :ensure_student?, only: [:update]
+
+  # GET api/challenges
+  def index
+    @challenges = current_course.challenges
+
+    return unless include_student_data?
```
jwright: What does returning nil here do to the controller? Should we return head: :ok or a different status?

app/controllers/concerns/predictor_data.rb
```diff
@@ -1,73 +0,0 @@
-module PredictorData
```
jwright: 😂

app/models/null_grade.rb
```diff
@@ -1,110 +0,0 @@
-class NullGrade
```
jwright: 😂

app/models/null_student.rb
```diff
@@ -1,115 +0,0 @@
-class NullStudent
```
jwright: 😋

app/models/predicted_earned_badge.rb
```diff
@@ -10,14 +10,6 @@ class PredictedEarnedBadge < ActiveRecord::Base
   validates :student, presence: true
   validates :badge, presence: true, uniqueness: { scope: :student_id }
 
-  def self.find_or_create_for_student(course_id, student_id)
```
jwright: 😮

app/assets/javascripts/angular/helpers/serviceHelpers.js.coffee
acousticrobot: Great idea, I didn't like having them in here either. I'll create a new service or helper module.

app/controllers/api/assignments_controller.rb
acousticrobot: It skips the next assignments, but still renders the json.  I was getting a "use guard clause" error in Rubocop on these controller actions when I wrapped the assignments in an `if current_user_is_student?`. I'd prefer another solution if there is one though, including turning off the cop.  return feels like an antipattern to me in a Rails controller. Suggestions?

app/assets/javascripts/angular/helpers/serviceHelpers.js.coffee
acousticrobot: We won't get unsaved grade objects in here we would have to worry about, but I realize going back to the JSON API spec that the id in the relationships section should also be a string. I'm changing all of these to strings in the json, and then `item.relationships[included].data.id` will be a string and ready for matching with `_.find`

app/assets/javascripts/angular/services/AssignmentService.js.coffee
jwright: I like it.

app/controllers/api/assignments_controller.rb
jwright: So does the following work:

```
render json: {} and return unless current_user_is_student?
```

That will return a 200. I think what we really want though is this:

```
render status: :unauthorized unless current_user_is_student?
```

jwright:
@acousticrobot 

* I like `allow_updates`
* I would follow the conventions established by the app. I don't know if there is an advantage over using `return` as a guard clause or `if...else`. I think they both have their advantages and disadvantages.
* I like the `GradeCraftPredictionAPI` solution.

PR #2581, opened 2016-10-18
Remove *.sample-env.yml files (not needed anymore)

ksofa2:
### Status
**READY**

### Description
This removes some sample config files that are no longer necessary, and cause confusion.

### Migrations
NO


ksofa2:
Fixes #2582

PR #2590, opened 2016-10-19
fix-submission-view-buttons

mariehooper:
### Status
**READY**

### Description
Fixed submission view button styling and added back missing border on criterion in rubrics and removed unless_current from two context menu links.

### Migrations
NO


chcholman:
@mariehooper Perfect, thanks so much for the quick turnaround on this!

PR #2594, opened 2016-10-20
summing of completed group members bugfix

acousticrobot:
### Description
Bugfix for group member sum for unlocked conditions. 

Create group for "Group-Assignment-Unlocked-By-Submission" and submit for one student.

#### before:

screen shot 2016-10-20 at 2 30 04 pm

#### after:

screen shot 2016-10-20 at 2 29 41 pm

#### Related

Based on the description, it's possible that this shouldn't be unlocked for any student until all have submitted? Currently is unlocked for any student who submits, locked for others until they submit.


chcholman:
@acousticrobot The intended functionality here is that it should be locked for all students in a group unless they've all submitted it.

PR #2595, opened 2016-10-20
1743 predictions close

### LINKED ISSUE #1743: Unpredictable assignments if requirements have passed

acousticrobot:
### Status
**READY**

### Description

This PR improves the display of predictions for Assignments that are now closed to this student.  The conditions that cause this are: the `assignment.submissions_have_closed` and the student has not submitted anything for this assignment.

It also manages closing Assignments when conditional assignments are also closed.  This is rather ugly because conditions can be badges or assignments, and each one has to be checked for this condition in the context of this student. The best solution I could come up with was to pass the ids of all condition `Assignments` through the json, and then check each of them in Angular for `submissions_have_closed`, which sets a `is_closed_by_condition` boolean on the original assignment.

Load the predictor and verify that the following sample Assignments are displayed as earning zero points and without a slider:

  * Under Predictor Settings
    * "Past Assignment no Grade but Prediction"
    * "Not Submitted, Closed"

  * Under Unlock Settings
    * "Past-Submission-Key"
    * "Unlocked-By-Past-Submission"

* Predictor display of assignments who's submission acceptance is closed 

### Related

closes #1743

### DIFF
--- a/app/assets/javascripts/angular/directives/student_panel/article_icon.coffee
+++ b/app/assets/javascripts/angular/directives/student_panel/article_icon.coffee
@@ -4,7 +4,10 @@
 # "is_required", "is_late", "has_info", "is_locked", "has_been_unlocked", "is_a_condition", "group"
 # example:
 #   .predictor-article-icon{'ng-repeat'=>'icon in icons', 'icon-name'=>'icon', 'article'=>'assignment', 'article_type'=>'assignment'}
-# Does not handle locked/unlocked icons, which include lists of lock conditions
+#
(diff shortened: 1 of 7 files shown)


ksofa2:
@acousticrobot Looks good.

Should these assignments be getting the `status-no-points` class? They are displaying with a green background for me.

screen shot 2016-10-21 at 11 04 42 am

PR #2602, opened 2016-10-24
Fix viewable bug in grade history

jwright:
### Status

**READY**
### Description

A bug was found with the display of the grade history for a student. The bug happened when the `raw_score` or `feedback` was changed when the grade was not visible but then it was made visible the very next change.

The change checks to see if the changeset for the previous grade was not visible to the student. If it was not, the history was transformed to blank out the previous change so the student could not view the previous change.

Instead of a student seeing this: `The feedback was changed from "This needs work" to "You rock"`, to this: `The feedback was changed to "You rock"`.
### Related PRs

| branch | PR |
| --- | --- |
| 1846-fix-released-grades-bug | [https://github.com/UM-USElab/gradecraft-development/pull/2374](https://github.com/UM-USElab/gradecraft-development/pull/2374) |
### Todos
- [x] Test production data to ensure the grade in question is handled
### Migrations

NO

To test on development:
1.  Mark an assignment as "release is necessary"
2. Grade a student under that assignment by specifying a raw score or give feedback
3. Release the grade for that student
4. Log in or preview as that student and view assignment
5. Latest grade and feedback should be visible but not the grade/feedback while the grade/feedback was not released.

- Grade history as a student.

### DIFF
--- a/app/models/history_filter.rb
+++ b/app/models/history_filter.rb
@@ -9,6 +9,23 @@ class HistoryFilter
     @history = history
   end
 
+  def clear_initial_value(*attributes)
+    attrs = Array(attributes).flatten
+
+    attrs.each do |attribute|
+      @history = history.select do |history_item|
+        unless history_item.changeset[attribute].nil? ||
+               history_item.changeset[attribute][0].nil?
+          history_item.changeset[attribute][0] = ""
+        end
+        history_item
+      end
+    end
+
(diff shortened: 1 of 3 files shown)


jonathoy:
Looks good! Sorry for the delay on this. I will be likely taking some cues from this PR soon for change tracking of submission drafts.

PR #2608, opened 2016-10-26
2456 recalculate grade scheme changes

### LINKED ISSUE #2456: Recalculate scores when adding new Grade Scheme Element
When a new Grade Scheme Element is added, scores for students who fall in the Element's point range should be recalculated so that their level is updated to the newly added Grade Scheme Element.

jonathoy:
### Status

READY
### Description

When an instructor adds, modifies, or deletes grade scheme elements for a course, affected students should have their scores recalculated.
### Migrations

NO

`GradeSchemeElements` and `earned_grade_scheme_element_id`s on `course_memberships`

Fixes #2456

### DIFF
--- a/app/controllers/grade_scheme_elements_controller.rb
+++ b/app/controllers/grade_scheme_elements_controller.rb
@@ -48,6 +48,7 @@ class GradeSchemeElementsController < ApplicationController
     end
     respond_to do |format|
       if @course.save
+        @course.students.pluck(:id).each { |id| ScoreRecalculatorJob.new(user_id: id, course_id: @course.id).enqueue }
         format.json do
           render json: current_course.grade_scheme_elements.select(
             :id,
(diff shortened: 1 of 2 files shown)


jonathoy:
@jwright I noticed that there was some weirdness with how params were coming in after I rebased this with master, which now has the Rails 5 changes. I had to make some changes similar to what you did for 14adc3b492116a2680b9bd120fd9fda43b8b473f. Could you explain why these changes were necessary?

jonathoy:
I think that if we made this page AJAXy (AJAXified?) so that these grade scheme elements updated on the fly, it would simplify this logic of having to find diffs from the current and the future changes each time the form is submitted.

Me and @chcholman had discussed the idea of just updating all the students in the course rather than just a subset. My guess is that you're right and that these grade scheme elements aren't often changed. If that is the case, it would definitely be easier just to go with your suggestion in the meantime.

ksofa2:
@acousticrobot :

> I was hoping once we removed `high_points` from the Grade Scheme Elements we could then lay the groundwork for updating GSEs individually rather than being bound to submitting a giant form and processing them all at once. I'm curious @ksofa2 to hear your thoughts on this since you were recently in that code and if that still seems like a good idea.

I think that does still seem like a good long-term approach; and concur with the agreed fix for this PR.

PR #2612, opened 2016-10-28
2597 earned badge refactor

### LINKED ISSUE #2583: Earned Badge Lifecycle
### Actual Behavior

Earned Badges have optional associations `grade`, `assignment`, and `level`.  Badges earned for a level store the `level` and the `assignment`. For a `grade` store only the grade.
### Expected Behavior

All should store the `assignment`, both should store the `grade`.
### Additional concerns.

There is a hack in place in the service that destroys and recreates all EarnedBadges on rubric submission.  These should be persisted and updated on resubmission rather than recreated. This should also target the `level_id` so we don't destroy the badges that were earned on the grade.

We should also verify that all badges associated with a grade are re-saved when a grade is released, since their visibility is dependent on the grade's visibility.

### LINKED ISSUE #2597: Award Earned Badges from Levels in the Service
### Expected Behavior

Students who earn a level on a criteria should be awarded all badges associated with the level via a level badge. These badges should be created or updated when the grade, or related criterion grades, are updated.
### Actual Behavior
- Students are only awarded badges that come through in the params (subject to errors if the badges don't load in Angular)
- All earned badges for a level are deleted rather than recreated, rather than updated.

acousticrobot:
### Description

Updates to the EarnedBadge lifecycle.
- reducing columns on the earned badge model (submission_id, assignment_id)
- reducing the fields permitted in the earned badge params
- managing visibility and course_id within the earned badge model on save
- updating visibility for associated earned badges whenever a grade is saved
- updating earned badges to always be visible when not associated with a grade. (Managing this state on save, rather than from a hidden field on the badge award form.) 
- awarding EarnedLevelBadges from backend and updating rather than deleting and recreating.

See the updated [EarnedBadge Wiki Page](https://github.com/UM-USElab/gradecraft-development/wiki/Earned-Badges) for full documentation on expected behavior.
### Migrations

YES!

Award badges from all paths (see wiki). Confirm updates to Grade visibility trigger student badge visibility 

All areas that report or manage EarnedBadges, including Badge pages that include earned badge count. 

closes #2597 
closes #2583


app/views/earned_badges/_form.html.haml
```diff
@@ -4,7 +4,6 @@
   .italic= "You are awarding the #{@badge.name} #{term_for :badge} to..."
 
   = f.hidden_field :badge_id, value: @badge.id
-  = f.hidden_field :student_visible, value: true
```
jwright: 👍

spec/models/earned_badge_spec.rb
```diff
@@ -30,7 +30,17 @@
...
+    let(:student) { create :user }
+
+    it "returns badge full points for points" do
+      badge = create(:badge, full_points: 123)
+      earned_badge = EarnedBadge.create(badge_id: badge.id, student_id: student.id)
+      expect(earned_badge.points).to eq(123)
+    end
+  end
+
+  describe "#ernable" do
```
jwright: Mis-spelling `ernable`.

spec/models/grade_spec.rb
```diff
@@ -235,31 +235,51 @@
...
+      expect { Grade.find_or_create_grades(assignment.id, ids) }.to \
         change{ Grade.count }.by(ids.length)
     end
   end
+
+  describe "after_save actions" do
+    let(:course) { create :course }
+    let(:assignment) { create :assignment, course: course }
+    let(:grade) { create :grade, assignment: assignment }
+
```
jwright: A few extra line breaks here.

spec/models/grade_spec.rb
```diff
@@ -235,31 +235,51 @@
...
     end
 
     it "creates a grade for assignment and student if none exists" do
-      expect { Grade.find_or_create_grades(world.assignment.id, ids) }.to \
+      expect { Grade.find_or_create_grades(assignment.id, ids) }.to \
         change{ Grade.count }.by(ids.length)
     end
   end
+
+  describe "after_save actions" do
```
jwright: I would change this to less "developer-y" to something like "when it is saved". That way it would read in the output: "Grade when it is saved updates earned badge visibility".

app/controllers/earned_badges_controller.rb
```diff
@@ -110,12 +99,24 @@ def destroy
       notice: "The #{@badge.name} #{term_for :badge} has been taken away from #{@student_name}."
   end
 
+  # Used for Badges Backpack integration
+  # GET /courses/:course_id/badges/:badge_id/earned_badges/:id/confirm_earned
+  def confirm_earned
+    @course = Course.find(params[:course_id])
```
acousticrobot: I think so -- I'm going to pass this on to @jonathoy since this is his work and has to do with fulfilling the open badges standard (I just moved the method below the standard CRUD ones). I also wonder if it should be moved into the `/api` namespace?

app/controllers/earned_badges_controller.rb
```diff
@@ -110,12 +99,24 @@ def destroy
...
+    @badge = @course.badges.find(params[:badge_id])
+    @earned_badge = @badge.earned_badges.where(id: params[:id]).first
+    if @earned_badge.present?
+      render nothing: true, status: 200
+    else
+      redirect_to root_path
```
acousticrobot: Same as above -- I'll make a new ticket and assign to @jonathoy.

app/controllers/earned_badges_controller.rb
jonathoy: This can be made as an api controller method with the variable made local. We haven't been getting any feedback from the OpenBadge devs, so this method isn't being used at this point since the integration isn't finished. It can probably be pulled out for now or assigned to me for clean up if we don't want it here.

acousticrobot: Great -- I have copied these suggestions into ticket #2621

app/models/grade.rb
```diff
@@ -166,6 +167,10 @@ def cache_associations
     self.course_id ||= assignment.try(:course_id)
   end
 
+  def update_earned_badges
+    self.earned_badges.reload.each(&:save)
```
acousticrobot: I think this is a good idea. Since it moved beyond the scope of this PR and the bugs it addresses, I have created a new issue for #2634 and we can leave it as is for now.

PR #2620, opened 2016-10-31
2611 assignment order

### LINKED ISSUE #2611: Assignments should be ordered by position in Grade Predictor
### Expected Behavior

Assignments should appear in the Grade Predictor in the order in which the instructor has ordered them - the `ordered` scope, reflecting the stored `position`
### Actual Behavior

Assignments are appearing in random order
### Steps to reproduce

Load up the predictor preview in production

acousticrobot:
### Description
Adds ordered to assignments in predictor, so that they are ordered by position to address #2611. 

Also adds a couple refinements to use of `position`: 
  * makes this field `null: false` for `Assignments` and `Badges` to align with the field in `AssignmentTypes`
  * Also adds a spec in Assignments as rudimentary documentation as to how the position works via the acts as list gem
  * Adds position to the samples for `Sorting` so that they can be more effectively used for testing. I'm not sure if we sort by Alfanumeric anywhere still, so I left this part of the samples intact.

closes #2611 

### Migrations
YES

run samples and confirm that they are in the correct order (by position) in the Predictor `Sorting Settings` section. Reorder Assignments as a Professor, and confirm that the order updates on refresh of the predictor.

Assignment ordering

### DIFF
--- a/app/assets/javascripts/angular/templates/predictor/article.html.haml
+++ b/app/assets/javascripts/angular/templates/predictor/article.html.haml
@@ -18,7 +18,6 @@
     %predictor-article-assignment-content{'ng-if' => 'article.type=="assignments"', 'article'=>'article'}
     %predictor-article-assignment-content{'ng-if' => 'article.type=="challenges"', 'article'=>'article'}
 
-  .predictor-position {{article.position}}
   .predictor-article-main-icons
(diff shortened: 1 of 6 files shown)


chcholman:
@acousticrobot Awesome fix! I confirmed that all `Badges` and `Assignments` have values for position, so this migration should proceed smoothly.

PR #2632, opened 2016-11-01
2617 - Fix no method for nil on reflection call

### LINKED ISSUE #2617: NoMethodError: undefined method `_reflect_on_association' for nil:NilClass
https://rollbar.com/gradecraft/gradecraft-development/items/1714/

jwright:
### Status
**READY**

### Description
This PR attempts to fix an issue when using `find_by` on a relationship in Rails 5 by replacing it with a `where#first` call.

This error occurred when a user is signed in and it attempts to find the current course they have visited last.

I am unsure if this fixes the issue because I was unable to replicate it locally. I traced down the issue and I believe to be from using the `find_by` method with associations.

### Migrations
NO

1.  Visit the home page while being signed on

* Course switcher

Fixes #2617

### DIFF
--- a/app/models/course_router.rb
+++ b/app/models/course_router.rb
@@ -8,7 +8,7 @@ class CourseRouter
 
     def current_course_for(user, current_course_id=nil)
       return nil if user.nil?
-      course = user.courses.find_by(id: current_course_id) unless current_course_id.nil?
+      course = user.courses.where(id: current_course_id).first
       course ||= user.current_course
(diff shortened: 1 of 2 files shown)


chcholman:
@jwright I know it wasn't possible to reproduce locally - would you recommend merging and testing in prod?

jwright:
@chcholman Yes. This will not negatively affect the current implementation.

PR #2639, opened 2016-11-02
2633 canvas api error response

### LINKED ISSUE #2633: JSON::ParserError
### Actual Behavior
When importing an assignment, the import hits a JSON error as documented here: https://rollbar.com/gradecraft/gradecraft-development/items/1644/?item_page=0&#traceback

### Steps to reproduce
I know that an instructor and their TA both hit this repeatedly around the same assignment, I don't fully know what steps they took to create it. Will update if/when I can find out more.

jwright:
### Status
**HOLD**

### Description
Gracefully handle any Canvas communication issues in the controller. This adds an exception handler block for any canvas call so the controller can redirect to a non-canvas controller method and display an alert to the user saying that the communication with Canvas did not work.

**WAITING ON WHAT TO DO ABOUT GETTING THE ACTUAL ERROR FROM CANVAS**

### Migrations
NO

1.  Login as an instructor
1. Link the Canvas course 97697 with any GC course
1. Import grades from assignment 103424
1. Should see error message and the error should be gracefully handled

* Linking Canvas courses
* Canvas importing of assignments
* Canvas importing of grades

Closes #2633

### DIFF
--- a/app/controllers/api/grades/importers_controller.rb
+++ b/app/controllers/api/grades/importers_controller.rb
@@ -15,8 +15,14 @@ class API::Grades::ImportersController < ApplicationController
   def show
     @assignment = Assignment.find params[:assignment_id]
     @provider_name = params[:importer_provider_id]
-    @grades = syllabus.grades(params[:id], [params[:assignment_ids]].flatten, nil, false, importer_params.to_h)
-    @provider_assignment = syllabus.assignment(params[:id], params[:assignment_ids])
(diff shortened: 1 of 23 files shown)


jwright:
@chcholman I have a question on the story for this PR.

chcholman:
@jonathoy Will you take a look at this work when you have a chance and see how it matches against our current Canvas work?

PR #2662, opened 2016-11-08
2630 misc ui updates

mariehooper:
### Status
**READY**

### Description
- [x] Remove underline under the breadcrumb slashes
- [x] Make the opacity full on badges on instructor side regardless of earned/awarded
- [x] Change name of ‘Teams’ everywhere to ‘Sections’ (updated `team_term` default in schema)

* Instructor badge show page 
* Instructor pages with breadcrumbs
* Instructor team pages

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -250,7 +250,7 @@ ActiveRecord::Schema.define(version: 20161107123814) do
     t.boolean  "has_badges",                                              default: false,                        null: false
     t.boolean  "has_teams",                                               default: false,                        null: false
     t.string   "student_term",                                            default: "Student",                    null: false
-    t.string   "team_term",                                               default: "Team",                       null: false
+    t.string   "team_term",                                               default: "Section",                       null: false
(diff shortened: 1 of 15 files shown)


db/schema.rb
```diff
@@ -250,7 +250,7 @@
     t.boolean  "has_badges",                                              default: false,                        null: false
     t.boolean  "has_teams",                                               default: false,                        null: false
     t.string   "student_term",                                            default: "Student",                    null: false
-    t.string   "team_term",                                               default: "Team",                       null: false
+    t.string   "team_term",                                               default: "Section",                       null: false
```
chcholman: @mariehooper Rather than manually changing this, you need to write a migration that persists this change. Let me know if you want to walk through that process!

chcholman:
@mariehooper Thanks so much for this!

Changes to make before merging:
[ ] Add a migration to persist the Teams -> Section change
[ ] Update the specs regarding the Teams -> Section change so that the test suite passes

Issue #2650, opened 2016-11-07
Display auditors on Students page


chcholman:
As a faculty member, I need to be able to see the student-auditors on the Students index page.

PR #2670, opened 2016-11-11
Auditors

chcholman:
### Status
**READY**

### Description
This work adds an "auditing" filter to the students page. This closes Issue #2650

In the process it:
* consolidates the methods to return students who are being graded/not for each course. 
* removes unused methods on the `Course` model 
* removes the on the `SubmissionExport` performer
* renames `cached_score_for_course` to `score_for_course`

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -79,6 +79,7 @@
 //= require pass_fail
 //= require student_dashboard
 //= require student_onboarding
+//= require students_table_filter
 //= require public_pages
 //= require levels_per_assign
 //= require earned_badges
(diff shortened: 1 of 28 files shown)


mariehooper:
@chcholman  - should be all set.

chcholman:
@acousticrobot So I did purposefully update font-awesome in order to get access to one of their newer icons - we don't specify a particular version in the Gemfile, but that update did update the version in the Gemfile.lock (and that triggered the update of nokogiri and rake). Preference on how we handle?

chcholman:
@acousticrobot Perfect, thank you!

PR #2676, opened 2016-11-15
Group Submissions Exports

zazaian:
### Status
**NEEDS REVIEW**

### Description
This PR adds the ability to generate submissions exports for group-submitted assignments. Students have been decoupled from the submissions export generation workflow and are now referred to "submitters".

Additional logic has been added to the Group and User models to clarify the roles each plays in the guise of a submitter. Relevant queries in Assignment for fetching appropriate resources have also been merged to reduce duplication.

### Deploy Notes
Migrations must be run on this branch prior to deployment.

### Migrations
YES

1. Visit an assignment that has group submissions.
2. Click the 'Options' button, then 'Download Group Submissions'.
3. Visit 'Course Data Exports' on the left-hand navigation menu.
4. A new submissions export for the given assignment should appear. The user requesting export will receive an email within :30 seconds or so of request depending on how many files are attached. Refreshing the page will offer a download link for the completed export.

* Group assignments.


db/migrate/20140910171347_transfer_role_to_course_memberships.rb
```diff
@@ -1,17 +0,0 @@
-class TransferRoleToCourseMemberships < ActiveRecord::Migration
```
acousticrobot: I'm not sure why you are deleting this migration?

db/migrate/20161013144318_add_group_id_to_submissions_export.rb
```diff
@@ -0,0 +1,5 @@
+class AddGroupIdToSubmissionsExport < ActiveRecord::Migration
+  def change
```
acousticrobot: Since this hasn't been merged in by anyone else, would it make a cleaner history to remove this migration, and then simplify the next one that undoes this work?

db/migrate/20161026131150_rename_submissions_export_student_ids.rb
```diff
@@ -0,0 +1,11 @@
+class RenameSubmissionsExportStudentIds < ActiveRecord::Migration
+  def up
+    rename_column :submissions_exports, :student_ids, :submitter_ids
+    remove_column :submissions_exports, :group_ids
```
acousticrobot: Same comment as above -- since there is no benefit to keeping this history, could we rewrite all these migrations to make the minimum number of changes needed between master and this branch's final schema?

db/migrate/20140910171347_transfer_role_to_course_memberships.rb
zazaian: Hey there Jonathan,

I removed this because it's not really a migration so much as a rake task that is being performed through the migration pattern. If the User table has not been created in the db:reset process before this particular migration is run, then this migration will fail and break the reset. This was happening to me, and needed to be removed in order to complete a full reset of the database.

zazaian:
Hey there Jonathan, I've merged the migrations as requested, and resetting the db with `rake db:migrate:reset` seems to be working fine now. If I was the only one experiencing the issue with the User migration you mentioned, I can reset that additional file, though it seems stale and unnecessary to leave in the repository.

zazaian:
Hey there Jonathan, I've removed the `20161024121501_change_group_id_to_group_ids.rb` migration which deals with `group_ids`. There should be no `group_ids` or `group_id` column on SubmissionsExport. The only changes we need are the addition of `use_groups` and renaming `student_ids` to `submitter_ids`. This should be taken care of by removing the aforementioned migration.

zazaian:
Hey guys -- sorry, I must've had a bogus connection at some point and the push bounced. I've rebased it against current master and pushed it. Should show the removed group_id migration now.

Issue #2498, opened 2016-09-29
file attachments for group grades


acousticrobot:
Currently attachments for group grades are not supported. This should follow the pattern on individual grades, and will have to be build for both the standard and rubric-graded grades. It is possible that the Angular directive used on rubric grades could work for both.  The file will have to be built and saved for each student's grade.

acousticrobot:
Grade file saving assumes a `grade_id`. If we remove `find_or_create_grade` methods, we will need to post new `GradeFiles` in a manner that also creates the `Grade`.

acousticrobot:
Current solution will use `GradeFileAssociation` model to tie one grade file to many Grades. This way file attachments can be deleted for the group.  Prep work is in PRs #2682 and #2683.

acousticrobot:
Once #2750 is tested and merged, we can continue with this work, since all file uploads for grade will be managed from one angular directive.

acousticrobot:
Step 1 (Complete)

FileUpload
----------
grade_id

Attachment
----------
grade_id
file_upload_id

Step 2:

Run the rake task from step one, and clean up
the creation of Attachments and File Uploads in the controller

Grade
-----
has_many :file_uploads, through :attachments

Attachment
----------
grade_id
file_upload_id

Now we can do group file attachments

Step 3:

Move all file upload types into one model:

Attachables
-----------
Grade
Submission
etc.

Attachment
----------
file_upload_id
attachable_id
attachable_type
belongs_to :attachable, polymorphic: true

FileUpload
----------
id
path
file
filename
...
and all methods needed for different attachable types

acousticrobot:
### Step 1 (Complete)

FileUpload
  grade_id

Attachment
  grade_id 
  file_upload_id

### Step 2:

Run the rake task from step one, and clean up
the creation of Attachments and File Uploads in the controller

Grade
  has_many :file_uploads, through :attachments

Attachment
  grade_id
  file_upload_id

Now we can do group file attachments

### Step 3:

Move all file upload types into one model:

Attachables
  Grade
  Submission
  etc.

Attachment
  file_upload_id
  attachable_id
  attachable_type
  belongs_to :attachable, polymorphic: true

FileUpload
  id
  path
  file
  filename
  ...
  and all methods needed for different attachable types

PR #2682, opened 2016-11-21
2498 group grade files

acousticrobot:
### Status
**READY**

### Description

This PR extracts and Angular `GradeService` functionality out of the `RubricService`, so that it can be used for both Standard and Rubric Grading.

The end result is that grade file uploads for the standard grade form and the rubric grade form are handled by the same angular directive, and modest cleanup of the UX should make file deletion a little more intuitive.

It also:

  * Reduces the init data for `GradeCtrl` to three simple ids
  * Removes the "Student" or "Group" nested in the assignment object, passing them as `recipient` instead
  * Removes the `AssignmentScoreLevel` and `Grade` Angular factories, relying on the same objects used by other Angular controllers for these models.
  * Adds a single Assignment show route to the API, reuses a jbuilder partial for show and index.
  * Improves the UX for awarding custom points when an assignment has score levels
  * fixes multiple bugs with assigning a custom values

### Related PRs
Prestages Group Grade File Functionality #2498

I decided to break these two apart since this works as-is and is already a pretty big PR.

Standard and Rubric grading forms should still work as before.

Changing a grade with score levels to a custom points value should still work, but with a better UX.

Attachments to standard and rubric grades should look and behave the same.


app/assets/javascripts/angular/controllers/GradeCtrl.js.coffee
```diff
@@ -1,36 +1,38 @@
...
-      $scope.assignmentScoreLevels.push scoreLevelPrototype
-    )
+@gradecraft.controller 'GradeCtrl', ['$scope', 'AssignmentService', 'GradeService', ($scope, AssignmentService, GradeService) ->
+
+  $scope.grade = GradeService.grade
+
+  $scope.assignment = ()->
+    AssignmentService.assignments[0]
+
+  $scope.init = (assignmentId, reciptientType, reciptientId)->
```
jonathoy: Being super nitpicky here, but this looks like an unintentional misspelling for recipient.

app/assets/javascripts/angular/controllers/GradeRubricCtrl.js.coffee
```diff
@@ -8,15 +8,12 @@
...
-  # Criterion factory is dependent on CriterionGrades existing in scope
-  $scope.init = (assignment_id, scope_type, scoped_id)->
-    $scope.assignment = {
-      id: assignment_id,
-      scope: {
-        type: scope_type,
-        id: scoped_id
-      }
-    }
+  $scope.init = (assignmentId, reciptientType, reciptientId)->
```
jonathoy: Same nitpick as before, sorry!

acousticrobot: Good catch thanks! Fixed.

app/assets/javascripts/angular/controllers/GradeCtrl.js.coffee
acousticrobot: Not nitpicky -- good catch!

chcholman:
@acousticrobot Testing this locally, submitting a group rubric grade seems to fail (the form doesn't submit, and the values aren't recorded). Will you test that? Aware that this may or may not be coming from this update, and that we don't have sample data/test coverage here.

chcholman:
@acousticrobot Awesome, thanks so much! Looks great. Merging, and then will test in staging before release.

PR #2683, opened 2016-11-22
add association for grade files

acousticrobot:
### Status

READY

### Description (updated 12/12/16)

This completes Phase One of moving GradeFiles into a polymorphic File record.

This PR renames `GradeFile` to `FileUpload`, the first step in moving all `model_file` types (submission, badge, assignment, challenge) into one file model.

It also adds a new `Attachment` model that will be used to manage one `FileUpload` for many (group) `Grades`. This will allow a `FileUpload` to be created and attached or deleted and removed from all grades for that Assignment and Group. FileUpload for individual assignments will also have one `Attachment` tying the `Grade` to the `FileUpload`.

It does not yet use the Attachment -- first we add the new model and maintain new model creation and deletion, but leave current functionality in place until we are able to run the included rake task.  Then Phase Two can be merged, which will transfer the relationship management fully over to the new `Attachment` model.

### Deploy Notes

This PR should work immediately on production, but before we can move on to the next task, the following rake task should be run:

`rake attachments:add_associations`

This will copy grade ids from each `FileUpload` onto an `Attachment`.

### Migrations

YES

* Attaching GradeFiles to Grades.
* Every reference to `GradeFile` or `grade_file`.
* s3 storage location for grade files (should remain the same).


db/schema.rb
```diff
@@ -236,7 +236,6 @@
     t.string   "role",                           default: "student", null: false
     t.boolean  "instructor_of_record",           default: false
     t.integer  "earned_grade_scheme_element_id"
-    t.index ["course_id", "user_id"], name: "index_course_memberships_on_course_id_and_user_id", unique: true, using: :btree
```
jwright: Do you know why this line is in the schema since there is no migration for it?

jwright:
@acousticrobot My question is why not turn `GradeFile` into a link table and have a `File` model and table?

```
# schema
create_table "grade_files", force: :cascade do |t|
  t.integer "grade_id",      null: false
  t.integer "file_id", null: false
end

create_table "files", force: :cascade do |t|
  t.string   "filename"
  t.string   "filepath"
  t.string   "file"
  t.boolean  "file_processing", default: false, null: false
  t.datetime "created_at"
  t.datetime "updated_at"
  t.string   "store_dir"
end
```

I always thought that is how it should be and then later on, we could move `assignment_files`, `submission_files`, etc.

My concern is that now we have another association on grades. Thoughts?

db/schema.rb
acousticrobot: ### re: line in schema 

I just checked -- if I run migrations on master, this line is also removed on my machine.

I commented on this line when it was added in a previous PR w/o a corresponding migration, perhaps it was merged into master w/o being removed?

There seems to be some problem with the indexing of these fields on course_membership committed to the schema. On master I see these three lines in a row:

```
    t.index ["course_id", "user_id"], name: "index_course_memberships_on_course_id_and_user_id", unique: true, using: :btree
    t.index ["course_id", "user_id"], name: "index_courses_users_on_course_id_and_user_id", using: :btree
    t.index ["user_id", "course_id"], name: "index_courses_users_on_user_id_and_course_id", using: :btree
``` 

Is there any reason to have more than one index on these two fields?  Could it be an error with the auto generation of the name of the index? I think we should figure out what the problem is and fix it on master first, then I'll rebase this branch.

### re: GradeFile -> File

I think that is a great idea and was wondering if that was a better way to go too. I see no reason to have different types of file models in the long term.

The one thing we will have to be very careful about though is the 3s storage directory naming if we change these models. Particularly vulnerable are the methods `AttachmentUploader#file_klass` and `#owner_name` which are used to create buckets like:

```
submission_file: uploads/<course-name_id>/assignments/<assignment-name_id>/submission_files/<student-name | group-name>/timestamp_file-name.ext
grade_file: uploads/<course-name_id>/assignments/<assignment-name_id>/grade_files/<timestamp_file-name.ext>
```

I'm not sure what the best way to make this change and assure we don't loose the files we have and keep putting the new ones in the same directories. Thoughts?

jwright: ## re: line in Schema

It looks like it was a naming thing from a `course_users` table. I think those can be removed and I think it should be done in master.

## re: GradeFile -> File

We can still keep those methods by overriding the `Uploader` methods on the class to keep the `file_klass` in tact. Let me know if that sounds doable.

app/controllers/api/grades/grade_files_controller.rb
```diff
@@ -6,25 +6,29 @@ class API::Grades::GradeFilesController < ApplicationController
   def create
     grade = Grade.find(params[:grade_id])
 
-    @grade_files = []
-    params[:grade_files].each do |f|
-      @grade_files << GradeFile.create(file: f, filename: f.original_filename[0..49], grade_id: grade.id)
+    @file_attachments = []
```
jwright: The readability of this could be improved a bit I think (if you agree this is more readable):

```
@file_attachements = params[:file_attachments].map do |f|
  grade.file_attachments.create(file: f, filename: f.original_filename[0..49])
end
```

ActiveRecord will take care of the relationships here and there is no need to create both a `FileAttachment` and a `GradeFile`.

app/controllers/api/grades_controller.rb
```diff
@@ -4,12 +4,19 @@ class API::GradesController < ApplicationController
 
   # GET api/assignments/:assignment_id/students/:student_id/grade
   def show
-    @grade = Grade.find_or_create(params[:assignment_id], params[:student_id])
-    @grade_files = GradeFile.where(grade_id: @grade.id)
-    if @grade.assignment.release_necessary?
-      @grade_status_options = Grade::STATUSES
+    if Assignment.exists?(params[:assignment_id].to_i) && User.exists?(params[:student_id].to_i)
+      @grade = Grade.find_or_create(params[:assignment_id], params[:student_id])
+      @file_attachments = FileAttachment.where(grade_id: @grade.id)
```
jwright: I think we should use `@grade.file_attachments` here instead of `FileAttachment.where`. I think this violates tell, don't ask.

app/controllers/grades_controller.rb
```diff
@@ -122,10 +122,9 @@ def grade_params
       :assignment_id, :assignment_type_id, :assignments_attributes, :course_id,
       :earned_badges_attributes, :excluded_by_id, :excluded_at, :excluded_from_course_score,
       :feedback, :feedback_read, :feedback_read_at, :feedback_reviewed, :feedback_reviewed_at,
-      :final_points, :grade_file_ids, :grade_files_attributes, :graded_at, :graded_by_id,
+      :final_points, :graded_at, :graded_by_id,
       :group_id, :instructor_modified, :is_custom_value, :pass_fail_status,
-      :full_points, :raw_points, :student_id, :submission_id, :team_id, :status,
-      grade_files_attributes: [:id, file: []]
```
jwright: Don't we need the `file_attachment_attributes` somewhere?

app/models/file_attachment.rb
```diff
@@ -0,0 +1,27 @@
+# This class currently handles grade files, it should be expanded to handle
+# submission files and badge files to reduce the number of models needed
+# to handle associated files. It is named FileAttachment since "File" is
+# already a ruby class.
+
+class FileAttachment < ActiveRecord::Base
+  include S3Manager::Carrierwave
+
+  belongs_to :grade, inverse_of: :file_attachments
```
jwright: This really should be `file_attachable` and made polymorphic here but I see your comment above and we can do this at a later time if you feel that is more appropriate when it handles `BadgeFiles`, etc. 

However, I fear we may loose that familiarity if we do it at a later time. If we move this to be polymorphic and move it into a concern for `Grade`, we will be able to just follow the same pattern for `Badge`, etc. If we do it later, we will have to come back to the pattern which may take longer. It's really a toss-up between pre-optimization and some thrashing that may occur later.

app/models/file_attachment.rb
```diff
@@ -0,0 +1,27 @@
...
+
+  validates :filename, presence: true, length: { maximum: 50 }
+
+  mount_uploader :file, AttachmentUploader
+  process_in_background :file
+
+  def course
```
jwright: `#course` and `#assignment` should be moved to `GradeFile` or this will be unusable as a polymorphic table/model. `klass_name` should not have to be used as we should have different uploaders.

Same caveat as above.

app/models/grade.rb
```diff
@@ -28,11 +28,11 @@ class Grade < ActiveRecord::Base
   after_save :update_earned_badges
 
   clean_html :feedback
-  multiple_files :grade_files
+  multiple_files :file_attachments
   releasable_through :assignment
 
-  has_many :grade_files, dependent: :destroy, inverse_of: :grade
-  accepts_nested_attributes_for :grade_files
+  has_many :file_attachments, dependent: :destroy, inverse_of: :grade
```
jwright: This could be made polymorphic and moved into a concern.

app/models/grade_file.rb
```diff
@@ -1,18 +1,4 @@
...
-  process_in_background :file
-
-  def course
-    grade.course
-  end
-
-  def assignment
-    grade.assignment
-  end
+  belongs_to :grade
```
jwright: Once `FileAttachment` is made polymorphic, this should have it's own uploader that inherits from a base uploader only to deal with `GradeFile`s though.

db/migrate/20161121202423_add_grade_file_association.rb
```diff
@@ -0,0 +1,9 @@
+class AddGradeFileAssociation < ActiveRecord::Migration[5.0]
+  def change
+    rename_table :grade_files, :file_attachments
+    create_table :grade_files do |t|
+      t.integer :grade_id, null: false
```
jwright: If this were polymorphic, the `grade_id` would be `file_attachable_id` and we would add `file_attachable_type`.

lib/tasks/grade_files.rake
```diff
@@ -0,0 +1,11 @@
+namespace :grade_files do
+
+  desc "Add grade file association for many to many relationship"
+  task add_associations: :environment do
+    FileAttachment.find_each(batch_size: 500) do |gf|
+      if GradeFile.where(file_attachment_id: gf.id, grade_id: gf.grade_id).empty?
```
jwright: I am not sure the performance difference if there is one, but this query would pull down all `GradeFile`s with the parameters above but if we used `.exists?`, it would do a SQL query to check for existance`.

spec/toolkits/uploaders/attachment_uploader.rb
```diff
@@ -19,6 +19,10 @@ def assignment
 
           def owner_name
           end
+
+          def klass_name
```
jwright: If `FileAttachment` were polymorphic, this could be the base uploader and this method would through a `NotImplementedError `. 

A new `GradeFileAttachmentUploader` could inherit from `AttachmentUploader` and define this method as `grade_file`.

PR #2686, opened 2016-11-28
report errors on failed grade save

### LINKED ISSUE #2659: Properly handle too-long filenames for instructor grade attachments
### Expected Behavior
As an instructor, if I upload a file with a name that is too long, the app should shorten the filename. Additionally, we need to display errors back to users on the grade form page as encountered. 

### Actual Behavior
Currently, the file appears to be uploading (percentage grows in the bottom left of the Chrome browser), and then the page reloads without the attachment present.

acousticrobot:
### Status
**READY**

### Description

Adds an error message when there is a failure saving a grade.

You will need to submit a grade that won't save to see the message. I was able to do this by submitting a grade file with a filename over 50 chars, but there might be other errors that can be triggered.

screen shot 2016-11-28 at 12 08 10 pm

closes #2659

### DIFF
--- a/app/controllers/grades_controller.rb
+++ b/app/controllers/grades_controller.rb
@@ -51,7 +51,7 @@ class GradesController < ApplicationController
     else # failure
       redirect_to edit_grade_path(grade),
         alert: "#{grade.student.name}'s #{grade.assignment.name} was not successfully "\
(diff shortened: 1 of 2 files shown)


chcholman:
@acousticrobot Looks great, thanks so much!

PR #2694, opened 2016-12-02
add groups to samples

acousticrobot:
### Status
**READY**

### Description

This PR adds groups to the sample data, so that we don't have to manually assign groups to test them.

It will assign the groups to an assignment only if `assign_groups: true` is set on the assignment.

### Considerations

Before merge, I could easily create another flag on each course, if it would be helpful to have a course with no pre-existing groups.

I set it to assign groups to all current group grading assignments, and added one new assignment without groups assigned. I could change this to have groups assigned to fewer assignments if this would be better.

It does not handle grading group assignments -- this might also be helpful in a future PR.

### DIFF
--- a/db/samples.rb
+++ b/db/samples.rb
@@ -34,6 +34,29 @@ def add_unlock_conditions(model, config, course_config)
   end
 end
 
+def create_groups(course_name, assignment)
+  group_names = ["The Clique", "The Cabal", "The Household", "The Community",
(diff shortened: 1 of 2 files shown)


chcholman:
@acousticrobot This is awesome, thanks so much for this. I appreciate the callout regarding potentially creating a class without groups, but think we should be fine without implementing that one for now. Yes, let's definitely do some group grades in the future!

PR #2709, opened 2016-12-07
1198 autosave student submissions

### LINKED ISSUE #1698: Add unique validation around student submissions for an assignment
As a student, I should only be able to create one submission for an assignment.

jonathoy:
### Status
Ready

### Description

screen shot 2016-12-07 at 2 10 43 pm

This PR describes a series of changes that were made to the `Submission` model in order to allow for the saving of draft content, specifically `text_comment`. The newly added field, `text_comment_draft`, is intended for this purpose.

For this PR, a submission draft will be referred to as a `Submission` with `text_comment = nil`, `link = nil`, `submitted_at = nil`, with no `submission_files`. This method has been defined on the submission model as `draft?`.

To make the new submission draft concept more apparent, there is a SubmissionProctor that was added for handling `viewable?` status.

The endpoints as described in `API::Assignments::SubmissionsController` are meant for the checking of saved content on page load as well as the creating and updating of draft submissions.

Because the affected submission form was not originally written in Angular, there are various Angular components that were added that are managed by an instance of the `StudentSubmissionService`. Text content is saved on a change event after a 3.5sec debounce and can also be manually triggered by clicking on the Save Draft button.

### Related PRs
An migration included in this PR closes #1698 

### Deploy Notes
Two migrations exist: One for adding the `text_comment_draft` field to the `Submission` table and one for adding a unique constraint, limiting submissions to one and only one per `Student` and `Assignment`.

In order for the latter to run, preexisting submission data must be cleaned up such that these conditions are met.

### Migrations
Yes

Assume that this PR affects all places where Submissions are referenced.

- Ensure that a submission can be submitted with and without a preexisting draft
- Verify that any submission stats are not taking submission drafts into account
- Check that submission exports do not return submission drafts
- Verify that instructors are not be able to view draft content
- Verify that all group members on a group submission are be able to view/edit autosaved content
- Ensure mail notifications are sent under the proper conditions

Again, assume that this PR affects all places where Submissions are referenced. One should expect to be able to properly create, edit, and delete submissions from the app both as an instructor and as a student where required.


jonathoy:
@acousticrobot I feel that I must apologize in advance for the size of this PR. Feel free to ask me questions if you find that there is ambiguity in any of these changes.

app/controllers/api/assignments/submissions_controller.rb
```diff
@@ -0,0 +1,61 @@
...
+        message: "No existing submission draft was found" }, status: 404
+    end
+  end
+
+  def create
+    assignment = Assignment.find(params[:assignment_id])
+    submission = assignment.submissions.new merged_submission_params(assignment)
+
+    if submission.save
+      render json: { submission: submission, message: "Successfully created a submission draft" }, status: 201
```
acousticrobot: See `/views/api/earned_badges/create.json.jbuilder` for an example create response.

app/models/submission.rb
```diff
@@ -148,6 +158,10 @@ def confirm_all_files
     end
   end
 
+  def draft?
+    link.blank? && text_comment.blank? && submission_files.empty? && submitted_at.nil?
+  end
```
acousticrobot: I haven't seen this from the UX perspective yet, but I'm wondering how this will work when a student is working on a text_comment_draft but has also entered a URL or file -- will the student think it's still all a draft, but the professor will think they submitted without a text comment?

app/proctors/submission_proctor.rb
```diff
@@ -0,0 +1,17 @@
+class SubmissionProctor
+  def initialize(submission)
+    @submission = submission
+  end
+
+  # Viewable if the submission is not an autosaved draft
+  # A submission is considered a draft if it has no link, text_comment,
+  # or submission_files
+  def viewable?
+    !@submission.draft?
```
acousticrobot: Should this take into account the current user as we do in the other models. If viewable is later called on a view shared with a student, shouldn't it always return true?

app/views/submissions/edit.html.haml
```diff
@@ -7,4 +7,7 @@
     - else
       %h4= "You are editing the assignment submission for Group #{presenter.submission.group.name }"
 
-  = render partial: "form", locals: { presenter: presenter }
+  - if current_user_is_student?
+    = render partial: "form", locals: { presenter: presenter }
```
acousticrobot: Maybe it's just the choice of file names, but I'm very confused why students would return the "form" but faculty the "autosave form", especially when you query `current_user_is_student?` in `autosave_form`.

app/views/submissions/new.html.haml
```diff
@@ -2,4 +2,7 @@
   = render partial: "layouts/alerts", locals: { model: presenter.submission }
 
   - if presenter.assignment.accepts_submissions?
-    = render partial: "form", locals: { presenter: presenter }
+    - if current_user_is_student?
+      = render partial: "form", locals: { presenter: presenter }
+    - else
+      = render partial: "autosave_form", locals: { presenter: presenter }
```
acousticrobot: Same comment about the if-else and `autosave_form`, these are the only two places where I see this partial rendered.

app/models/submission.rb
jonathoy: Link, text_comment, and submission_files aren't persisted until the student actually submits the submission. A "draft" submission ends up being just a container with autosaved text_comment_draft content.

I can see how this might be confusing for the user in the sense that not all elements of the form are saved. I'm open to being more explicit about autosaved changes being limited to only text. Thoughts @chcholman?

app/assets/javascripts/angular/directives/student/submission/hidden_form_method.js.coffee
```diff
@@ -0,0 +1,22 @@
+@gradecraft.directive 'hiddenFormMethod', ['StudentSubmissionService', (StudentSubmissionService) ->
```
jonathoy: Good point, this was definitely an odd one. I've gone ahead and added some comments to each directive.

jonathoy:
Thanks for the comments! I am in total agreement with the changes that you suggested. In response to the rake task, I believe that @chcholman has already written up one for this specific purpose.

app/controllers/api/assignments/submissions_controller.rb
```diff
@@ -0,0 +1,60 @@
...
+    if assignment.is_individual?
+      @submission = Submission.for_assignment_and_student(assignment.id, current_user.id).first
+    else
+      @submission = Submission.for_assignment_and_group(assignment.id, current_student.group_for_assignment(assignment).id).first
+    end
+
+    if @submission.present?
+      render "api/assignments/submissions/submission", status: 200
+    else
+      render json: { message: "Submission not found" }, status: 404
```
acousticrobot: I know I'm deviating from my last comment, but now I'm wondering if we should return a [JSON null resource](http://jsonapi.org/format#document-resource-object-linkage) if a submission doesn't exist?

The reason I'm altering my suggestion is that after testing in dev, I noticed we will always receive and ignore a 404 the first time the form is loaded, which doesn't feel right to me.

If the route was `api/submissions/:id` I think it would be necessary to stick to returning a 404, since it would be bound to a specific submission id. However, with the route of `/api/assignments/:assignment_id/submissions` this is technically acting more like a index route, so I think it makes sense to return a status OK with empty load.

This will change slightly the way you handle a success in your Service too. 

Alternatively, we could try to suppress the GET whenever the form URL is `assignments/:id/submissions/new`

Thoughts?

app/models/submission.rb
acousticrobot: We should do something about this before merge, but I think just adding an alert to the submit draft button for now that states "Warning, only text comments, no links or files, will be saved with a draft. Continue?" would be fine. Then we can decide if we want to handle this with a better UX, or add these fields to draft functionality in the next PR.

app/proctors/submission_proctor.rb
```diff
@@ -0,0 +1,17 @@
+class SubmissionProctor
+  def initialize(submission)
+    @submission = submission
+  end
+
+  # If the user is not considered staff, the submission should be viewable
+  # Otherwise, it should only be visible if it is not a draft
+  def viewable?(user)
+    return true if user.is_student?(@submission.course)
```
acousticrobot: I think we need to refine this so that only the submitting student or group member can view a draft, not any student in the course.

app/models/submission.rb
jonathoy: I discussed this with Cait since autosave also happens automagically on a 3.5s timer. We ultimately decided to put a little disclaimer underneath where the Last Saved caption is.

jonathoy:
@acousticrobot In response to the 404 vs 200 dilemma, I wasn't sure that I was a big fan of suppressing the Show/Index lookup based on the initial URL, so I changed the response code to the latter and simply returned an error message. For all intents and purposes, everything else works the same. Does that seem reasonable to you?

jonathoy:
@acousticrobot Through our discussion today, I made the change to return a 200 OK status with a response body containing a singular message and data set equal to null.

I also realized that one of the responses was not quite structured correctly according to the JsonApi standard in that the ID, type, and attributes were not being nested under data. I made this change and added an additional helper method to `serviceHelpers.js` to facilitate the copying of a formatted response to a single Angular model.

PR #2733, opened 2016-12-14
2704 refactor students page

### LINKED ISSUE #2704: Refactor Students page
### Expected Behavior
The students page is one of the central places in the app, and needs to load in < 1 second. 

### Actual Behavior
With 300 students in a class on staging, it takes > 1 minute to load the page. 

### Steps to reproduce
Navigate to staging.gradecraft.com/students

jwright:
### Status
**READY**

### Description
Fixes performance issues with the student index view.

The current performance of the student index view is around ~1M for 300 students. It runs ~11k sql statements.

screen shot 2016-12-14 at 5 07 11 pm

The new performance of the student view is around ~5s for 300 students. It runs 17 sql statements.

screen shot 2016-12-14 at 5 03 01 pm

### Migrations
NO

* Student leaderboard

Closes #2704

### DIFF
--- a/app/models/course_membership.rb
+++ b/app/models/course_membership.rb
@@ -56,19 +56,24 @@ class CourseMembership < ActiveRecord::Base
     update_attributes earned_grade_scheme_element_id: earned_grade_scheme_element.try(:id)
   end
 
-  def earned_grade_scheme_element
-    elements_earned = []
-    course.grade_scheme_elements.order_by_lowest_points.each do |gse|
+  def earned_grade_scheme_element(grade_scheme_elements=nil)
+    elements = grade_scheme_elements || course.grade_scheme_elements.order_by_lowest_points
+    element_earned = nil
+
+    elements.sort_by{ |element| element.lowest_points }.each_with_index do |gse, index|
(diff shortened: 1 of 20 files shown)


app/models/course_membership.rb
```diff
@@ -53,22 +53,27 @@ def recalculated_student_score
   end
 
   def check_and_update_student_earned_level
+    course.grade_scheme_elements.order_by_lowest_points.map { |gse| gse.check_unlock_status(user) }
```
acousticrobot: Unless I'm missing something, I don't think this line is doing anything.

jwright: `GradeSchemeElement#check_unlock_status` creates `unlock_state`s for the user.

acousticrobot: OK I see now this is just maintaining existing code side effects.  I wonder if this method should be changed to something like `update_unlock_status` or at least `check_unlock_status!` to signify it changes persistent data.

jwright:
@acousticrobot I changed the method to `unlock!`. I am just passing this back to you so you can approve the changes or provide feedback.

PR #2743, opened 2016-12-16
2742 submission constraint fixes

### LINKED ISSUE #2742: Constraint Prevents Multiple Group Submissions
### Expected Behavior

We should be able to create many group submissions for a given assignment.

### Actual Behavior

A recently added constraint prohibits the creation of more than one submission per assignment and student. Current functionality is that one submission is created per group. A group submission is expected to have `student_id = nil`, therefore this validation will fail.

### Steps to reproduce

On a group assignment, create and submit a submission. Then try to create and submit another submission for the same assignment.

jonathoy:
### Status
Ready

### Description
This PR addresses a few issues that were discovered with a newly added `Submission` constraint. This should also fix the failing specs on master.

The constraint limited submissions to being one per assignment and student. However, the default behavior for the ActiveRecord constraint was to validate against `nil` values. This prevented multiple group submissions from being created, since a group submission is defined as being a `Submission` with a `group_id` but no `student_id`.

Additional changes were also made to the database as well as the `Submission` model to ensure that no more than one group submission can exist per assignment.

Finally, there is a newly added ActiveRecord constraint that dictates that a `Submission` is valid if it has either a non-nil value for `student_id` or `group_id`, but not both.

**If there is the possibility that there is preexisting production data that would not fit this constraint, a rake task should be included to clean this up**

Fixes #2742

### Deploy Notes
Since we define a group submission as being a `Submission` with a `group_id` but no `student_id`, and an individual `Submission` as being one with `student_id` but no `group_id`, we should make sure that there is no preexisting data that would violate this.

### Migrations
YES

- Verify that for any given student, one and only one `Submission` can be created on an individual assignment
- Verify that for any given group, one and only one `Submission` can be created on a group assignment
- Ensure that a `Submission` that is considered invalid cannot be created, e.g. it must have either a `group_id` or `student_id`, but not both

### DIFF
--- a/app/models/submission.rb
+++ b/app/models/submission.rb
@@ -52,8 +52,8 @@ class Submission < ActiveRecord::Base
 
   validates :link, format: URI::regexp(%w(http https)), allow_blank: true
   validates_length_of :link, maximum: 255
-  validates :assignment, presence: true, uniqueness: { scope: :student,
-    message: "should only have one submission per student" }
+  validates :assignment, presence: true, uniqueness: { scope: [:student, :group],
+    message: "should only have one submission per student or group" }, allow_nil: true
   validates_with SubmissionValidator
 
   clean_html :text_comment
--- a/app/models/submission.rb
+++ b/app/models/submission.rb
@@ -179,8 +179,6 @@ class Submission < ActiveRecord::Base
   end
 
   def student_xor_group
-    unless student.nil? ^ group.nil?
-      errors.add(:base, "must have either a student_id or group_id, but not both")
-    end
+    errors.add(:base, "must have either a student_id or group_id, but not both") unless student.nil? ^ group.nil?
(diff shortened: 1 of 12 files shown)


jonathoy:
@chcholman Let me know if there is need for a rake task to ensure there is no bad `Submission` data.

app/models/submission.rb
```diff
@@ -176,4 +177,8 @@ def cache_associations
     self.assignment_id ||= assignment.id
     self.course_id ||= assignment.course_id
   end
+
+  def student_xor_group
+    errors.add(:base, "must have either a student_id or group_id, but not both") unless student.nil? ^ group.nil?
```
acousticrobot: I think this is a great example of a well named definition that self-documents a somewhat obtuse logic gate.

PR #2747, opened 2016-12-19
2688 group comments autosave

### LINKED ISSUE #2688: bug - group rubric grading criterion comments autosave
### Expected Behavior

I expect autosave on CriterionGrade comments to work for group graded assignments.

### Actual Behavior

`RubricService#updateCriterion` does not take into account groups, so it saves comments on the student id, rather than the group id.

### TO FIX:

After #2682 is merged (which disables the api call and adds the scope type), we need to handle when `recipientType == "group"`.

acousticrobot:
### Status
**READY**

### Description

The comment field on criterion grades has an autosave feature that does not currently work for group grading. This PR adds autosave functionality for the `CriterionGrade.comment` fields for group graded assignments.

Additionally, it adds autosave for the grade summary feedback for each of the group member's grades.

(A minor fix in the standard edit form fixes autosave for individual grades w/o rubrics, the group equivalent has it's own issue: #2748).

It also updates the $http calls from the Angular `GradeService` and `RubricService` to use the `.then` syntax.

closes #2688.

### Migrations
NO

As a professor, grade a group assignment with a rubric.  Without hitting submit, the text comment fields on criterion should still autosave.

### DIFF
--- a/app/controllers/api/criterion_grades_controller.rb
+++ b/app/controllers/api/criterion_grades_controller.rb
@@ -46,6 +46,22 @@ class API::CriterionGradesController < ApplicationController
     end
   end
 
+  # PUT api/assignments/:assignment_id/groups/:group_id/criterion_grades
+  def group_update
+    result = Services::CreatesGroupGradesUsingRubric.create params, current_user.id
+    if result.success?
+      render json: {
+        message: "Grade successfully saved", success: true
+        },
+        status: 200
+    else
+      render json: {
+        errors: [{ detail: result.message }], success: false
+        },
+        status:  result.error_code || 400
+    end
+  end
+
   # PUT api/assignments/:assignment_id/students/:student_id/criteria/:id/update_fields
   def update_fields
     cg = CriterionGrade.find_or_create(params[:assignment_id], params[:id], params[:student_id])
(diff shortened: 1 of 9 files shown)


app/controllers/api/criterion_grades_controller.rb
```diff
@@ -46,34 +46,55 @@ def update
...
+    if result.success?
+      render json: {
+        message: "Grade successfully saved", success: true
+        },
+        status: 200
+    else
+      render json: {
+        errors: [{ detail: result.message }], success: false
+        },
+        status:  result.error_code || 400
```
jonathoy: I'm curious to know your opinion on when to return a 400 versus a 500. My personal preference is that I would return a 500 here if there was no `error_code`, but I think that both cases could be argued for. I find that I usually return a 400 if I am doing some sort of explicit check with the request and there is something unexpected.

acousticrobot: My hunch is that if the error message is coming from AR it most likely would be a database constraint triggered by the request, but you are right -- since we aren't checking for what caused the error, this could very well be an error on the server itself. We should look into a better common strategy to employ in all our controllers for handling errors, including when saving multiple models.

Issue #2487, opened 2016-09-27
Remove Grade.find_or_create


jwright:
Remove the use of `Grade.find_or_create` so that empty grades are not created.

This allows a grade to mean that an assignment has been graded.

jwright:
Blocked due to changes waiting in #1492

chcholman:
@jwright No longer blocked!

acousticrobot:
I just went though the app to try and identify the places where a Grade or associated models are updated. Many of these will still work without an existing Grade but we should examine how each on would work in concert with a more RESTful routing that includes a `create` path. It would be great if we could reduce the number of controller actions needed for creating and updating grades.

## Places where a grade is updated:

| route                                         | purpose                                          | used from                  |
| --------------------------------------------- | ------------------------------------------------ | -------------------------- |
| API::CriterionGradesController#update         |  saves CriterionGrades and Grade for student     |  Rubric Grading Page       |
| API::CriterionGradesController#group_update   |  saves CriterionGrades and Grade for group       |  Rubric Grading Page       |
| Assignments::GradesController#update_status   |  changing the grade status for a group of grades |  assignments/grades/edit_status.html.haml |
| Assignments::GradesController#mass_update     |  saves all the grades for an assignment          |  mass edit Grading Page    |
| Assignments::GradesController#self_log        |  students saving a self-logged grade             |  student's self-logged assignment show page |
| Assignments::GroupsController#graded          |  saves grades for Assignment and Group           |  Standard Group Grading Page   |
| Assignments::StudentsController#grade         |  saves grade for a student                       |  6 calls to assignment_student_grade_path   | 
| GradesController#update                       |  saves grade for a student                       |  Standard Grading Page   | 
| GradesController#exclude                      |  updates grade to be excluded from course score  |  Faculty's Assignment Index page   | 
| GradesController#include                      |  updates grade to be included from course score  |  Faculty's Assignment Index page   | 

## Places where a grade's associated model is updated:

| route                                         | purpose                                          | used from                  |
| --------------------------------------------- | ------------------------------------------------ | -------------------------- |
| API::CriterionGradesController#update_fields   |  autosaves feedback on CriterionGrade           |  Rubric Grading Page      |
| API::Grades::GradeFilesController#create       |  saves a grade file with a grade_id           |  Rubric Grading Page       |
| API::Grades::GradeFilesController#destroy      |  destroys a grade file                          |  Rubric Grading Page      |
[...]

PR #2754, opened 2016-12-20
2488 remove empty grades

### LINKED ISSUE #2488: Remove empty grades
Create rake tasks to perform the following:
1. Delete any grade that has a `predicted_score` > 0
2. Delete any grade with a nil `raw_points` and a nil `feedback`
3. Populate the remaining grades with a nil `graded_by_id` with Cait's user id (1) (Do not update the `updated_at` field)
4. Populate the remaining grades with a nil `graded_at` with the grade's `updated_at` field (Do not update the `updated_at` field)
5. Create a migration to remove the `Grade#predicted_score` column

jwright:
### Status
**HOLD**

### Description
This PR cleans up several grade related items to remove ungraded grades and grades with predicted scores.

**This should be merged in after #2487 or having a `Grade.find_or_create` will create additional grades that these migrations are trying to clean up**

### Related PRs

branch | PR
------ | ------
#2487 | [link](https://github.com/UM-USElab/gradecraft-development/issues/2487)

### Deploy Notes

**Do not run the migration before these rake tasks are run.**

After this is merged into production, run the following tasks in this order:

1. `bundle exec rake grades:remove_empty_grades:delete_predicted_grades`
1. `bundle exec rake grades:remove_empty_grades:delete_empty_grades`
1. `bundle exec rake grades:remove_empty_grades:update_ungraded_grades`
1. `bundle exec rake grades:remove_empty_grades:update_missing_timestamps`

After these are run a new task can be created to remove all of the jobs under `grades:remove_empty_grades`

### Migrations
YES (Do not run before rake tasks are run)

* Grading

Closes #2488

### DIFF
--- a/lib/tasks/grades.rake
+++ b/lib/tasks/grades.rake
@@ -35,15 +35,23 @@ namespace :grades do
     desc "Deletes any grade that has a predicted score greater than 0"
     task delete_predicted_grades: :environment do
       grades = Grade.where("predicted_score > 0")
-      puts "Deleting #{pluralize grades.count, "predicted grade"}..."
+      puts "Deleting #{ActionController::Base.helpers.pluralize grades.count, "predicted grade"}..."
       grades.destroy_all
     end
 
     desc "Deletes any grade without raw points or feedback"
     task delete_empty_grades: :environment do
       grades = Grade.where(raw_points: nil, feedback: nil)
-      puts "Deleting #{pluralize grades.count, "empty grade"}..."
+      puts "Deleting #{ActionController::Base.helpers.pluralize grades.count, "empty grade"}..."
       grades.destroy_all
+    end
+
+    desc "Update grades without a graded by id to be graded by user id 1"
+    task update_ungraded_grades: :environment do
+      user = User.find 1
+      grades = Grade.where(graded_by_id: nil)
+      puts "Updating #{ActionController::Base.helpers.pluralize grades.count, "ungraded grade"} with user #{user.name}..."
+      grades.update_all(graded_by_id: user.id)
     end
   end
 end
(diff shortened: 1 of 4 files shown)


lib/tasks/grades.rake
```diff
@@ -30,4 +30,35 @@ namespace :grades do
       prediction.save
     end
   end
+
+  namespace :remove_empty_grades do
+    desc "Deletes any grade that has a predicted score greater than 0"
+    task delete_predicted_grades: :environment do
+      grades = Grade.where("predicted_score > 0")
```
acousticrobot: I feel like I'm not reading this task correctly. 

 1) Won't the migration run, removing the `predicted_score` column, before we can run this rake task?
 2) Since legacy `Grade` rows were used to store both the grade and prediction information, I think there are legit grades out there that also have predictions on them.

I think we can remove this task entirely, and start by running task 2: `delete_empty_grades` 

Let me know if I'm missing something.

jwright: 1. Yes. We want to run the rake task before the migrations are run. I will add that to the deployment notes.

2. If that is the case, what we discussed a few months ago is no longer valid. These steps were created back then when we talked about this task and probably needs to have another discussion.

Legacy Grades seem to not be agreed upon and this whole task really doesn't matter if we still create grades early by not removing `Grade.find_or_create` anyway. I am going to block this PR as well until we chat about legacy grades.

chcholman:
@acousticrobot @jwright Can you both confirm that this PR is not blocked or on Hold as it states? Just want to make sure before I merge!

chcholman:
Got it! Thank you both, I'll add this to my list of things that I need to review and have thoughts regarding how to proceed for when we meet in the New Year @acousticrobot @jwright

PR #2755, opened 2016-12-20
2728 accessibility review instructor side

mariehooper:
### Status
**IN DEVELOPMENT**

### Description
This PR is an accessibility review of the instructor side of GradeCraft.

### Todos
- [x] final pass through buttons vs links
- [x] make sure ngAria isn't causing errors
- [x] finish adding roles and `aria-describedby` to tooltips

* Instructor pages

Closes #2755


chcholman:
@mariehooper This is looking fantastic. I'm merging it, but going to leave the PR open and would suggest that @jonathoy @acousticrobot @jwright take a look through the work to start getting familiar with the elements we *should* be adding to support accessibility constantly.

Issue #2736, opened 2016-12-15
Student preview events firing analytics updates as though student


chcholman:
### Expected Behavior
As an instructor, if I preview as a student, that should not update last_activity values or fire any other analytics. 

### Actual Behavior
Currently, an instructor previewing as a student updates their last_activity metric. We should also investigate to confirm whether or not this affects other analytic events more broadly.

acousticrobot:
This definitely will occur for all events when previewing as a student and needs to be fixed. Unfortunately AFAIK there is no central place to limit analytics events, so we will have to check for impersonating agent every time a background job is enqueued, and pass a message to disable any updates to analytics.

PR #2756, opened 2016-12-20
2736 student preview events

jwright:
### Status
**READY**

### Description
This PR overrides the [Sorcery](https://github.com/Sorcery/sorcery) activity logging methods so that they take into account impersonation.

The Sorcery controller methods for registering the last logout timestamp, the last activity timestamp, and the last IP address were updated so the impersonating agent is used instead of the current user if the user is impersonating a student.

In addition, the impersonation helper methods were moved into a dedicated helper for better segregation and the method names were updated to be more descriptive.

This does not complete the work needed for #2736 but it is a good stopping point to get the activity functionality complete.

### Migrations
NO

* Logging in
* Impersonation

### DIFF
--- a/app/controllers/api/assignments_controller.rb
+++ b/app/controllers/api/assignments_controller.rb
@@ -7,10 +7,10 @@ class API::AssignmentsController < ApplicationController
 
     if  current_user_is_student?
       @student = current_student
-      @allow_updates = !student_impersonation?
+      @allow_updates = !impersonating?
       @grades = Grade.for_course(current_course).for_student(current_student)
 
-      if !student_impersonation?
+      if !impersonating?
         @assignments.includes(:predicted_earned_grades)
         @predicted_earned_grades =
           PredictedEarnedGrade.for_course(current_course).for_student(current_student)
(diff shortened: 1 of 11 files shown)


app/controllers/application_controller.rb
```diff
@@ -110,6 +111,26 @@ def save_referer
     session[:return_to] = request.referer
   end
 
+  # Sorcery activity logging overrides
```
acousticrobot: Do we need to override [`register_login_time_to_db`](https://github.com/NoamB/sorcery/blob/master/lib/sorcery/controller/submodules/activity_logging.rb#L45) as well?

app/controllers/api/assignments_controller.rb
```diff
@@ -7,10 +7,10 @@ def index
 
     if  current_user_is_student?
       @student = current_student
-      @allow_updates = !student_impersonation?
+      @allow_updates = !impersonating?
```
acousticrobot: I approve of this name change -- especially considering how many times we need to call this method!

app/controllers/application_controller.rb
jwright: We don't because there is no way to login being impersonated. There is no session state for the impersonating agent until there user is already logged in.

jwright:
@acousticrobot Answered your question. Let me know if you don't think that's right. If it is, I think it's ready for merging.

PR #2766, opened 2016-12-22
2736 job refactor

jwright:
### Status
**READY**

### Description
The PR attempts to replace the `LoginEventLogger` class and the `LoginEventPerformer` with Rails' [ActiveJob](http://edgeguides.rubyonrails.org/active_job_basics.html) functionality to replace the `LoginEventLogger` and [Porch](https://github.com/jwright/porch) to replace the `LoginEventPerformer` functionality in an attempt to make it clearer what is happening with the background jobs.

The performers and event loggers are complicated to understand because there are several methods that are overridden from base classes but they are not immediately understood how they are used. The `messages` method is one such method. They are defined in a subclass but used in a base class. The `InheritableIvars` module is also very confusing as the opposite of the inheritance chain is true.

The inheritance chain for loggers is confusing as there are arguments passed into the constructor but no initialize method is defined. I still am unsure how that is working.

In addition, this PR also clears up the complicated tests around the logger and the performer with a single inclusive test for the job stuff which uses just the standard Rails RSpec job matchers. The performer stuff was replaced with specs that follow the current LightService standard where the  organizers are tested to ensure that the correct steps are ran and individual tests for each step to ensure they perform the correct things.

The following classes would be deleted from the application if this PR is taken further:

1. `app/event_loggers/login_event_logger`
1. `app/performers/login_event_performer`
1. `spec/event_loggers/login_event_logger_spec`
1. `spec/performers/login_event_performer_spec`

[Porch](https://github.com/jwright/porch) was created to appease some of the issues we have with LightService. For one, it allows for blocks and private methods to run in the context chain. In addition, the actions are instances instead of class methods. There is other functionality we can add such as chaining services together and iterating over services. I believe we can use this to replace our current LightService services as well.

This framework was used in place of the "Performer" framework that was created which has the same functionality as LightService or Porch but it is not yet another additional framework. Once we replace the existing services with Porch (trivial update), we will have one service framework.

Since the logger is being replaced with `ActiveJob`, we can easily replace the `ResqueJob` implementation with `ActiveJob` which will then allow us to use Sidekiq as well.

I'd love to get feedback on what I have so far and what everyone thinks about the future direction with using the service architecture for the performers and ActiveJob for the event loggers. The next step would be to replace the other event loggers and performers.

### Todos
- [X] Create an `ActiveJob` sub-system that is specific to running logging events
- [X] Use `RecordLoginEvent` in the `ApplicationController`
- [x] Add Resque plugins (`Retry` and `ExponentialBackoff`) to the default resque instance
- [x]  Update the log messages to use the new classes
- [x] Log the error if one occurs within the service (using `rescue_from`?)
- [x] Remove event logger and performer for login events (moved to #3028)

### Migrations
NO

* Logging the last login event


jwright:
@chcholman @acousticrobot @jonathoy Let me know if you guys have any feedback on this PR. I am going to continue with the work here shortly.

chcholman:
@jwright This is looking awesome Jamie! Much easier to read.

jwright:
In addition, one of the most undesirable side effects of the LS service pattern is a "different class for each step". So, since these middleware pieces are instances of classes, you could use blocks instead:

```
module EventRecorders
  class LoginEventRecorder
    configure_middleware do |chain|
      chain.add :log_start do |context|
        context.logger.info "The recording has started..."
      end

      chain.add FindCourseMembership
      chain.add RecordAnalyticsEvent
      chain.add UpdateLastLogin
      chain.add LogEnd
    end
  end
end
```

or use methods:

```
module EventRecorders
  class LoginEventRecorder
    configure_middleware do |chain|
      chain.add LogStart
      chain.add :find_course_membership
      chain.add RecordAnalyticsEvent
      chain.add UpdateLastLogin
      chain.add LogEnd
    end

    private

    def find_course_membership(context)
      context.course_membership = CourseMembership.where(user_id: context.user_id, 
        course_id: context.course_id).first
    end
  end
end
```

jwright:
@chcholman @acousticrobot @jonathoy I updated this PR which now replaces the performers with [Porch](https://github.com/jwright/porch) and the event loggers/resque_job with ActiveJob.

If you like this approach, let me know. We can probably merge this PR (after a little bit of clean up for the TODOs) in and then we can work on replacing the others once we find this works well.

jwright:
@acousticrobot Thanks for the feedback.

I like the idea of making `guard`, `guard!`, `next`, and `skip_remaining` more clear. I will think through some better verbs and see what you think.

In terms of adding the `on_guard` clause in the organizer, I am a bit skeptical on that because I don't like the idea of someone seeing `guard` in the action (may be a different file) and then having to go back to the organizer and see what the "setting" was. I like the idea of making it more reusable however so it's hard to strike a balance there.

jwright:
@acousticrobot I updated the `Context` methods to be more descriptive

* `Context#next` -> `Context#skip_next`
* `Context#guard` -> `Context#guard_with_skipping`
* `Context#guard!` -> `Context#guard_with_failure`

jwright:
@acousticrobot This is ready for review finally :)

jwright:
@acousticrobot Thanks man.

We will be removing the files pretty quickly. I just want to ensure we at least get some successful transactions.

In addition, I will be starting on removing all the loggers soon afterwards.

chcholman:
@jwright @acousticrobot Okay! Merging now so that we can get some local dev test time in, and then aiming to get it into production early this weekend. I'll be 👀

PR #2777, opened 2016-12-28
Test suite factories

chcholman:
### Status
**IN DEVELOPMENT**

### Description
This PR attempts to both simplify and speed up the test suite by removing World, leveraging factories, and creating single objects that can be referenced consistently per file. This PR will only cover Models, more work to come.

### DIFF
--- a/spec/models/assignment_spec.rb
+++ b/spec/models/assignment_spec.rb
@@ -1,9 +1,9 @@
 describe Assignment do
-  include Toolkits::Models::AssignmentsToolkit
-
   subject { build(:assignment) }
 
-  it_behaves_like "a model that needs sanitization", :description
+  context "with a persisted assignment" do 
+    it_behaves_like "a model that needs sanitization", :description
+  end
 
   context "validations" do
     it "is valid with a name and assignment type" do
@@ -227,7 +227,6 @@ describe Assignment do
   end
 
   describe "position" do
-
     it "sets the position by assignment type on save (using acts_as_list gem)" do
       expect(subject.position).to be_nil
       subject.save
--- a/spec/models/badge_spec.rb
+++ b/spec/models/badge_spec.rb
@@ -1,5 +1,7 @@
 describe Badge do
-  subject { build(:badge) }
+  subject { create(:badge) }
+  let(:student) { create(:user) }
+  let(:assignment) { create(:assignment) }
 
   context "validations" do
     it "is valid with a name and a course" do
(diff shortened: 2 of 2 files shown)


spec/models/assignment_spec.rb
```diff
@@ -1,13 +1,15 @@
 require "active_record_spec_helper"
 require "toolkits/sanitization_toolkit"
-require "toolkits/models/assignments_toolkit"
```
jwright: 👍

spec/models/assignment_spec.rb
```diff
@@ -1,13 +1,15 @@
...
 
 describe Assignment do
-  include Toolkits::Models::AssignmentsToolkit
 
-  subject { build(:assignment) }
+  subject { build_stubbed(:assignment) }
+  let(:student) { build_stubbed(:user) }
 
-  it_behaves_like "a model that needs sanitization", :description
+  describe "sanitization of the description" do
```
jwright: In this situation, I would use `context` since it has a different sut (subject under test) and context feels better here.

```
context "with a persisted assignment" do
  subject { create :assignment }
  it_behaves_like "a model that needs sanitization", :description
end
```

spec/models/assignment_spec.rb
```diff
@@ -17,178 +19,149 @@
     it "is invalid without a name" do
       subject.name = nil
       expect(subject).to_not be_valid
-      expect(subject.errors[:name]).to include "can't be blank"
```
jwright: I think taking this out reduces the reliability of the test. The when clause (`subject.name = nil`) may or may not of caused the expectation (`expect(subject).to_not be_valid`).  The subject may be invalid for several reasons and by not checking the specific error, it may get a false positive. In addition, the expectation of `expect(subject.errors[:name]).to include "can't be blank"` also ensures that the error message is correct.

spec/models/assignment_spec.rb
```diff
@@ -200,25 +173,29 @@
...
+      subject.max_group_size = 2
+      subject.min_group_size = 5
+      expect(subject).to_not be_valid
+    end
   end
 
   describe "position" do
-
     it "sets the position by assignment type on save (using acts_as_list gem)" do
+      subject = build(:assignment)
```
jwright: I would change this to have one expectation by using the `to change` matcher. (https://www.relishapp.com/rspec/rspec-expectations/v/2-2/docs/matchers/expect-change)

```
expect { subject.save }.to change(:position).from(nil).to(1)
```

spec/models/assignment_spec.rb
```diff
@@ -250,16 +227,9 @@
...
-    it "errors out if the max group size is smaller than the minimum" do
-      subject.max_group_size = 2
-      subject.min_group_size = 5
-      expect !subject.valid?
-    end
-  end
-
+  
   describe "#graded_or_released_scores" do
+    subject { build :assignment }
```
jwright: We could organize these specs within contexts for easier understanding of the setup:

```
RSpec.describe Assignment do
  context "with a stubbed Assignment" do
    subject { build_stubbed(:assignment) }
    # ...
  end

  context "with a built Assignment" do
    subject { build(:assignment) }
    # ...
  end

  context "with a saved Assignment" do
    subject { create(:assignment) }
    # ...
  end
end
```

spec/models/assignment_spec.rb
```diff
@@ -393,7 +365,7 @@
   end
 
   describe "#unlock!" do
-    let(:student) { create :user }
+    subject { build :assignment }
```
jwright: We could reduce these 2 lines to `subject { create :assignment }`.

spec/models/assignment_spec.rb
chcholman: Cool! I was testing to see how significantly these affected the length of the test suite - all of them totaled to ≤10 seconds. Seems like it's worth it leave them in. They're missing in a couple of places so I'm aiming to add them in/make it consistent.

PR #2834, opened 2017-01-06
Password reset page

### LINKED ISSUE #2813: Fix broken reset password page inside app
### Expected Behavior
As a student, I expect to be able to fill out a form to reset my password.

### Actual Behavior
Password reset form page styling is broken and form elements are partially hidden behind header.

### Steps to reproduce
[Trello Card](https://trello.com/c/coafl3Go/72-fix-the-unfinished-broken-style-for-resetting-your-password)
Branch: `password-reset-page`
URL: http://localhost:5000/passwords/reset

mariehooper:
### Status
**READY**

### Description
This PR restyles the password rest page for users who request a new password as well as removes a legacy view, route and controller method for requesting a new password inside app.

### Migrations
NO

* Password reset page (tokenized link from email)

Closes #2813

### DIFF
--- a/app/models/breadcrumb_trail.rb
+++ b/app/models/breadcrumb_trail.rb
@@ -354,6 +354,10 @@ class BreadcrumbTrail < Croutons::BreadcrumbTrail
   def grades_importers_import_results
     assignments_index
   end
+  
+  def passwords_new
+    breadcrumb('Dashboard', dashboard_path)
+  end
 
   def rubrics_index_for_copy
     dashboard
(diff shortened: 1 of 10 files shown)


app/models/breadcrumb_trail.rb
```diff
@@ -408,6 +408,10 @@ def grades_importers_import_results
     breadcrumb('Dashboard', dashboard_path)
     breadcrumb('#{ term_for :assignments }', assignments_path)
   end
+  
+  def passwords_new
+    breadcrumb('Dashboard', dashboard_path)
```
chcholman: Seems like we should be able to remove this method if we've deleted the rest of the `new` method?

mariehooper: yep! will do.

chcholman:
@mariehooper Finally found the last page - account activation also has a password set form. Pulled in your styles to the form, but we can refactor to have both pages reference the same view partial next!

mariehooper:
@chcholman - got it! Thank you, Cait!

Issue #1633, opened 2016-02-07
LTI Certification


chcholman:
We have built a functional implementation of LTI 1.1, but we've never completed the certification tests. The test suite is here:
https://www.imsglobal.org/lti-certification-suite

The piece that we're missing are various error responses to handle every condition the spec requires. This task is to go through the test suite, add the necessary errors in response, and to suggest/do any cleanup of the current implementation if necessary.

jonathoy:
Some Test Suite Results:

Test 4.5: Launch as user with no context role
- Status: Failure
- Expected result: User should be given whatever privileges are appropriate to a user with no role, which could be that they are denied access
- Observed result: User is assumed to be a student
- Action: Present user with a message stating that they should be granted a role in the course

Test 4.6: Launch as a user with an institution role but no context role
- Status: Needs Further Investigation
- Expected result: User should be given privileges appropriate to a user with no context role, which could be that they are denied access
- Observed result: 500 Internal Server Error
- Action: Investigate and possibly implement institution roles: https://www.imsglobal.org/specs/ltiv1p0/implementation-guide#toc-19

Test 4.7: Launch as a user with an institution role which has no corresponding context role 
- Status: Needs Further Investigation
- Expected result: User should be given privileges appropriate to a user with no context role, which could be that they are denied access
- Observed result: 500 Internal Server Error
- Action: See Test 4.6

jonathoy:
Per our discussion, we will not be pursuing LTI 1.1 certification since our grading structure does not align with the way that LTI expects grades. Since it appears that LTI 1.2 certification requires the implementation of simple outcomes as in LTI 1.1, we will likely be aiming for LTI 1.0 certification.

chcholman:
@jonathoy Emailed for a consultation on this 1/30, will let you know as soon as I've got something scheduled.

PR #2852, opened 2017-01-10
2784 observer role

### LINKED ISSUE #2784: Observer role
We need the ability to assign users to "Observer" roles. In this mode, they should be able to see the assignments, but no grade analytics. They should not be able to submit anything. They should be able to use the grade predictor. They should not appear in any student list, but rather should appear on the staff page under an "Observer" filter.

jonathoy:
### Status
READY

### Description
This PR addresses the need for an observer role. This role now becomes the default role when creating a new course membership.

As an observer, the user should be able to see assignments but not any grade analytics. They should also not be allowed to submit anything. However, the grade predictor is accessible. Said observers will be listed in a separate view via the staff navbar, accompanying listings for Learners, Teams, Groups, and Staff.

Closes #2784 

**Notable changes in this PR:**

Since the addition of a new default role affected numerous spec tests, some concurrent refactoring was done to some of the FactoryGirl factories. Notice that in the excerpt below, `course_membership_factory` now uses traits:

```FactoryGirl.define do
  factory :course_membership do
    association :course
    association :user

    instructor_of_record false
    role :observer

    trait :student do
      instructor_of_record false
      role :student
    end

    trait :staff do
      instructor_of_record true
      role :gsi
    end

   ...
```

This allows us to do the following: 
`create(:course_membership, :student, course: course, user: student)`

Furthermore, by introducing transient attributes to the `user_factory` there now is a simpler method for creating/building/stubbing a new user with a role. We no longer have to write an extra line to explicitly create the course membership.

These lines: 
```
  let(:professor) { create(:user) }
  let!(:professor_course_membership) { create(:course_membership, :professor, course: course, user: professor) }
```
can now be condensed to:
```
  let!(:professor) { create(:user, courses: [course], role: :professor) }
```
### Related Issues
This addresses a concern related to #1633 LTI Certification.

### Todos
- [ ] Documentation

### Deploy Notes
There is a migration to change the default role when creating a new course membership

### Migrations
YES

- [ ] Ensure that no tests were broken or inadvertently altered during the refactoring
- [ ] Create a new user with the observer role. As an observer, this user should have only limited access in GradeCraft
- [ ] An observer for a course should appear in the Observers page, clickable from the staff navigation pane
- [ ] Student sample data should work as expected. See 20ce13a45a25ff16849e9ddb0f375fb2604d207e, which addresses the change of default role from "student" to "observer".
- [ ] Canvas integration should continue to work as expected (92e8722)


app/controllers/application_controller.rb
```diff
@@ -107,6 +107,11 @@ def ensure_admin?
     return not_authenticated unless current_user_is_admin?
   end
 
+  def check_observer_role
```
acousticrobot: Can we change this to `ensure_not_observer?` to align with all of the other core role authentication? (`ensure_student?`, `ensure_not_impersonating?` etc.)

app/controllers/user_sessions_controller.rb
```diff
@@ -36,14 +36,16 @@ def lti_create
...
-    @user.courses << @course unless @user.courses.include?(@course)
-    @course_membership = @user.course_memberships.where(course_id: @course).first
-    @course_membership.assign_role_from_lti(auth_hash) if @course_membership
-    save_lti_context
-    session[:course_id] = @course.id
-    auto_login @user
-    record_course_login_event user: @user
-    respond_with @user, location: dashboard_path
+    # TODO: should we rollback user, course creation if the course membership cannot be created?
+    if CourseMembership.create_course_membership_from_lti(@user, @course, auth_hash)
```
acousticrobot: This reads much better!

jonathoy:
@acousticrobot I've made some changes to address the comments you made.

I also was looking back at UserSessionsController and noticed that there was a slight difference in how the LTI hook worked after my refactor. Mainly, there was no longer anything to stop a duplicate course membership from being created. If you could do a quick sanity check for that change, that would be great.

chcholman:
@jonathoy I made one tiny tweak (allowing observers to see the badge show page), and have two additional requests.

[] Prevent observers from seeing the submission button 
[] Prevent observers from seeing the Create a Group button

They don't have access to either of these functions, but keeping this as simple and clear as possible will be ideal!

@mariehooper  will you take a look at this branch and see whether or not you think the observer role should see the empty bar charts on the assignments page (since without grades, they'll never fill in) or if we should hide those as well?

mariehooper:
Hey @jonathoy - I agree with @chcholman re: the bar graph representation of points earned on the accordion on the Assignments page. Since the observer role wouldn't earn points, it's probably better to just show the total points for the category rather than the graph + earned/total points. I will take care of this in another issue since I would like to update the instructor view of assignments also and the styling will be almost identical so don't worry about doing anything in this branch with that! :)

jonathoy:
@mariehooper Great, thanks for the feedback!

@chcholman As we discussed, there are a couple issues that I found that need to be addressed if we want to handle observers correctly as a separate role. Since we are in agreement that an observer should not fit the context of being the `current_student`, in revising the scope I have discovered the following issues:

1. The badges show page renders, but does not work as expected. We render the student partial, but the presenter logic returns all earned badges for that badge. So if 40 students total have earned Badge A, there will be 40 awarded details (when it was earned, description, etc.)

2. Grade Predictor seems to work well, as far as I can tell.

3. Grade Scheme Elements does not render, since there is a lot of `current_student`-oriented logic in the view relating to points.

4. Assignments does not render, for the same reason as above. I believe this one would require a solid amount of work since it is very Grades-oriented.

In most cases, we see this initial conditional block:

```
- if current_user_is_staff?
  = render partial: "show_staff", locals: { presenter: presenter }
- else
  = render partial: "show_student", locals: { presenter: presenter }
```

which renders the student view since an observer is not considered staff. I believe the best solution would be to have an `elsif current_user_is_student?` block and then have observers fall into an `else`, but that requires us to add all the compatible views. Would you like me to move forward with creating these views?

chcholman:
@jonathoy Yes, please do continue on with creating the new views. Will you start with 1 (Badges) and 3 (Grade Scheme) since they're the clearest? I'd love to review it at that stage, and then strategize how to do Assignments.

jonathoy:
@chcholman I'm passing this along to you for your analysis on the work required for the Assignments show/index pages.

jonathoy:
Looks good to me! I had a minor suggestion (10ffbeb) and that was to just pass in nil for the student, since the current_student should always be nil as an observer. Would you agree with this? Either way, I think that this PR is about ready.

chcholman:
@jonathoy Great catch, thank you. Finally, finally merging!

PR #2857, opened 2017-01-11
add delayed event angular helper

acousticrobot:
### Status
**READY**

### Description

This PR extracts the queue logic from the student submission service to be used by other services.

I am submitting this PR in advance of a need I have for a single service to handle multiple updates (one per each criterion grade in a rubric).  This PR takes the pattern from the student submission debounce logic and expands it to handle a object store of queued events. It also removes the logic to a helper to reduce boilerplate within the services.

Each event is given a type and an id, so that timeout events are stored per object type and unique id, avoiding cross-cancelling of events.

Ideally each id would be the objects id, but this may not be possible when creating new objects.  In the refactor for instance, the submission is stored by assignment id, since this id is passed in. I'm open to other ways of solving this.

### DIFF
--- a/app/assets/javascripts/angular/helpers/delayedEvent.js.coffee
+++ b/app/assets/javascripts/angular/helpers/delayedEvent.js.coffee
@@ -1,0 +1,34 @@
+# Debounce manager used to auto update items
+#
+# Each item is given a unique timeout stored by type and id.
+# It can be cancelled and updated before fire,
+# or fired immediately, cancelling the standing delayed event.
+
+angular.module('helpers').factory('DelayedEvent', ['$timeout', ($timeout)->
+
+  queueStore = {}
+  TIME_TO_DELAY = 3500
+
+  addQueue = (type, id, event, args=[], immediate=false)->
+    queueStore[type] = queueStore[type] || {}
+
+    if immediate is true
+      cancelQueue(type, id)
+      event.apply(null, args)
+    else
+      cancelQueue(type, id)
+      queueStore[type][id.to_s] = $timeout(() ->
+        event.apply(null, args)
+      , TIME_TO_DELAY)
+
+
+  cancelQueue = (type, id)->
+    return false if !queueStore[type]
+    $timeout.cancel(queueStore[type][id.to_s])
+
+  return {
+    addQueue: addQueue
+    cancelQueue: cancelQueue
+  }
+])
+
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/angular/helpers/delayedEvent.js.coffee
```diff
@@ -0,0 +1,34 @@
...
+# Each item is given a unique timeout stored by type and id.
+# It can be cancelled and updated before fire,
+# or fired immediately, cancelling the standing delayed event.
+
+angular.module('helpers').factory('DelayedEvent', ['$timeout', ($timeout)->
+
+  queueStore = {}
+  TIME_TO_DELAY = 3500
+
+  addQueue = (type, id, event, args=[], immediate=false)->
```
jonathoy: Would it make more sense to name this addEvent since it adds a singular event to the queue?

app/assets/javascripts/angular/helpers/delayedEvent.js.coffee
```diff
@@ -0,0 +1,34 @@
...
+      cancelQueue(type, id)
+      event.apply(null, args)
+    else
+      cancelQueue(type, id)
+      queueStore[type][id.to_s] = $timeout(() ->
+        event.apply(null, args)
+      , TIME_TO_DELAY)
+
+
+  cancelQueue = (type, id)->
```
jonathoy: This would be cancelEvent, if following convention above.

app/assets/javascripts/angular/helpers/delayedEvent.js.coffee
```diff
@@ -0,0 +1,34 @@
+# Debounce manager used to auto update items
+#
+# Each item is given a unique timeout stored by type and id.
+# It can be cancelled and updated before fire,
+# or fired immediately, cancelling the standing delayed event.
+
+angular.module('helpers').factory('DelayedEvent', ['$timeout', ($timeout)->
+
+  queueStore = {}
+  TIME_TO_DELAY = 3500
```
jonathoy: Since we'll be sharing this delayed event logic, I wonder if it would be of benefit to pass in a custom TIME_TO_DELAY value.

acousticrobot: I thought about this, but didn't see a practical reason to add it now. Plus we already have optional `args` and `immediate`, so I didn't want to overload the function call.  If you find a reason for it to be handy in the future, we should revisit.

app/assets/javascripts/angular/helpers/delayedEvent.js.coffee
acousticrobot: I like it -- I wasn't happy with the naming and this gives me a good start -- more changes coming.

PR #2894, opened 2017-01-23
2740 omit drafts from gradable

jonathoy:
### Status
READY

### Description
This PR is a review of the Gradable concern, as well as various other locations throughout the app to better handle "draft" submissions.

We mention "drafts", since really our intent in these situations is to filter out unsubmitted submissions. I have renamed the method accordingly in ade9627.

As a future task, it might be advantageous to refactor the `draft_submission` factory from:

```
factory :draft_submission do
  text_comment_draft "Dear professor, "
  submitted_at nil
end
```

to:

```
factory :draft_submission do
  text_comment_draft "Dear professor, "
end

trait :unsubmitted do
  submitted_at nil
end
```

### Migrations
NO

- Ensure that when grading next for rubric graded assignments, the path points only to the next student with an ungraded submitted submission if there is one present
- Submission exports should work as before, where unsubmitted submissions are not included
- Counts for assignment analytics should not include unsubmitted submissions


app/models/concerns/gradable.rb
```diff
@@ -74,13 +74,13 @@ def ungraded_students(ids_to_include=[], team=nil)
     students - (User.find(grades.graded_or_released.pluck(:student_id)) - User.find(ids_to_include))
   end
 
-  def ungraded_students_with_submissions(ids_to_include=[], team=nil)
-    ungraded_students(ids_to_include, team) & (User.find(submissions.pluck(:student_id)|ids_to_include))
+  def ungraded_students_with_submitted_submissions(ids_to_include=[], team=nil)
```
acousticrobot: I don't know if it's better to leave the method name as: `ungraded_students_with_submissions` for simplicity, since we don't need `ungraded_students_with_unsubmitted_submissions`, but I'll leave that call up to you.

spec/models/assignment/student_submissions_spec.rb
```diff
@@ -6,82 +6,6 @@
   let(:assignment) { create(:assignment) }
   let(:student) { create(:user) }
 
-  describe "#student_submissions" do
-    subject { assignment.student_submissions }
```
acousticrobot: 👍

app/models/concerns/gradable.rb
jonathoy: Good point - I've gone ahead and changed it back since it does sound a little bit obtuse.

PR #2897, opened 2017-01-23
Recalculate all scores for a course on update

### LINKED ISSUE #1249: Recalculate scores for students when add_team_score_to_students? changes
Changing the add_team_score_to_students? boolean on the Course model will potentially impact the cached scores of all students in that course. In order to avoid added complexity in tangential helper methods, scores should be recalculated for all students in the course via either ScoreRecalculatorJob enqueuing or another more efficient method whenever the add_team_score_to_students boolean is changed on a course.

Current as of 899cf45bda8bf20dafb9893d43d43f22c9bbde5a on master.
https://github.com/UM-USElab/gradecraft-development/blob/master/app/models/course.rb

jonathoy:
### Status
READY

### Description
To ensure that there are no potential issues with cached scores, whenever the `add_team_score_to_student` flag changes all scores in the course should be recalculated.

Fixes #1249 

### Migrations
NO

The ScoreRecalculator job should run for every grade associated with the course, if the `add_team_score_to_student` field changes.

### DIFF
--- a/app/controllers/courses_controller.rb
+++ b/app/controllers/courses_controller.rb
@@ -76,6 +76,7 @@ class CoursesController < ApplicationController
     authorize! :update, @course
     respond_to do |format|
       if @course.update_attributes(course_params)
+        @course.recalculate_student_scores if add_team_score_to_student_changed?
         bust_course_list_cache current_user
         format.html do
           redirect_to edit_course_path(@course),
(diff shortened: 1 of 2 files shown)


app/controllers/courses_controller.rb
```diff
@@ -75,7 +75,9 @@ def copy
   def update
     authorize! :update, @course
     respond_to do |format|
+      add_team_score_to_student_changed = add_team_score_to_student_changed?
```
chcholman: I think you don't need this assignment - you should just be able to add a question mark to the method call beneath it

jonathoy: Yep you're right - good catch.

Issue #2790, opened 2016-12-31
Rework styles on the Assignment Settings page


chcholman:
### Reference:
https://trello.com/c/PEcKtEEx

PR #2905, opened 2017-01-24
2790 assignment review settings

mariehooper:
### Status
**IN DEVELOPMENT**

### Description
This PR restyles the assignment settings page to provide an overview of point distribution in the index and eventually a new page for each assignment type with a table of assignments and their settings for a quick review after setting up assignments.

### To-do
- [x] confirm which levels to show
- [x] make 'total point' maximum upper threshold for highest grade level * 110%
- [x] style for mobile (if needed after extending assignments classing)
- [x] add logic/styling to handle when any one assignment type's points exceed the total for the top level threshold
- [x] specs
- [x] add accordions with table under each

* Assignment settings page for instructors

Closes issue #2790


mariehooper:
@chcholman  - not sure if showing all the levels is the way to go. I know the dev classes are not a great example since the max points versus grading scheme is skewed due to number of assignments/points... but this looks a lot more crowded to me than I would love. What do you think? Ps - I left the original table in tact as a partial `app/views/assignments/_settings.html.haml` so that when the interaction layer and additional templates are added, it will be easier to grab the tables from what is there.

chcholman:
@mariehooper I actually really like it - it is crowded, but I think that's an incredibly valuable piece of information we can give them if that's the case. I do think we either need to find a way to stagger the height of the grade letters on the top chart, or else move them into the hover state. 

I think this is going to be super useful, and I'm really excited to get to see it in real classes - depending on what we see against real data, we may want to consider tweaking.

PR #2914, opened 2017-01-27
2757 refactor lti create

### LINKED ISSUE #2757: Refactor lti_create to use the CreatesOrUpdatesUser Service

jonathoy:
### Status
READY

### Description
This PR is a refactor of the user and course creation process when authenticating via LTI. The replacement of the `find_or_create_by_lti_auth_hash` method in the User model with `Services::CreatesOrUpdatesUserFromLTI` allows us to reuse some of the existing logic in the preexisting services.

The original goal for pulling the `find_or_create_by_lti_auth_hash` logic from the Course model and into `Services::CreatesOrUpdatesCourseFromLTI` was so that we could include the services together in one organizer chain and potentially have actions rolled back on failure.

We can revisit this approach at a later time if it is still desired. The CourseMembership creation can also be pulled from the CourseMembership model and moved into a service as well if we want to be totally consistent in the refactor.

Note: Since we are using the OmniAuth-LTI strategy already in UserSessionsController, I have simplified the parsing of relevant fields from the auth hash. See `Services::Actions::ParseCourseAttributesFromAuthHash` and `Services::Actions::ParseUserAttributesFromAuthHash` for details. The auth hash, once passed through the middleware, should be of type `OmniAuth::AuthHash`.

### Gem dependencies
N/A

### Migrations
NO

Ensure that our Canvas users are still able to authenticate as they did before.

======================
Closes #2757


chcholman:
@jonathoy Just tested this out on the staging server - worked perfectly. Fantastic job refactoring this!

Issue #2567, opened 2016-10-14
Create a view to handle when a user has no course


chcholman:

jonathoy:
<img width="1288" alt="screen shot 2016-12-20 at 9 16 07 am" src="https://cloud.githubusercontent.com/assets/17604722/21353655/021264de-c695-11e6-85b7-d2ea157c74dd.png">

@chcholman We had briefly discussed how to best word this message. Let me know if you come up with some ideas.

chcholman:
@jonathoy Apologies for the delayed reply. Let's do: 

"You don't belong to any courses in GradeCraft! 

If you're a student, ask your instructor to add you to their course site. 

If you're an instructor, email help@gradecraft.com to have them create a course shell for you. 

Log me out..."

PR #2931, opened 2017-02-01
2854 lti launch with email but no name

### LINKED ISSUE #2854: Handle LTI launch when there's an email but no name
### Expected Behavior
When the app receives an LTI launch that includes an email but no name, it should respond with an error message that prompts the user to confirm that the LTI setup is set to "Public" 

Investigate how this error is handled and whether we should be more detailed.

jonathoy:
screen shot 2017-02-01 at 10 36 15 am

### Status
READY

### Description
When a user authenticates from LTI with an email address but no first or last name information, we should provide a custom error that states that they need to check their LTI configuration.

This PR addresses that need by introducing a custom 500-like error page that renders a partial with the customized message and information.

To do this, we redirect to the ErrorsController#show method with the `error_type` corresponding to the name of the partial and the custom `status_code`, if desired. This page can be reused to handle other caught exceptions in the app.

As part of this change, I split out some conditional logic from our one default layout and moved it to another. I have also revised how a user is handled in the app when they have no course memberships, so that they are directed to this new page via a `before_action` filter.

### Gem dependencies
N/A

### Migrations
NO

- [ ] As a user with no course memberships, you should be redirected as before to the error page as described in #2567 
- [ ] Test that a user authenticating via LTI is directed to the appropriate error page when email information is provided but not first or last name
- [ ] The addition of the `require_course_membership` filter to ApplicationController should not cause any unexpected restrictions

======================
Closes #2854


chcholman:
@jonathoy Will you pull the focus flag and deal with the `earned_badges_controller` conflict?

chcholman:
@jonathoy Just tested this out, it looks great. Thanks so much!

PR #2932, opened 2017-02-01
Revise method definition to use submitted_at

### LINKED ISSUE #2909: Submissions this week method should not account for draft updates
In the `submission` model, there is the following method:

```
def self.submitted_this_week(assignment_type)
  assignment_type.submissions.submitted.where("submissions.updated_at > ? ", 7.days.ago)
end
```

Since this query relies on the `updated_at` timestamp, which is updated automatically on any record change, this may not properly represent submissions that are submitted this week. This probably should be updated to go off of the `submitted_at` date.

jonathoy:
### Status
READY

### Description
The method `submitted_this_week` on the `Submission` model was previously based off of the `updated_at` timestamp, which potentially causes issues since the field is automatically updated.

This has been revised to compare against the `submitted_at` time.

### Gem dependencies
N/A

### Migrations
NO

Observe the weekly stats on the dashboard and confirm that only Submissions submitted within the past week are included in the submitted this week count.

Weekly Stats

======================
Closes #2909

### DIFF
--- a/app/models/submission.rb
+++ b/app/models/submission.rb
@@ -61,7 +61,7 @@ class Submission < ActiveRecord::Base
   multiple_files :submission_files
 
   def self.submitted_this_week(assignment_type)
-    assignment_type.submissions.submitted.where("submissions.updated_at > ? ", 7.days.ago)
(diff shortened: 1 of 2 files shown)


chcholman:
@jonathoy Glad to see the solution for this was so simple!

PR #2938, opened 2017-02-02
2889 html safe bug

jwright:
### Status
**READY**

### Description
Ensures that the assignment description is an empty string instead of `nil` before wrapping it with an `#html_safe` call.

### Migrations
NO

```sh
git pull --prune
git checkout <feature_branch>
bundle; script/server
```

1. Link Canvas course id 127990
2. Import an assignment
3. There should not be an error

* Importing assignments from Canvas

======================
Closes #2899

### DIFF
--- a/app/views/assignments/importers/assignments.html.haml
+++ b/app/views/assignments/importers/assignments.html.haml
@@ -16,7 +16,7 @@
         - @assignments.each do |assignment|
           %tr
             %td= assignment["name"]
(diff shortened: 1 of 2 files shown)


jwright:
@chcholman I cannot test this with this specific Canvas course. Can you test this on staging/production and let me know if this does in fact, fix the issue? I don't know why the previous fix would not of fixed it.

PR #2953, opened 2017-02-08
2942 add tooltip helper

### LINKED ISSUE #2942: Add tooltip helper
Make these more extensible by creating helper that contains accessibility considerations and applicable classes

mariehooper:
### Status
**IN DEVELOPMENT**

### Description
This PR creates a helper for tooltips in the app so that accessibility considerations and necessary class names are built-in.

### Todos
- [x] Add right, left, bottom positioned tooltips
- [x] Update tooltip positioning as needed in app
- [x] Tests

* All tooltips

======================
Closes #2942

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -65,15 +65,15 @@ module ApplicationHelper
     "#{category}/#{arguments.join("/")}"
   end
   
-  def tooltip(tooltip_id, icon, &block)
+  def tooltip(tooltip_id, icon, placement="top", &block)
     content_tag(:span, class: "has-tooltip", tabindex: "0", "aria-describedby": tooltip_id) do
       concat(glyph(icon))
-      concat(tooltip_content(tooltip_id, &block))
+      concat(tooltip_content(tooltip_id, placement, &block))
     end
   end
 
-  def tooltip_content(tooltip_id, &block)
+  def tooltip_content(tooltip_id, placement="top", &block)
     hover_content = capture(&block)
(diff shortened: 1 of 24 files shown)


mariehooper:
@chcholman - I would love to get your initial thoughts on this structure for these helpers. You will see two that work together or separately for this... the `tooltip` helper is used when an **icon** is the trigger, which accounts for the vast majority of our tooltip use in GradeCraft. However, there are about 3 instances where we use either an img or text as the trigger for a tooltip. In those cases, I am simply using `tooltip_content` to handle the tooltip itself and relying on our original structure for the tooltip trigger.

Also checkout the styling - there are 4 available 'directions'. Right now most are 'top' and a few are 'right'. I would love to hear any/all feedback about this. My goal was to make it easier for roles, tags, tabindex and necessary classes to be applied to help ensure these stay accessible.

app/helpers/application_helper.rb
```diff
@@ -64,4 +64,20 @@ def multi_cache_key(category, *args)
...
+  def icon_tooltip(tooltip_id, icon, placement="top", &block)
+    content_tag(:span, class: "has-tooltip", tabindex: "0", "aria-describedby": tooltip_id) do
+      concat(glyph(icon))
+      hover_content = capture(&block)
+      concat(content_tag(:span, hover_content, class: "display-on-hover hover-style hover-style-#{placement}", id: tooltip_id, role: "tooltip"))
+    end
+  end
+
+  def img_tooltip(tooltip_id, image_src, placement="top", &block)
+    content_tag(:span, class: "has-tooltip", tabindex: "0", "aria-describedby": tooltip_id) do
```
chcholman: Looks like we still need some test coverage here?

PR #2958, opened 2017-02-08
update text to use term for pass/fail

### LINKED ISSUE #2955: Student assignment show page not showing custom course terminology
As a student, my “win/lose” assignment shows my grade as “pass” in the assignment index and on the assignment show page after it has been graded. As an instructor, my “win/lose” assignment shows the “pass/fail” terminology when i have changed this to “win/lose"

Appears that term_for is not working to display the correct term in this case when custom terminology is being used.

mariehooper:
### Status
**READY**

### Description
This PR updates the pass/fail grading partial to use the customized course terminology if applicable. 

* Student and instructor assignment index and show/grading pages

======================
Closes #2955


chcholman:
@mariehooper Thanks so much for doing this! I had been working on fixing this bug in another branch, and pulled in a change from your branch that you identified needed to be done as well - great catch! I'm closing this and bringing the work in over there.

PR #2969, opened 2017-02-10
2936 import pass fail grades

### LINKED ISSUE #2936: Import pass/fail grades
If the assignment is pass/fail, then the import utility should allow them to upload 1's and 0's for pass/fail and have that record properly.

jonathoy:
### Status
READY

### Description
Since the import grade process does not currently account for Pass/Fail type assignments, we want to be able to upload a score of 1 or 0 to specify whether the student has passed or failed, respectively.

### Gem dependencies
N/A

### Migrations
NO

For any given Pass/Fail type assignment, you should be able to Import Grades via CSV and input a 0 or 1 for the score to specify the status.

Grade Import via CSV

======================
Closes #2936

### DIFF
--- a/app/controllers/grades/importers_controller.rb
+++ b/app/controllers/grades/importers_controller.rb
@@ -20,7 +20,6 @@ class Grades::ImportersController < ApplicationController
     @assignments = syllabus.assignments(params[:id])
   end
 
-  # rubocop:disable AndOr
   # GET /assignments/:assignment_id/grades/download
   # Sends a CSV file to the user with the current grades for all students
   # in the course for the asisgnment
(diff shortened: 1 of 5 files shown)


jonathoy:
I also noticed this oddity but wasn't sure which part of the css was causing the issue:

screen shot 2017-02-10 at 1 29 52 pm

jonathoy:
Instead of using `to_i` on the score I changed it to use `Integer()`, which throws an exception if the value is considered invalid. Although handling exceptions is arguably less performant, it seems to me that it would be a fairly uncommon occurrence if its known that the field should be numeric.

chcholman:
@jonathoy This looks great! I made a tiny update to the breadcrumb for the imported results page. One last thing to handle: could you possibly handle conditions where they've got non-0 or 1 values in the spreadsheet? I'm worried that could happen all too easily, and it will break the whole assignment page currently.

jonathoy:
@chcholman I did some minor revisions to the error handling so that it was a bit cleaner. It's ready for you to take a look at whenever it is convenient!

chcholman:
@jonathoy Awesome, works perfectly. Thank you!

PR #2964, opened 2017-02-09
2941 resubmission after grading

jonathoy:
### Status
READY

### Description
Currently, if an existing submission is edited it is considered a resubmission even if it has not been graded yet.

Also, to avoid grades and submissions from becoming out of sync. We want to ensure that a student cannot edit their submission if an unreleased grade exists. Once the grade has been released, then it can be resubmitted if the assignment allows it.

### Gem dependencies
N/A

### Migrations
NO

### Steps to Test or Reproduce

- Ensure that the Edit Submission/Edit Draft button does not appear on the assignments show page for a student if it has been graded and the grade has not yet been released.
- Ensure that the routes to Edit/Update an existing submission for a student are protected if the submission has been graded and the grade has not yet been released.
- Verify that submission edits are not considered resubmissions until the submission has been graded and released.

### Impacted Areas in Application
Submissions

======================
Closes #2941, #2965

PR #2982, opened 2017-02-14
Revert "2941 resubmission after grading"

chcholman:
Reverts UM-USElab/gradecraft-development#2964


jonathoy:
@chcholman I just noticed this PR - please let me know how and if this needs to be revisited.

PR #2988, opened 2017-02-16
2706 predictor graph unlocks

### LINKED ISSUE #2706: Grade Predictor should reference unlock requirements at each predicted total grade
### Expected Behavior
As an instructor, if I set an unlock requirement on a course grade, I expect students not to be able to predict that grade without having achieved/predicted success on those items.

### Actual Behavior
The Grade Predictor chart is only paying attention to total points predicted/earned.

acousticrobot:
### Description

This PR separates the predicted points for the  locked assignments from the unlocked assignments in the grade predictor graph, so students can see which points they intend to earn that still require an unlock for completion.

closes #2706 

### Todos

@chcholman, @mariehooper I would like a little feedback on the UX before merge. I'm not totally happy with the key, but I'm not sure what would be a better solution. Here's the problem:

screen shot 2017-02-16 at 9 25 55 am

  * I feel like the locked (light blue) should be at the right end of the graph, since unlocked/predicted (dark blue) are closer to completion (green).
  * I feel like calculating out "Total Predicted" is ultimately the most useful number to show a student, but labeling the dark blue as "Total Predicted" is misleading since this number includes "all the blue", and labeling the light blue as "Total Predicted" would also be confusing.
  * I feel listing **Earned** AND **Predicted/Unlocked** AND **Predicted/Locked** AND **Total Predicted** AND **Predicted Final Grade** might be too much clutter in the key unless we redesign the key so that it is visually better separated from the graph above and the predictor below.
  * I want the Locked Predicted Points to only show if it is greater than zero (visible in the graph).
  * This is all coupled with the fact that these are svg elements, and so responsive design solutions take a lot of work to work in all browsers. Any solution that has elements left aligned will be much simpler than trying to mimic a float (think absolute positions, with `<g>` widths subject to numbers 0 to 100,000+).

One Solution that I've considered: turn the color key for "Total Predicted" into a diagonally bisected square, half light blue, half dark blue.

### DIFF
--- a/app/assets/javascripts/angular/directives/predictor/graph.coffee
+++ b/app/assets/javascripts/angular/directives/predictor/graph.coffee
@@ -15,6 +15,9 @@
       scope.allPointsPredicted = ()->
         PredictorService.allPointsPredicted()
 
+      scope.lockedPointsPredicted = ()->
+        PredictorService.lockedPointsPredicted()
+
       scope.predictedGradeLevel = ()->
         PredictorService.predictedGradeLevel()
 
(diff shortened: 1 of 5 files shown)


mariehooper:
@acousticrobot @chcholman - what if the fill for the 'locked' points was a stripey or not solid version of the same blue used in total predicted rather than light blue? This would help show that these aren't points you can count on while not creating confusion between what 'Total Predicted' is.

chcholman:
@acousticrobot @mariehooper Funny - I came up with the same solution as Marie! Switching the blues to the same base color, and doing the locked+predicted points section as a pattern on top of that blue. Otherwise your requirements for the display make total sense, and look great!

mariehooper:
@acousticrobot @chcholman Weee weee! I love it! Awesome work, Jonathan!

chcholman:
@acousticrobot Looks fantastic! Merging :)

PR #2997, opened 2017-02-17
add impersonating? to current_course logic

### LINKED ISSUE #2926: Faculty impersonation sends them to student's last GC course
As a faculty member, when I impersonate a student, I expect that it will send me to the course site I'm currently in. Instead, I am sent to whatever course that student was last in - even if I wouldn't have access to it normally!

acousticrobot:
### Status
**READY**

### Description
 
Adds a query in the `current_course` scope to check the session for an impersonating user. If present, use this user to determine `current_course_for(user,...)`

I think this is the best way to ensure that student previews remain inside the professor's course, without affecting any student sessions.  It's possible though that there are other ways to circumnavigate the permissions that would normally restrict what the professors could view.

closes #2926

### DIFF
--- a/lib/current_scopes.rb
+++ b/lib/current_scopes.rb
@@ -6,7 +6,8 @@ module CurrentScopes
 
   def current_course
     return unless current_user
-    @__current_course ||= CourseRouter.current_course_for(current_user, session[:course_id])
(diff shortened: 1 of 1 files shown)


chcholman:
@acousticrobot I like this solution a lot. Works perfectly!

PR #3000, opened 2017-02-17
Fix uncaught exception thrown in CSVGradeImporter

### LINKED ISSUE #2999: A non-integer grade causes grade import to fail under certain conditions
For a non pass/fail graded assignment, if a non-integer value is provided in the CSV the import will fail.

jonathoy:
### Status
READY

### Description
If the assignment is not of pass/fail type and a non-integer grade is found, the import will fail.

See [this](https://rollbar.com/gradecraft/gradecraft-development/items/1918/) Rollbar error for additional details.

### Gem dependencies
N/A

### Migrations
NO

Ensure that invalid grades do not cause the import to fail. They should be marked as invalid on the summary screen with the error message: "Grade is invalid"

Grade Import

======================
Closes #2999

### DIFF
--- a/app/importers/grade_importers/csv_grade_importer.rb
+++ b/app/importers/grade_importers/csv_grade_importer.rb
@@ -29,7 +29,7 @@ class CSVGradeImporter
             append_unsuccessful row, "Grade not specified"
             next
           end
-          if assignment.pass_fail? && !is_valid_pass_fail_grade(row.grade)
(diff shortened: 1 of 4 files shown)


chcholman:
@jonathoy Looks great, thank you! I'm curious if we'll need to build out more robust/separate methods for the different types of validating/importing in the future, but like the simple implementation to start with.

PR #3026, opened 2017-02-22
use GradeService in earned badge directive

acousticrobot:
### Status
**READY**

### Description

A bug has been discovered when awarding badges on the new grade page -- we were creating a bad request by not supplying a grade id.

It seems that the `grade.id` is not guaranteed to update via Angular magic when passed directly  to the `gradeEarnedBadgesSelect` directive.   This solution uses the `GradeService` aa a dependency, so that the id is bound correctly and available when the badge is awarded.


app/assets/javascripts/angular/directives/grades/earned_badges_select.coffee
```diff
@@ -14,12 +14,13 @@
       badge.available_for_student && !vm.badgeEarnedForGrade(badge)
 
     vm.awardBadge = (badge)->
+      debugger
```
chcholman: Do you want this debugger to stay in place?

Issue #3006, opened 2017-02-20
Canvas naming bug


chcholman:
Course names from Canvas are overriding names in GradeCraft

PR #3027, opened 2017-02-22
3021 Overridden Course Attributes From LTI

### LINKED ISSUE #3021: CreatesOrUpdatesCourseByUID overrides course name
### Expected Behavior
As an instructor, I expect to be able to customize my course name in GradeCraft, and to have that name persist.

### Actual Behavior
If I use the LTI link to click into GradeCraft, it updates the course name and course number to match a wrangled version of the Canvas course site. We should stop this update.

### Steps to reproduce
Step through the LTI Course creation process, then update the name in GradeCraft and relaunch - confirming that the edited name stays.

jonathoy:
### Status
READY

### Description
If the user is authenticating via LTI, we want to ensure that course attributes are not overridden if the course already exists in Gradecraft.

### Gem dependencies
N/A

### Migrations
NO

Ensure that course attributes are not overridden when authenticating via LTI for a course that has a matching `lti_uid` in Gradecraft.

======================
Closes #3021

### DIFF
--- a/app/controllers/user_sessions_controller.rb
+++ b/app/controllers/user_sessions_controller.rb
@@ -35,7 +35,7 @@ class UserSessionsController < ApplicationController
       and return unless result.success?
 
     @user = result[:user]
-    @course = Services::CreatesOrUpdatesCourseFromLTI.create_or_update(auth_hash)[:course]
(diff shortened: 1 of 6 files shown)


jonathoy:
@chcholman This may address #3006 as well

chcholman:
@jonathoy It does indeed - great catch, thank you!

PR #3031, opened 2017-02-23
Grade scheme elements refactor

### LINKED ISSUE #3030: Refactor Grade Scheme Elements Page
In anticipation for some upcoming features requested for the Grade Scheme Elements page, we should refactor and clean up the current implementation so that we are following a component-based pattern.

jonathoy:
### Status
IN PROGRESS

### Description
In anticipation for some upcoming features requested for the Grade Scheme Elements page, we should look at tidying up the current implementation so that it's more consistent and readable with the rest of our more recent Angular work.

### Todos
- [ ] Tests

### Gem dependencies
N/A

### Migrations
NO

- Validation occurs onblur, so any conflicts should appear when focus on the current input grade input is lost. Ensure that if there is not a zero threshold, one is automatically added to the form.
- Verify that GradeSchemeElements are being saved correctly and that errors are presented as expected.

Edit GradeSchemeElements

======================
Closes #3030


jonathoy:
@mariehooper Would you be interested in taking a quick pass at styling this page? 😄

jonathoy:
@acousticrobot I would like to pass this along to you once it's ready for review, but I did notice an oddity while using smartnumber on the GradeSchemeElements score inputs.

If I clear the value out in the input, the value is still being treated as 0 where I would expect it to be undefined. Since this page validates for a zero threshold, this causes an issue.

On line 18 and possible 15 as well, I propose that the inputValue be set to undefined rather than 0 because of this. However, I want to run it by you to make sure this isn't going to cause an issue elsewhere.

mariehooper:
@jonathoy Looking good and I have added styles for the alerts! One thing to consider: I like the way the '0' level shows up onBlur, however, I think we should add a similar 'on bar' alert that shows if a person tries to delete that level saying something like 'lowest level threshold must be 0'. Just for a little reassurance that the button is not broken. :)

jonathoy:
@mariehooper Thanks for taking a look at this! I had meant to ask you what your suggestion was for handling that auto-added element, and I think that is an appropriate message to give to the user.

jonathoy:
@acousticrobot If you have a chance, let me know what your thoughts are on these changes. If our goal is still to test the Angular work, I will separate that out in a different PR.

mariehooper:
I agree with @acousticrobot - within one point is a strange validator :)

jonathoy:
@acousticrobot @mariehooper Cait and I had discussed this as well whether it was still an issue to allow for thresholds that were within one point of each other, and I think we came to the conclusion that it wasn't. However, since we still have all sorts of back-end logic that accounts for `highest_points` I would want to do some additional investigation before removing this.

At a glance we would need to remove the block that "fixes" grade scheme elements params in the `GradeSchemeElementsController`, which adjusts the highest point to be one point lower than the next level's lowest points.

jonathoy:
@chcholman Would you like to take a look at this PR? I think that there's a lot of refactoring we can potentially do if we want to try to remove `highest_points`, but for now this Angular refactor can likely be moved on while we continue to work through GSEs iteratively.

PR #3034, opened 2017-02-24
2850 group student overrides

### LINKED ISSUE #2850: Add grade override per student to the group grade form
### Expected Behavior
As an instructor, I expect to be able to override a single student's grade directly within the grade form - currently you have to submit the group grade and then manually override per student. Consolidating this to one view and leveraging the points override field would be a great way to simplify this!

acousticrobot:
### Status
**ALMOST**

### Description

This PR adds points adjustment and feedback for each group member to the grading page.  It also removes access to editing individual grades, to keep all other fields in alignment on group grades.

In addition, it:
  * enhances the group presenter, to reduce clutter in the group table body partial
  * removes the service which iterated over group rubric grades, and instead makes a request for each group member with their own grade params.
  * updates the `GradeService` to handle both individual and group grades by storing grades in an array for all grading types.

### Gem dependencies
- ADDED to dev: [binding_of_caller](https://github.com/banister/binding_of_caller)

This was removed from [better_errors](https://github.com/charliesome/better_errors) but is necessary for the REPL in the interface.

### Migrations
NO

  * Grade for a group, test adjustment points for each member
  * Verify Grading for a group with a rubric
  * Verify standard grading still works
  * View Assignment page for a Group Assignment with Grades, see new and removed buttons.

  * This impacts all grading forms as the `GradeService` has been changed quite a bit.

======================
Closes #2850


app/assets/javascripts/angular/services/GradeService.coffee
```diff
@@ -101,32 +118,42 @@
           GradeCraftAPI.logResponse(response)
       )
 
-  _updateGradeById = (id, returnURL=null)->
-    $http.put("/api/grades/#{id}", grade: grade).then(
+  _updateGradeById = (id, params, returnURL=null)->
```
acousticrobot: Yes, we actually have pretty good closure in the services, but I was doing this just to keep track of which functions are only used internally.  I also have thought about exposing them in the return function to allow jasmine tests, at which point it would be more important that we follow this as a convention.

app/assets/javascripts/angular/services/GradeService.coffee
```diff
@@ -101,32 +118,42 @@
...
+    $http.put("/api/grades/#{id}", grade: params).then(
       (response) ->
-        grade.updated_at = new Date()
+        modelGrade.updated_at = new Date()
```
acousticrobot: There is a good chance this is a cary over from the old factory based service -- I'll double check and see if we can remove it before merge.

chcholman:
@acousticrobot I have some UX questions that I need to think through, and am going to chat with Marie tomorrow about a couple of possible design tweaks to make it more apparent what's happening, but the functionality of it is really feeling great. Two specific things I'm currently aware of: 

1. It looks like the creation of a new grade for a group may currently be broken? 
2. We should cover adjusting pass/fail for a specific student (that might be better off in another PR)

chcholman:
Per our discussion: 
- [ ] Remove Froala from Adjustment points feedback
- [ ] Move final grade display up to adjustment points for individual student
- [ ] Hide adjustment points under collapsible form section

chcholman:
@acousticrobot This is looking awesome. One tiny consistency thing: 
- [ ] Consistently order the students override forms alphabetically

PR #3038, opened 2017-02-27
Synchronize values and allow resetting to null for smart-number directive

### LINKED ISSUE #3033: Synchronize values and allow resetting to null for smart-number directive
Currently the smart-number directive has the potential to create out of sync values, specifically between the DOM input field and the ng-model.

See the blocks starting at line 13 and line 16 in the smart-number directive. In the former, a filtered input value of `'-'` or `'0-'` becomes `'-'` in the view, but `0` in the model. In the latter, a filtered input value of `' '` becomes `' '` in the view, but `0` in the model.

This can cause issues where, for instance, we clear an input field entirely. We would expect the model value to be null and not to be 0. This has to potential to interfere with validations.

Allow for the ability to specify an `allow-null` parameter where we might not want to return null and ensure that the value on the DOM element matches the model value when applicable.

jonathoy:
### Status
IN PROGRESS

### Description
There are some edge cases where it appears that the use of smartNumber causes the model value to become out of sync with the value on the input element on the DOM.

### Migrations
NO

======================
Closes #3033

### DIFF
--- a/app/assets/javascripts/angular/controllers/GradeSchemeElementsCtrl.coffee
+++ b/app/assets/javascripts/angular/controllers/GradeSchemeElementsCtrl.coffee
@@ -6,6 +6,7 @@
     $scope.grade_scheme_elements = $scope.gradeService.elements
 
     $scope.validateElements = (index)->
+      console.log "lowest_points: #{$scope.grade_scheme_elements[index].lowest_points}"
       for element, i in $scope.grade_scheme_elements
         $scope.gradeSchemeForm["lowest_points_#{i}"].$setValidity('directConflict', true)
         $scope.gradeSchemeForm["lowest_points_#{i}"].$setValidity('nearConflict', true)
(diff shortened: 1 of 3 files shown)


jonathoy:
@acousticrobot I found another oddity when taking a look at smartNumber, so I thought maybe it might be best to defer this work to you since you are most familiar with the directive.

In the commit that I made, I included some debugging output in console to demonstrate the current value represented in the model when making changes to the `lowest_point` threshold value on the grade scheme elements edit page.

My functional testing during this work has produced two edge cases:

1. On line 14 of smartNumber.coffee if the value is either `'-'` or `'0-'`, I believe the intent is that the input element should receive a hyphen (-) and the model should receive a value of 0. This works fine, but if our goal is to replicate how the traditional number input field works in HTML5 then by typing in a hyphen the model value should be undefined. Observe that if I change line 16 to the following:

`inputValue = if scope.allowNull then null else 0`

The directive no longer allows a hyphen as an input.

2. Notice how if I type an arbitrary number with more than one digit, `123456` for instance, and I input a comma after the 3 with `allow-negatives=true`, the input field is cleared out but the model value is replaced with the value prior to the comma location (e.g. the model value becomes 123).

jonathoy:
I was able to replicate this on the predictor by adding `allow-negatives'=>'true'. Since it wasn't set to anything, it automatically defaulted to false.

PR #3039, opened 2017-02-27
allow smartnumber to accept empty string

acousticrobot:
### Status
**EXPERIMENTAL, NEEDS TESTING**

### Description

This may be a simpler solution that the one I proposed for smart number.  It will require managing models fields which require non-null values from elsewhere, but could ultimately save some of the headaches associated with smart-number by reducing it further rather than trying to add functionality. 

I have tested it with the predictor -- it fixes a bug introduced in [this commit](https://github.com/UM-USElab/gradecraft-development/commit/0c4f4cc3b6211e4d6e3f4fff96835811646aac54) where the field breaks sincronicity with the model. I needed to update a couple summing functions to work with non-integer items but this was much easier than handling everything through the smart-number directive.

@jonathoy I would like to know if this solution works for your use case with GSE, if it does we can run through the other instances tomorrow and verify that they still work as well.

### Related PRs

#3033
#3038

### Todos
- [ ] Test with all uses of smart-number

### DIFF
--- a/app/assets/javascripts/angular/directives/common/smartNumber.coffee
+++ b/app/assets/javascripts/angular/directives/common/smartNumber.coffee
@@ -13,9 +13,6 @@
         if inputValue == '-' || inputValue == '0-'
           modelCtrl.$setViewValue('-')
           inputValue = 0
-        else if inputValue == ''
-          modelCtrl.$setViewValue('')
-          inputValue = 0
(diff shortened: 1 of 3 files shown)


jonathoy:
@acousticrobot I do believe this solves scenario #1 that I described in #3038.

PR #3040, opened 2017-02-27
2978 grading form specific styling

### LINKED ISSUE #2978: Style New Grading Form - Pass/Fail

mariehooper:
### Status
**IN DEVELOPMENT**

### Description
This PR standardizes the styling and structure of switches throughout the app as well as removes legacy pass/fail switch jQuery and styling.

### To-do
- [ ] decide if angular pass-fail-switch directive needs update with different input
- [ ] decide about styling for grade predictor switch styling

* Pass/Fail grading form
* Assignment with levels grading form
* Grade predictor pass/fail and badge switch
* Graded Rubric view (class analytics) switch

======================
Closes #2978


chcholman:
@mariehooper I totally agree that these should all be unified - but I'm not in love with the styling on that we're selecting. What would you think about using the switch style that we use on the Grade Predictor universally? Or something else? Very open to ideas, and happy to discuss pros/cons of the new consistent switch styling in person!

mariehooper:
@chcholman - Cool! I am thinking to delete this branch/PR then and make this a separate issue to consolidate styling for switches. The grading forms should be ready to go in terms of styling after the first pass through. I would like to ultimately make the switches all underlying checkboxes and have unified styling and I think that now stands separate from the grading scheme. Thoughts?

PR #3041, opened 2017-02-27
Google Calendar Integration

### LINKED ISSUE #1538: Integrate with Google Calendar
As both an instructor and student
I want to be able to sync GradeCraft timeline events (challenges, assignments, and events with dates) with Google Calendar

dovek:
### Status
**IN DEVELOPMENT**

### Description
Integrate Google Calendar with Gradecraft.

### Todos
- [x] Tests
- [x]  Write up Mocks

### Deploy Notes
Need to generate Google client and Secret for Gradecraft App. Need google_client_id and google_secret for Calendar API.

MUST be tested in non-dev environment to assure that user receives auth_failure page on any authorization error. One method of triggering is denying service when attempting to add event to google calendar for the first time.

### Gem dependencies
- ADDED [google-api-client 0.10](https://github.com/google/google-api-ruby-client) | REMOVED google-api-client, 0.8.2

### Migrations
NO

======================
Closes #1538


dovek:
@chcholman This is a first pass for the Google Integration. As you will see it isn't complete as I still have hard coded information into the event JSON object. I was wondering if you would take a look at it to give me any direction. Also, I'm using google-api-clinet 0.8.2. Currently google-api-client 0.10 is in alpha. I was wondering if you had any input on my using 0.8.2 as opposed to 0.9 or 0.10. Please let me know if you have any comments, questions, or concerns.

Gemfile
```diff
@@ -37,6 +37,8 @@ gem 'dotenv-rails'
 gem 'fast_blank'
 gem 'fog'
 gem 'font-awesome-rails', '~> 4.7'
+gem 'google-api-client', '0.8.2', require: 'google/api_client'
```
jonathoy: If we do decide to move to the latest stable, I would imagine that we can change this to:

gem 'google-api-client', '~> 0.9'

config/routes.rb
```diff
@@ -300,6 +306,10 @@
 
   get "lti/:provider/launch", to: "lti#launch", as: :launch_lti_provider
 
+  #Google Calendar
+  get 'auth/failure', to: redirect('/')
```
jonathoy: We may need to re-investigate why this route was needed, since there is already a route defined by:

`get "auth/failure", to: "pages#auth_failure", as: :auth_failure`

that would get overwritten by this line.

config/routes.rb
```diff
@@ -300,6 +306,10 @@
...
+  get 'signout', to: 'user_sessions#destroy', as: 'signout'
```
jonathoy: This route also seems redundant, since we already have: 

`get :logout, to: "user_sessions#destroy", as: :logout`

If we are using `signout_path` or some variation of a helper for this route, we can probably change it to something like `logout_path`

app/helpers/google_calendar_helper.rb
```diff
@@ -0,0 +1,34 @@
...
+  end
+
+  def create_google_secrets(google_authorization)
+    secrets = Google::APIClient::ClientSecrets.new({"web" =>
+      {"access_token" => google_authorization.access_token,
+        "refresh_token" => google_authorization.refresh_token,
+        "client_id" => ENV['GOOGLE_CLIENT_ID'],
+        "client_secret" => ENV['GOOGLE_SECRET']}
+        })
+    secrets
```
jonathoy: I think this can be reduced to just:
```
def create_google_secrets(google_authorization)
   Google::APIClient::ClientSecrets.new({"web" => {
      "access_token" => google_authorization.access_token,
      "refresh_token" => google_authorization.refresh_token,
      "client_id" => ENV['GOOGLE_CLIENT_ID'],
      "client_secret" => ENV['GOOGLE_SECRET']}
   })
end
```

app/views/events/index.html.haml
```diff
@@ -26,5 +26,6 @@
             - if current_user_is_staff?
               .right
                 %ul.button-bar
+                  %li= link_to decorative_glyph(:copy) + "Add to Google Calendar ", add_event_to_google_calendar_google_calendar_index_path(event), class: "button"
```
jonathoy: This will need to be a form POST action, if we follow the comment I made about resourceful routing.

app/controllers/events_controller.rb
```diff
@@ -1,8 +1,12 @@
 class EventsController < ApplicationController
+  include OAuthProvider
```
jonathoy: If the `before_action` block below is moved, I don't think that we need to include `OAuthProvider` here anymore.

config/routes.rb
```diff
@@ -188,6 +188,13 @@
     resources :courses, only: [:create, :destroy, :index], module: :integrations
   end
 
+  resources :google_calendar, only: [ :index] do
```
dovek: I tried to do this as a POST. There is a slight issue though: by convention POST calls do not contain a parameter. I need to send the event id so that the event object can be retrieved by the Google Calendar Controller. I looked around and couldn't find an example of a POST containing a parameter. Furthermore, it doesn't really make sense to make the call a POST because although the call relays to a Google API call that makes a POST and creates data, there is no data being created in the Gradecraft database.
I thought about making this a PUT/PATCH, but again, it doesn't modify any data in the Gradecraft database. For now, it's a GET. But I would be open to other suggestions for such a call, as there will likely be other calls where Gradecraft makes a call to Google to do some operation but does not affect Gradecraft.

jonathoy: We can leave it as-is if you feel that's how it should be. As a future note, you're right that a POST doesn't take parameters, but a POST request does have a request body. When we need to include data in a POST request we can include a hidden input field to the form with the proper value.

chcholman:
@dovek Will you update the pull request with the current state/what's left to be done when you have a chance?

spec/controllers/google_calendar_controller_spec.rb
```diff
@@ -0,0 +1,109 @@
...
+    let(:student_membership) { create :course_membership, :student, course: course }
+    let(:student) { student_membership.user }
+
+    before(:each) { login_user(student) }
+
+    describe "POST add_event_to_google_calendar" do
+      # Authorized User attempting add standard event
+      context "with an existing authentication" do
+        let! (:user_auth) { create :user_authorization, :google, user: student, access_token: "token", expires_at: 2.days.from_now}
+        describe "and a standard event" do
```
jonathoy: I don't think this `describe` block here is necessary. If the info describing that this is for a standard event is relevant, I would just include it on the `it` block underneath. In general, we usually only describe the methods we're testing.

spec/controllers/google_calendar_controller_spec.rb
```diff
@@ -0,0 +1,109 @@
...
+            expect(response).to redirect_to events_path
+            expect(flash[:notice]).to eq("Event " + standard_event.name + " successfully added to your Google Calendar")
+          end
+        end
+
+        # Authorized User attempting to add an event with no start date
+        describe "and an event with no open date" do
+          it "redirects to events path with an alert notice" do
+
+            stub_const('ENV', ENV.to_hash.merge('GOOGLE_CLIENT_ID' => 'GOOGLE_CLIENT_ID'))
```
jonathoy: Since these stubs are the same in this test and the test before it, I would place them in a `before` block to DRY this up.

spec/controllers/google_calendar_controller_spec.rb
```diff
@@ -0,0 +1,109 @@
...
+
+            post :add_event_to_google_calendar, params: { id: standard_event.id}
+
+            expect(response).to redirect_to events_path
+            expect(flash[:notice]).to eq("Event " + standard_event.name + " successfully added to your Google Calendar")
+          end
+        end
+
+        # Authorized User attempting to add an event with no start date
+        describe "and an event with no open date" do
```
jonathoy: I think I see what you're trying to do here with the `describe`s. In this case, my personal preference is that you just add this extra info to the `it` block of the test, although if you really wanted to break this up I would use a `context` instead of a `describe`. Typically, I don't feel the need to use an extra `context` block unless I have some `let` or `before` logic that should belong specifically in it.

spec/controllers/google_calendar_controller_spec.rb
```diff
@@ -0,0 +1,109 @@
...
+
+            post :add_event_to_google_calendar, params: { id: no_start_event.id}
+
+            expect(response).to redirect_to events_path
+            expect(flash[:alert]).to eq("Google Calendar requires event have both START and END time!")
+          end
+        end
+
+        # Authorized User attempting to add an event with no end date
+        describe "and an event with no due date" do
```
jonathoy: Same comment as above. Our convention has been to `describe` the methods.

spec/helpers/google_calendar_helper_spec.rb
```diff
@@ -0,0 +1,50 @@
+require "rails_spec_helper"
+
+  include GoogleCalendarHelper
+
+describe GoogleCalendarHelper do
+
+  describe "#get_google_authorization" do
+    let(:user) { create :user }
+    it "calls #refresh_if_google_authorization_is_expired if the token has expired" do
```
jonathoy: I've been called out on one of my first PRs for doing this and sounding too programmery (no hard feelings, you know who!), since the philosophy of RSpec is to be behavior-driven. It's a nitpick, but if we follow that language this might be better worded along the lines of: `it "refreshes the google authorization if the token has expired"`

spec/controllers/google_calendar_controller_spec.rb
```diff
@@ -0,0 +1,109 @@
+require "rails_spec_helper"
+require "api_spec_helper"
+
+describe GoogleCalendarController, type:[:disable_external_api, :controller] do
+  let(:provider) { :google_oauth2 }
+
+  before do
+    stub_request(:post, "https://accounts.google.com/o/oauth2/token").
```
jonathoy: These stubs look good! Though a minor suggestion, they can probably be moved into their respective contexts so that they don't have to be stubbed in situations where they are not needed. For instance, it looks like there is only one test where you would hit the request that returns a 401.

app/controllers/application_controller.rb
```diff
@@ -87,6 +88,10 @@ def event_session
 
   protected
 
+  def initialize_omniauth_state
+    session['omniauth.state'] = response.headers['X-CSRF-Token'] = form_authenticity_token
```
jonathoy: What was the reasoning for this addition? Since this is going into the application_controller, will this affect our other omniauth strategies?

PR #3044, opened 2017-02-28
Refactor class distribution boxplot

### LINKED ISSUE #1921: Refactor student boxplot to use D3
We show a boxplot of a student's performance against their peers in the student sidebar - this should be refactored to use D3 rather than sparkline.

mariehooper:
### Status
**READY**

### Description
This PR is the first of a series of refactors of GradeCraft's info visuals. This PR tackles the box plot for the class distribution for both student and instructor dashboards.

### To-do
- [x] show 'empty' graphic when no scores exist
- [x] update styling
- [x] decide about keeping or ditching tooltips for data points
- [x] format userScore on student box plot with commas

* Dashboard box plot

======================
Closes #1921


mariehooper:
@chcholman would love to get your thoughts on the styling - check out student in particular since you'll see the user score added in. I went with arrow but open to suggestion :) I also opted out of hover labels on the data points since we describe them below the graph...

chcholman:
@mariehooper Oh wow, it's looking so great! I love it. I really do like the arrow, I'm curious if people immediately 'get' that that's them. Would it be possible to add a label like 'You:' prior to the score? Open to other ideas, and yes, loving the removal of the hover state but would like to follow that up by removing the hand indicator.

chcholman:
@mariehooper This is looking awesome! What's the goal behind the instructor and student dashboards having different dimensions/scales? I'm guessing the instructor one looks different because it doesn't have the student arrow/text...is there any way we can make them slightly more standardized in regards to the height of the actual blue box?

PR #3060, opened 2017-03-02
3049 checkbox award badges

### LINKED ISSUE #3049: change earned badge award UX on grading page.
Awarding badges on the Grading Page should work via checkboxes. We need to change the directive, which currently used the custom start-plus-minus widget to work with check boxes.

acousticrobot:
### Status
**READY**

### Description

Rebuilds the `gradeEarnedBadgesSelect` directive and corresponding template to match proposed style and behavior.  Includes checkbox functionality. 

screen shot 2017-03-02 at 2 42 00 pm

### To Test

Award badges on a grade, verify that they persist. Verify on reload awarded badges are shown with a checkmark in the checkbox.

closes #3049

### DIFF
--- a/app/assets/javascripts/angular/directives/grades/earned_badges_select.coffee
+++ b/app/assets/javascripts/angular/directives/grades/earned_badges_select.coffee
@@ -4,18 +4,21 @@
     vm.BadgeService = BadgeService
     BadgeService.getBadges(vm.studentId)
 
-    vm.badgeEarnedForGrade = (badge)->
+    # Has the student currently earned this badge on this grade?
+    vm.badgeIsEarnedForGrade = (badge)->
       BadgeService.studentEarnedBadgeForGrade(vm.studentId, badge.id, GradeService.grade.id)
 
-    vm.badgeAvailable = (badge)->
-      badge.available_for_student || vm.badgeEarnedForGrade(badge)
+    # Can the badge be awarded or unawarded for this grade?
+    vm.badgeIsActionable = (badge)->
+      badge.available_for_student || vm.badgeIsEarnedForGrade(badge)
 
-    vm.badgeAwardable = (badge)->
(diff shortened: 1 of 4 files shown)


mariehooper:
Hey @acousticrobot - the functionality looks good! I saved and the badges persisted, etc. However, in order to comply with accessibility standards, the checkboxes need to have an associated label. This can be done by making the `.badge-name` element a label instead of a div and associating it to the checkbox by giving it a `for` attribute that is equal to an id given to the checkbox itself (something unique like` {{badge.name}}`. I did that on my local copy of your branch, but did not push it up because it created some weirdness with ngClick. The addition of the label allows the label itself to check the checkbox - so I think ngClick becomes unnecessary. When I click the label in the instance when the checkbox is already checked, I have to click twice. 

The other way to do it that doesn't require a unique ID or for attribute on the label is wrapping the whole input element in a label. Again, I think this would have to update how ngClick is being used...

mariehooper:
Hi @acousticrobot  - thanks so much for the detailed response. The double-checking is resolved and there are no accessibility flags right now except the lack of alt text for the badges which I don't mind at all fixing myself in a final pass through of the styling. 

I find the spinner to be somewhat unexpected since this is on a page with a 'Submit' button. Is there a reason we would not just have the update to the database occur when the Submit button is pressed and the page reloads rather than on each individual checkbox? I think that would make the whole process feel a lot more seamless to the end user, but I understand that there may be reasoning that I am missing for this update occurring on click.

chcholman:
@mariehooper @acousticrobot I definitely want the data to persist the way that @acousticrobot has it working, but I agree with you both that the spinner interaction is an odd one. I'm just as happy for it to not have an interaction at all, and for the update to be entirely in the background/unknown to the user. Thoughts?

PR #3061, opened 2017-03-02
Remove Highest Points from GSE

### LINKED ISSUE #3048: Remove highest_points from GradeSchemeElements
Remove all instances of `highest_points`

### LINKED ISSUE #3064: GSE breaks in Predictor Graph
### Expected Behavior

screen shot 2017-03-03 at 1 30 00 pm

### Actual Behavior

screen shot 2017-03-03 at 1 28 20 pm

### Steps to reproduce

Via git bisect, this bug was introduced somewhere in this commit:

https://github.com/UM-USElab/gradecraft-development/commit/44c69744d5e0b33f8779e289ed7d18e2c3b24ce1

jonathoy:
### Status
READY

### Description
In addition to removing the _within one point_ validation on the GradeSchemeElements mass_edit page, we should try to remove/reduce our use of `highest_points` in its current state as much as possible.

By explicitly calculating the upper limits of a level, we aim to follow this refactor up by removing the `highest_points` column altogether on the GradeSchemeElement table. See `grade_scheme_element.rb` for revisions related to the calculation of ranges and upper-limits.

This PR also fixes a preexisting bug in master where the upper point limit for the highest GSE was being set incorrectly.

We may also want to consider renaming and/or adding a unique constraint to `lowest_points` and `course` in order to maintain data integrity, since there is currently only client-side validation stopping an instructor from creating multiple GSEs with the same point threshold for a given course.

### Related PRs
- One to rename column/add unique constraint to `lowest_points`, `course`, if desired
- One to remove `highest_points` column, when ready

### Todos
- [ ] Update documentation, if needed

### Gem dependencies
- ADDED `binding_of_caller` for debugging, which was similarly done in #3034 

### Migrations
NO

- Ensure that editing and saving GradeSchemeElements works as expected
- Ensure that GSE-related views for students and instructors render as expected
- Verify that points to next level, progress percentage bars, and other statistics are correct

GradeSchemeElements

======================
Closes #3048 
Closes #3064

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -115,6 +115,7 @@ group :development, :test do
   gem 'pry'
   gem 'pry-byebug'
   gem 'better_errors'
+  gem 'binding_of_caller'
   gem 'jasmine-rails'
 end
 
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -90,6 +90,8 @@ GEM
       coderay (>= 1.0.0)
       erubis (>= 2.6.6)
       rack (>= 0.9.0)
+    binding_of_caller (0.7.2)
+      debug_inspector (>= 0.0.1)
     bson (4.1.1)
     builder (3.2.3)
     byebug (9.0.5)
@@ -133,6 +135,7 @@ GEM
     dalli (2.7.6)
     database_cleaner (1.5.3)
     db-query-matchers (0.6.0)
+    debug_inspector (0.0.2)
     diff-lcs (1.2.5)
     docile (1.1.5)
     dotenv (2.1.1)
@@ -629,6 +632,7 @@ DEPENDENCIES
   autoprefixer-rails
   aws-sdk (~> 2)
   better_errors
+  binding_of_caller
   cancancan
   capybara
   capybara-select2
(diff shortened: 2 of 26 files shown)


jonathoy:
@acousticrobot Good catch with the predictor. It was a simple fix related to my renaming of `elements` in the GradeSchemeElementsService to `gradeSchemeElements` and I have included it in this PR.

In regards to whether `highest_points` should be infinity vs. `full_points`, that raises a question that I have for you. What is the difference between `course.total_points` and `course.full_points`? In our test data, if I do a `Course.pluck(:full_points)` I get all `nil`s. If one represents the total amount of points that one can possibly earn in a course, then I suppose that might make more sense for the highest element rather than infinity. Additional thoughts on this @chcholman?

chcholman:
@jonathoy @acousticrobot `full_points` is legacy - it exists on the `Course` but is not actually used anywhere. It's been left because we are currently considering reinstating it, and trying to figure out where it would feel intuitive for instructors to declare and whether or not they actually know it. In the meantime, let's continue using `total_points`.

PR #3072, opened 2017-03-06
limit check/uncheck to items within the table

### LINKED ISSUE #2696: 'Check' on the Grading Status page should select only that table's grades
### Expected Behavior
As an instructor when I click 'Check' on the Grading Status page, it should should select only that table's grades. 

### Actual Behavior
It selects all grades that are unreleased in the entire course. 

### Steps to reproduce
1. Make sure that there are grades ready to be released for several different assignments. 
2. Go to the Grading Status page 
3. Click "Check" in table header for the first assignment
4. Observe that all unreleased grades have been selected

acousticrobot:
### Description

Limits jquery "check/uncheck" button action to items within the table.

note: I added a new jquery function, because I didn't want to affect other selectors handled by the old one. I'm also not sure if the view partial is used on other pages, if so, we should make sure it works as intended there.

I also outdented the table body in the partial, please verify that this is the proper indentation; it seems to work both ways.

1. Make sure that there are grades ready to be released for several different assignments. 
2. Go to the Grading Status page 
3. Click "Check" in table header for the first assignment
4. Observe that all only grades in that assignment have been selected

======================
Closes #2696

### DIFF
--- a/app/assets/javascripts/utilities.js
+++ b/app/assets/javascripts/utilities.js
@@ -57,6 +57,20 @@ $(".select-none").click(function(e){
 
 });
 
+// select all but only within the table
+$(".table-select-all").click(function(e){
+  var $link = $(this);
+  e.preventDefault();
+  $link.parents("table").find('input[type="checkbox"]').prop('checked', 'checked').trigger('change');
+});
+
+// select none, but only within the table
+$(".table-select-none").click(function(e){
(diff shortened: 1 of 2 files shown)


chcholman:
@acousticrobot Awesome fix, thank you. Tested every place this applies and it looks great.

PR #3085, opened 2017-03-08
Cleaner response for trying to view announcements from another course

chcholman:
### Status
**READY**

### Description
This PR improves the response to a student trying to load an announcement from a different course. 

### Migrations
NO

Create an announcement for one course. Log into a different course as a student, and explicitly load the announcement you created. It should redirect you to the dashboard with an alert message to try switching courses. 

* Announcements


chcholman:
@acousticrobot That's an awesome improvement, thank you!

PR #3105, opened 2017-03-14
2243 quick add grade letter levels

### LINKED ISSUE #2243: Allow instructors to add all grade letter levels at once via a button
This would create 6 levels (A-F) with no numbers entered by default

### LINKED ISSUE #2244: Allow instructors to add all +/- grades at once via a button
This would add A+, A-, B+, B-, C+, C-, D+, D- as grade letters with no numbers present, but in the appropriate order

### LINKED ISSUE #2245: Allow instructors to add X number of non-letter grades in bulk
As an instructor I should be able to bulk add empty GSE levels

jonathoy:
### Status
READY

### Description
The goal of this PR is to allow an instructor to quickly set up grade scheme elements by filling out a few related questions.

This requires us to remove the validation for `lowest_points` on the GradeSchemeElement model so that on submission of the setup questions, "empty" GradeSchemeElements can be created for editing on the `mass_edit` page

Grade scheme elements are not considered "valid" until they have a value for `lowest_points`, meaning they should not be visible or taken into account as a student.

### Migrations
NO

Instructors - Setting up new grade scheme elements and mass editing existing
Students - Grade scheme elements index

======================
Closes #2243 
Closes #2244 
Closes #2245

### DIFF
--- a/app/assets/javascripts/angular/directives/grade_scheme_elements/element.coffee
+++ b/app/assets/javascripts/angular/directives/grade_scheme_elements/element.coffee
@@ -9,8 +9,11 @@
     templateUrl: 'grade_scheme_elements/element.html'
     restrict: 'E'
     link: (scope, element, attrs) ->
+      # Set as invalid to disable submit, since we don't want the user to create elements
+      # on this page without lowest_points values
       scope.addElement = () ->
         GradeSchemeElementsService.addElement(@gradeSchemeElement)
+        @setFormIsInvalid()(true)
 
       scope.removeElement = () ->
         GradeSchemeElementsService.removeElement(@gradeSchemeElement)
--- a/app/assets/javascripts/angular/directives/grade_scheme_elements/element.coffee
+++ b/app/assets/javascripts/angular/directives/grade_scheme_elements/element.coffee
@@ -5,29 +5,20 @@
   {
     scope:
       gradeSchemeElement: '='
-      setFormIsInvalid: '&'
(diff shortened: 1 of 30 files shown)


jonathoy:
@mariehooper Would you mind taking a stab at the styling for the grade scheme elements setup questions and giving me your input on the implementation? You should be able to see a basic set of questions as an instructor when going to the grade scheme elements index page when there are no grade scheme elements for the course.

jonathoy:
I took a quick pass at styling this if you wanted to take a look - it's otherwise ready for testing.

chcholman:
@jonathoy Grade letters we don't need to create: D-, all E's, and F+/F-.

chcholman:
@jonathoy If you start the process but then click off, even if you don't enter point values, it keeps the created levels. That probably makes sense given the context, but I wonder about adding a "delete all" button for the case where you are playing around and accidentally create a lot of levels.

chcholman:
Ah, one final thing - if you say "No" to having grade letters, we want you to be asked how many levels you would like to create to start with.

jonathoy:
As a side note, given the changes that were made recently it wouldn't be too difficult to have the grade scheme elements autosave on the `mass_edit` page if that would be preferred.

chcholman:
@jonathoy This looks fantastic. Yes, I think it makes sense to have the elements autosave on the mass_edit page as well. I'll merge this work and add a ticket for that.

PR #3111, opened 2017-03-14
2896 Manually Activate Account first pass

dovek:
### Status
**READY**

### Description
Enable admins to activate student and faculty accounts manually instead of by email.

### Related PRs
N/A

### Deploy Notes
N/A

### Gem dependencies
N/A

### Migrations
NO

* Students Page
* Staff Page

======================
Closes #[2896]

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -95,6 +95,13 @@ class UsersController < ApplicationController
     redirect_to root_path, alert: "Invalid activation token. Please contact support to request a new one." and return unless @user
   end
 
+  def manually_activate
+    session[:return_to] ||= request.referer
+    @User = User.find(params[:id])
+    @User.activate!
+    redirect_to session.delete(:return_to), notice: "#{@User.first_name} #{@User.last_name} has been activated!" and return
+  end
+
   def activated
     @token = params[:token]
     @user = User.load_from_activation_token(@token)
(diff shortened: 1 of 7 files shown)


app/controllers/users_controller.rb
```diff
@@ -95,6 +95,13 @@ def activate
     redirect_to root_path, alert: "Invalid activation token. Please contact support to request a new one." and return unless @user
   end
 
+  def manually_activate
+    session[:return_to] ||= request.referer
```
jonathoy: It looks like there's already some logic in `ApplicationController` to handle keeping track of the request referrer. I think it would probably be better to add to the top of your controller the following:

```
before_action :save_referer, only: :manually_activate
```

so that you can remove this line.

app/controllers/users_controller.rb
```diff
@@ -95,6 +95,13 @@ def activate
...
+    @user = User.find(params[:id])
+    @user.activate!
+    redirect_to session.delete(:return_to), notice: "#{@user.first_name} #{@user.last_name} has been activated!" and return
```
jonathoy: I would make sure that you have a default in case there is no `return_to` path saved in the session for some reason - maybe something like this to go back to dashboard: 

```
redirect_to session[:return_to] || dashboard_path, notice: "#{@User.first_name} #{@User.last_name} has been activated!"
```

I believe that you shouldn't need the `and return` as well since this is the last line in your method.

PR #3113, opened 2017-03-15
3096 errant leaderboard ordering

### LINKED ISSUE #3096: Students out of order on the leaderboard
(this is the team leaderboard)

jonathoy:
### Status
READY

### Description
The team leaderboard is not ordering properly by score. By default, the students should also be listed in descending order by their scores.

### Gem dependencies
N/A

### Migrations
NO

- [ ] Team leaderboard sorting
- [ ] Team show
- [ ] Grade audit page for admins (`StudentsController#grade_index`)
- [ ] Users index
- [ ] Challenges view
- [ ] Assignments show

======================
Closes #3096

### DIFF
--- a/app/assets/javascripts/dynatable.js
+++ b/app/assets/javascripts/dynatable.js
@@ -14,7 +14,11 @@ $('table.dynatable').bind('dynatable:init', function(e, dynatable) {
         score: 'numeric',
         rank: 'numeric',
         date: 'date',
-        blog: 'alphanum'
+        blog: 'alphanum',
+        scoreWithWeights: 'numeric',
+        finalScore: 'numeric',
+        predictedScore: 'numeric',
+        id: 'numeric'
       }
     }
 });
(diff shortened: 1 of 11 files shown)


jonathoy:
@chcholman I did some digging and came across a lot of Dynatable-related commits that were of particular interest to me - this one for instance: 8c639e24faa62411d7193d50f9502858d974940d.

I see that Richard removed the `data-dynatable-sorts` attribute in `_in_team_rankings.html.haml` upon cleaning up `dynatable.js`, but then it was re-added later on. I looked through the app and there are various other places where we used to add custom sorting attributes and it appears that those sorting functions no longer exist.

A global search for `data-dynatable-sorts` shows more uses for this that don't seem to make sense anymore. I'm going to propose removing these, but we will want to do some testing to ensure that the sorting still works as expected in these locations. If I remove the attribute on the `/users/index.html` for instance, that seems to fix a similar issue.

PR #3120, opened 2017-03-16
Fix radio button grading for pass/fail assignments

### LINKED ISSUE #2956: Quick grade for pass/fail assignments gives wrong grading form
As an instructor, I have selected radio boxes as the setting for quick grade on  pass/fail assignments and i see a select box instead for each student

jonathoy:
### Status
READY

### Description
Previously if the quick grade style for an assignment was set as radio and the assignment was of pass/fail type, then the instructor would be presented with a select list. Radio buttons would only appear with this option if the assignment also had levels.

This is a quick fix for the bug mentioned in #2956 and not a PR that addresses the need for a larger usability refactor.

### Gem dependencies
N/A

### Migrations
NO

1. Set the quick grade styling for a Pass/Fail type assignment to radio buttons.
2. Ensure that grades persist on save and that there are radio buttons for when there are and are not assignment levels.

Assignment mass_edit

======================
Closes #2956

### DIFF
--- a/app/models/assignment.rb
+++ b/app/models/assignment.rb
@@ -237,11 +237,7 @@ class Assignment < ActiveRecord::Base
   end
 
   def grade_radio?
-    if pass_fail?
-      mass_grade_type == "Radio Buttons"
-    else
-      mass_grade_type == "Radio Buttons" && has_levels?
-    end
+    mass_grade_type == "Radio Buttons" && has_levels?
   end
 
   def grade_text?
--- a/app/models/assignment.rb
+++ b/app/models/assignment.rb
@@ -237,7 +237,11 @@ class Assignment < ActiveRecord::Base
   end
 
   def grade_radio?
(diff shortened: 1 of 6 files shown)


app/models/assignment.rb
```diff
@@ -235,7 +235,11 @@ def grade_select?
   end
 
   def grade_radio?
-    mass_grade_type == "Radio Buttons" && has_levels?
+    if pass_fail?
```
chcholman: I think this conditional can now be taken out, since we're handling this at the view level?

jonathoy: Yep! Although as a side note when I reflect back on our decorator/presenter discussion, I wonder if these `mass_grade_type`-oriented methods might be better left off the model since they mainly deal with the views.

chcholman: Totally agree with that direction

PR #3124, opened 2017-03-17
1949 archived limitations

### LINKED ISSUE #1949: Add 'archived' limitations
An 'inactive' course should 
- [ ] not allow you to create/edit grades
- [ ] not allow you to predict grades
- [ ] not allow you to submit assignments
- [ ] not allow you to award badges

jonathoy:
### Status
READY

### Description
Per the issue description, an inactive course should not allow you to:

- Create/Edit grades
- Predict grades
- Submit assignments
- Award badges

To achieve this without adding additional logic and bloat to the views, view helpers were added to `link_helper.rb` and `submit_tag_helper.rb` to conditionally render the appropriate element only when the course is active.

Use the latter vs the former when we need the element to display based on course status:
- `link_to` => `active_course_link_to`
- `submit_tag` => `active_course_submit_tag`
- `link_to_unless_current` => `active_course_link_to_unless_current`

Usage of these custom helpers is the same as their standard Rails counterparts.

### Migrations
NO

Pretty much any page that allows you to grade, submit, or award badges 😢 

======================
Closes #1949

### DIFF
--- a/app/assets/javascripts/angular/directives/grades/edit.coffee
+++ b/app/assets/javascripts/angular/directives/grades/edit.coffee
@@ -65,10 +65,9 @@
          recipientType: "@",
          recipientId: "=",
          submitPath: "@",
-         gradeNextPath: "@"
+         gradeNextPath: "@",
+         isActiveCourse: "="
         },
       templateUrl: 'grades/edit.html'
     }
 ]
-
-
--- a/app/assets/javascripts/angular/directives/student/submission/save_draft_button.coffee
+++ b/app/assets/javascripts/angular/directives/student/submission/save_draft_button.coffee
@@ -16,6 +16,7 @@
     restrict: 'C'
     scope: {
       assignmentId: '@'
+      isActiveCourse: '='
     }
     templateUrl: 'student/submission/save_draft_button.html'
   }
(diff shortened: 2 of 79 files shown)


jonathoy:
Observations while testing:

1. `http://localhost:5000/badges/5/earned_badges/65` is broken due to an undefined method error.
2. `assignments/grades#index` appears to be either an unused route or a hidden route. Either way, this view is currently broken.

jonathoy:
@acousticrobot Tagging you for review so you can take a look at the approach that we previously discussed.

jonathoy:
@acousticrobot Would you mind taking double-checking the predictor changes since you're the most familiar with it? As far as I can tell, changes no longer persist as a student for an inactive course. Feel free to test using the newly created course GC106 in the sample data.

chcholman:
@jonathoy On the Grading Status page, we should remove the ability to select grades (the check all/checkboxes) (the button to update them has been removed, but the interaction prompt still exists. 

screen shot 2017-03-25 at 10 13 23 am

chcholman:
@jonathoy On the Assignment Analytics page, we need to remove the Quick Grade button: 

screen shot 2017-03-25 at 10 15 24 am

chcholman:
@jonathoy Staff should not be able to create Announcements or Events

PR #3130, opened 2017-03-19
Remove touch: true unless there's a cache_key involved

chcholman:
### Status
**READY**

### Description
This PR removes a series of `touch: true`s that were added two years ago. The examples that are left are the three places (grades -> courses, submissions -> courses, and course_memberships -> courses and users) where we have cache_keys in place that may actually be utilizing this approach.

### DIFF
--- a/app/models/assignment.rb
+++ b/app/models/assignment.rb
@@ -9,8 +9,8 @@ class Assignment < ActiveRecord::Base
 
   attr_accessor :current_student_grade
 
-  belongs_to :course, touch: true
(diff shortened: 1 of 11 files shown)


jwright:
@chcholman This is a great change. This should speed up the application a bit too.

Have you tested the cached keys for this? For example, the cache key for current courses? Is that updated appropriately if an admin adds a user to a course still? I would recommend testing the cache keys manually or by adding specs (preferred).

PR #3136, opened 2017-03-20
3135 grade importer bug

### LINKED ISSUE #3109: Manage warnings regarding constants

### LINKED ISSUE #3135: CSVGradeImporter - Cannot update existing grade to fail on a pass/fail assignment
If a student already has a grade for the course and a CSV is uploaded with the intent on setting the status as failed, the status will not update.

The current method for determining whether the grade should be updated needs to be revised to accommodate the latest pass/fail enhancements.

jonathoy:
### Status
READY

### Description
If a grade already exists for a student on a pass/fail type assignment and an instructor attempts to set their status to `Fail` via CSV grade upload, no change will be made.

This is the result of a check that was previously put in place to prevent unnecessary grade updates.

This PR also addresses #3109, which pertains to warnings being generated due to already initialized constants.

### Migrations
NO

Upload a CSV and set a student's score to 1 to signify passing. Ensure that it is possible to update their grade to 0 afterwards to signify failing, and vice versa.

CSV Grade Upload

======================
Closes #3135 
Closes #3109

### DIFF
--- a/app/importers/grade_importers/csv_grade_importer.rb
+++ b/app/importers/grade_importers/csv_grade_importer.rb
@@ -2,9 +2,11 @@ require "csv"
 require "quote_helper"
 
 class CSVGradeImporter
-  PASS_FAIL_GRADE_VALUES = [0,1].freeze
+  include GradesHelper
   attr_reader :successful, :unsuccessful, :unchanged
   attr_accessor :file
+  
+  PASS_FAIL_GRADE_VALUES = [0,1].freeze
 
   def initialize(file)
     @file = file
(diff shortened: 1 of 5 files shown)


app/importers/grade_importers/csv_grade_importer.rb
```diff
@@ -113,9 +114,19 @@ def is_valid_grade?(assignment, grade)
     end
   end
 
+  # If the assignment is pass/fail type, update the grade if the status or feedback changes
+  # Else, update if the raw points or feedback changes
+  def update_grade?(row, grade, assignment)
+    score = Integer(row.grade || "")
```
acousticrobot: Did this work for you in testing? I think trying to coerce an empty sting into an integer results in an error.

jonathoy: I've gone ahead and added some commenting and clarification to the logic to document our discussion about this. As previously mentioned, the assumption is that the input will be validated prior to this point so that an exception from `Integer("")` should never occur.

PR #3137, opened 2017-03-20
2950 Set passwords manually - first pass

dovek:
### Status
**IN DEVELOPMENT**

### Description
As a K-12 instructor, I need a way to manually set/reset someone else's password.

### Related PRs
N/A

### Todos
- [ ] Tests

### Deploy Notes
N/A

### Gem dependencies
N/A

### Migrations
NO

As an administrator, navigate to learners, unique student and click "Edit" in the user's profile, then fill out change password fields and click the "Update User" button. Upon resetting a user's password, current user should see alert reading "[User]'s profile was successfully updated!".

* Administrator Views of Students

======================
Closes #[2950]


app/views/users/change_password.html.haml
```diff
@@ -0,0 +1,42 @@
...
+
+      - if ! @user.kerberos_uid.present?
+        .form-item
+          = f.label :password
+          = f.password_field :password, as: :password
+        .form-item
+          = f.label :password_confirmation
+          = f.password_field :password_confirmation, as: :password
+
+      %hr.dotted
```
jonathoy: We can try to clean up some of this view to remove some unneeded cruft, but it's looking good otherwise!

dovek:
@chcholman The changes to the view _gradecraft_user_form.html.haml seemed to have broken a feature test creating_a_new_user.spec. This is a feature test using capybara which seems to crawl through the form. I'm speculating that because the form has changed, the xpath to the fields in the form have changed as well and that is leading to the failing of the test. How would you like me to proceed? Would you like to discuss this? Thoughts?

chcholman:
@dovek I believe this is actually failing because it reveals a bug - will you test out the process of creating a new user and setting their password? The form submission fails because it reports the password confirmation doesn't match.

dovek:
@chcholman Yes. You are correct. The view for Editing a user looks the same as the view for creating a new user. And I'm getting the error when I try to create a new user as well.

app/views/users/_gradecraft_user_form.html.haml
```diff
@@ -13,5 +13,12 @@
     .form-item
       = f.label :display_name, "Pseudonym"
       = f.text_field :display_name
+    - if ! @user.kerberos_uid.present? && @user.persisted?
```
chcholman: Let's also check to see if the current_user is an admin

PR #3146, opened 2017-03-22
3133 update pass fail csv sample

### LINKED ISSUE #3133: Update download example file to show 1/0 if pass/fail

jonathoy:
### Status
READY

### Description
Current grades were not being represented in the sample file download found on the Import Grades page, if the assignment was of a pass/fail type.

This PR addresses this by returning either "0" or a "1" for the Score column, relative to the `pass_fail_status` on the grade.

Also included in this fix are changes to the other grade exporters, which were previously not returning relevant pass/fail grade information when the assignment was pass/fail type.

For a regular grade export on a pass/fail assignment, the value for the "Score" column on the CSV will be replaced with either "Pass" or "Fail", whereas before it was represented with `grade.score`.

### Migrations
NO

- Test grade exports for individual and group assignments, ensuring that "Pass" or "Fail" grades are represented in the CSV.
- Ensure that downloading the sample CSV from the Grade Import page returns the current up-to-date grades on a pass/fail assignment, represented by "0" or "1".

All CSV exports

======================
Closes #3133

### DIFF
--- a/app/exporters/grade_exporter.rb
+++ b/app/exporters/grade_exporter.rb
@@ -8,7 +8,7 @@ class GradeExporter
         grade = student.grade_for_assignment(assignment)
         csv << [student.first_name, student.last_name,
                 student.email,
-                score_for_assignment_type(assignment, grade, pass_fail_statuses_as_int) || "",
+                score_for_assignment(assignment, grade, pass_fail_statuses_as_int) || "",
                 grade.feedback || ""]
       end
     end
@@ -22,7 +22,7 @@ class GradeExporter
       groups.each do |group|
         grade = group.grade_for_assignment(assignment)
         csv << [group.name,
-                score_for_assignment_type(assignment, grade, false) || "",
+                score_for_assignment(assignment, grade, false) || "",
                 grade.feedback || ""]
       end
     end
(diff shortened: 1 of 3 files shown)


jonathoy:
@chcholman This might be scope creep in terms of the issue at hand, but while I was in the `GradeExporter`, it had appeared that it needed to account for pass/fail grades elsewhere in CSV-related exports. Let me know if the changes that I proposed are desired.

app/exporters/grade_exporter.rb
```diff
@@ -1,30 +1,35 @@
...
+  # if pass_fail_statuses_as_int is true
+  def export_grades(assignment, students, pass_fail_statuses_as_int=false, options={})
     CSV.generate(options) do |csv|
       csv << headers
       students.each do |student|
         grade = student.grade_for_assignment(assignment)
         csv << [student.first_name, student.last_name,
                 student.email,
-                grade.score || "",
+                score_for_assignment_type(assignment, grade, pass_fail_statuses_as_int) || "",
```
chcholman: I'm cautious about the name (the `assignment_type` piece) given that we have `AssignmentType` as a model itself. Perhaps just `score_for_assignment`?

app/exporters/grade_exporter.rb
```diff
@@ -57,4 +61,20 @@ def headers
...
+      pass_fail_status_as_int(grade.pass_fail_status)
+    else
+      grade.score
+    end
+  end
+
+  def pass_fail_status_as_int(status)
+    case status
+    when "Pass" then 1
+    when "Fail" then 0
```
chcholman: It looks like we might not be testing this condition?

jonathoy: I've added a condition to test this, thanks!

app/exporters/grade_exporter.rb
jonathoy: I agree with this change.

PR #3157, opened 2017-03-27
Fix Predictor Bug

### LINKED ISSUE #3154: Student Predictor View Sometimes Does Not Reflect Pass/Fail Status
This is a odd one.

The issue is that if students are graded via CSV for a pass/fail assignment, their grade is not reflected on the student's predictor page. This seems to be because the `PredictorService` conditionally renders the toggle switch vs a graded state based on the grade.score.

See the condition block starting at line 129 of `PredictorService`:

```
if article.grade.score == null
  return false
else
  return true
```

The problem is that a pass/fail grade ultimately hits this conditional statement, but there are inconsistencies with how a grade score is set in the app. If the grade is created via CSV import, the grade score is `nil`. If the grade is created in quick grade, for instance, the grade is 0.

Since the documentation is a bit murky as to what the true value of the grade score should be on a pass/fail type assignment, we should do or discuss the following before issuing a fix:

- We should agree on whether the grade.score should be `nil` or `0` on a pass/fail assignment. Code should be updated to be consistent.
- We should update documentation
- We should update the method on `PredictorService`, as needed
- We should add sample data for completed Pass/Fail grades

jonathoy:
### Status
READY

### Description
This PR reorders the `calculate_points` and the `zero_points_for_pass_fail` callbacks on the Grade model to ensure that a Grade score is always set to `0` for a pass/fail type assignment.

### Migrations
NO

1. Upload grades for a Pass/Fail type assignment via CSV
2. As a student, verify that the assignment appears as graded on the predictor page.

======================
Closes #3154

### DIFF
--- a/app/models/grade.rb
+++ b/app/models/grade.rb
@@ -22,8 +22,8 @@ class Grade < ActiveRecord::Base
     reject_if: proc { |a| (a["score"].blank?) }, allow_destroy: true
 
   before_validation :cache_associations
-  before_save :calculate_points
   before_save :zero_points_for_pass_fail
+  before_save :calculate_points
   after_save :check_unlockables
   after_save :update_earned_badges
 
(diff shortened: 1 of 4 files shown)


jonathoy:
Per our discussion, we will try to ensure that the grade score is set to `0` for now. Ideally, we may want to do some refactoring here since it appears that it may be more preferable to have the value set to `nil` instead.

PR #3164, opened 2017-03-29
enhanced sample data

acousticrobot:
### Status
**READY**

### Description

As I was working on analytics graphs I found a couple areas where I wasn't able to test that things were working properly with the current sample data. This PR adds the following enhancements to the sample data:

  * It adds a pass/fail assignment with grades
  * It simulates less than 100% participation on assignments with grades
  * It adds a "quick and dirty" algorithm to better simulate grades grouped around 75% of full points
  * It rounds grades to the nearest 100 points so that more students end up with the same grade
 
It also: 

  * Moves the assignments with submissions out of the Predictor type, and into the Submission type, since assignments with submissions weren't covered in there.
  * Removes the possibility to add grades with `instructor_modified: false`, which was only needed when predictions were housed on grades.
  * Adds the tag `[No Grades]` to more of the assignment type names to make finding grades assignments easier

### To Test

Run the sample data, but mostly I need confirmation that these are all changes that work for others' use cases.

### DIFF
--- a/db/samples.rb
+++ b/db/samples.rb
@@ -482,6 +482,10 @@ PaperTrail.whodunnit = nil
 
       if config[:grades]
         @students.each do |student|
+
+          # skip grades for some students, based on participation rate
+          next if config[:participation] && rand(100) > config[:participation]
+
           grade_attributes = config[:grade_attributes] || {}
           grade = student.grades.create! do |g|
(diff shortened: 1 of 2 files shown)


chcholman:
@acousticrobot This is such a fantastic update. Yes, it definitely works for a variety of use cases that we need to test regularly. Merging!

PR #3169, opened 2017-04-03
Test suite time

chcholman:
### Status
**READY**

### Description
This PR implements a consistent use of `let` in controller specs. It also switches multiple creates for builds at the course level, and removes the now unused `active_record_spec_helper`. 

### Todos
- [x] Add Tests

### Migrations
NO

* Test suite


spec/controllers/events_controller_spec.rb
```diff
@@ -1,32 +1,25 @@
 describe EventsController do
-  before(:all) { @course = create(:course) }
-  before(:each) do
-    session[:course_id] = @course.id
-    allow(Resque).to receive(:enqueue).and_return(true)
-  end
+  let(:course) { build(:course) }
```
acousticrobot: This is very minor, but a lot of the descriptions in this file still refer to `@event`: `"assigns a newly created event as @event"`...etc.

spec/controllers/events_controller_spec.rb
```diff
@@ -82,30 +75,24 @@
...
 
     describe "DELETE destroy" do
       it "destroys the requested event" do
-        expect{ get :destroy, params: { id: @event }}.to change(Event,:count).by(-1)
-      end
-
-      it "redirects to the events list" do
-        expect{ get :destroy, params: { id: @event }}.to change(Event,:count).by(-1)
+        @event = event
+        expect{ get :destroy, params: { id: @event } }.to change(Event,:count).by(-1)
```
acousticrobot: I don't know any performance reason to change this, and I don't even know which one reads better, but another way that I have used elsewhere to initialize a lazy-loaded model is like this:

```
      it "destroys the requested event" do
        event
        expect{ get :destroy, params: { id: event } }.to change(Event,:count).by(-1)
        ...
```

spec/controllers/events_controller_spec.rb
chcholman: Fixed!

spec/controllers/events_controller_spec.rb
chcholman: 👍 🎉

PR #3174, opened 2017-04-03
Allow Toggling of Canvas-Related UI Elements

### LINKED ISSUE #2973: Add admin-only setting to hide/display Canvas import buttons
We should add a setting at the course level to describe whether or not to show all Canvas related buttons/prompts to faculty who are not on the UM Ann Arbor campus. 

This task requires: 
- [ ] adding a boolean field to the `Course` model
- [ ] Adding the new field to the Course Settings form, right next to the active status 
- [ ] only displaying that new field to admin users
- [ ] reviewing the application to see where Canvas integration components are. Primarily it should be the Import assignments utility, the integrations setup panel on the Course form, and the import grades buttons

jonathoy:
### Status
READY

### Description
This PR adds an additional column called, `allows_canvas`, to the Course table. The intention is to allow an admin only to conditionally set whether Canvas-related elements should render for a specific course.

### Todos
- [x] Update Documentation

### Gem dependencies
N/A

### Migrations
YES, to add an additional non-nullable field `allows_canvas` to the Course table that defaults to true

- Ensure that the option to Import Grades from Canvas is not visible when `allows_canvas` is false
- Ensure that the option to Import Assignments from Canvas is not visible when ""
- Ensure that no Canvas references exist from the Integrations tab while editing a course

======================
Closes #2973

### DIFF
--- a/app/controllers/courses_controller.rb
+++ b/app/controllers/courses_controller.rb
@@ -114,7 +114,7 @@ class CoursesController < ApplicationController
 
   def course_params
     params.require(:course).permit :course_number, :name,
-      :semester, :year, :has_badges, :has_teams, 
+      :semester, :year, :has_badges, :has_teams,
       :team_term, :student_term, :section_leader_term, :group_term, :lti_uid,
       :user_id, :course_id, :course_rules, :syllabus,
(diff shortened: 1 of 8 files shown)


jonathoy:
@chcholman This should be fixed according to our discussed changes!

PR #3175, opened 2017-04-03
Update inactive course link helper

### LINKED ISSUE #3167: Button bar border-radius standardization
Make sure that button bar has curved radius on first and last button.

mariehooper:
### Status
**READY**

### Description
This PR updates the `active_course_link_to` method  to wrap the link in a `li` tag. This prevents empty list items from showing up in the views and causing unexpected styling or behavior.

### Migrations
NO

* Everywhere that `active_course_link_to` was being used and the content was empty. Mostly able to be seen in `button-bar` in inactive courses.

======================
Closes #3167

### DIFF
--- a/app/assets/stylesheets/pages/_assignments.sass
+++ b/app/assets/stylesheets/pages/_assignments.sass
@@ -23,6 +23,13 @@
   .button
     margin: 0.5rem 0
 
+.assignment-action-buttons
+  margin: 0
+  padding-left: 0
+  list-style-type: none
+  li
+    display: inline-block
+
 .add-unlock-condition
   margin-bottom: 1rem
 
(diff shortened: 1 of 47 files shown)


mariehooper:
Hey @jonathoy - I added the `else` to cover when `tag` is nil, but what I had overlooked was moving that up to the front means we need to add a nil before any of these that aren't using `tag` (and really two nils now since I needed a class on the `li` in one view so added a `tag_class` as well. Should we move them to the end of the helper to get around passing nil in as parameters everywhere? I figured we can chat about the best thing to do tomorrow. I won't push any more up for now.

jonathoy:
Looks great! Glad we were able to fix and clean this up!

Issue #3087, opened 2017-03-09
move all :json request paths to /api namespace


acousticrobot:
### Expected Behavior

Now that we have implemented an `api/` route namespace for the majority of our apps `json` endpoints, we should move the remaining ones there.

### Actual Behavior

The following controllers still accept json:

  * `/app/controllers/assignments_controller.rb:`
  * ~`/app/controllers/criteria_controller.rb:`~
  * `/app/controllers/events_controller.rb:`
  * `/app/controllers/grade_scheme_elements_controller.rb:`
  * `/app/controllers/grades_controller.rb:`
  * ~`/app/controllers/level_badges_controller.rb:`~
  * `/app/controllers/levels_controller.rb:`
  * `/app/controllers/observers_controller.rb:`
  * `/app/controllers/rubrics_controller.rb:`
  * `/app/controllers/staff_controller.rb:`
  * `/app/controllers/submissions_controller.rb:`
  * `/app/controllers/teams_controller.rb:`
  * `/app/controllers/users_controller.rb:`

I don't know how many of these endpoints are still being actively used by the app to serve json to jQuery components etc. 

I have found these in use:

`/javascripts/assignment_types.js:`
  * ~url: `/assignment_types/sort`~
  * ~url: `/assignments/sort`~

`/javascripts/earned_badges.js:`
  * ~url: `/badges/sort`~

acousticrobot:
'/assignment_types/sort' moved in #3181

acousticrobot:
`/assignments/sort` moved in #3187

acousticrobot:
A lot of these routes will be cleaned up as a by-product of #3182 (alerts refactor). I am going to put any work removing `respond_to do |format|` on hold until this branch is merged, as it would cause conflict headaches for @chcholman.

acousticrobot:
/badges/sort moved in #3215

acousticrobot:
See also: #3280

chcholman:
@acousticrobot Have we made any more progress on this?

acousticrobot:
This is waiting for #3182, see note above.

PR #3182, opened 2017-04-05
Alerts refactor

### LINKED ISSUE #2331: refactor alerts
currently using notices when we should be using successes

chcholman:
### Status
**In Progress/DO NOT MERGE**

### Description
This PR refactors the alerts used on the assignments controller leverage the [responders gem](https://github.com/plataformatec/responders) and allow for future internationalization. Due to the changes made at the layout level, we need to complete this across all views before merging. This first pass is a proof of concept to see what people think about the implementation. 

### Related PRs

 #3087 move all :json request paths to /api namespace

### Todos
- [x] Remove all layouts/alerts partial references from views except at the user layout level
- [x] Update all forms to reference layouts/errors
- [ ] Remove notices from all controller methods
- [ ] Add interpolation options to all controller methods
- [ ] Review Tests

### Migrations
NO

Try updating, copying, deleting, or creating an assignment

* All alerts throughout the app, so every place we submit a form

======================
Closes #2331


app/controllers/assignments_controller.rb
```diff
@@ -162,4 +150,8 @@ def assignment_params
       assignment_score_levels_attributes: [:id, :name, :points, :_destroy],
       assignment_groups_attributes: [:group_id]
   end
+
+  def flash_interpolation_options
```
acousticrobot: I had to go to the responder gem documentation to confirm that this bit of magic was doing something since we don't call the method ourselves -- maybe just a little comment above this method so it doesn't get accidentally removed down the line.

PR #3193, opened 2017-04-07
Misc ux updates

### LINKED ISSUE #3189: Add hint about point adjustment
### Expected Behavior
As an instructor, it should be clear to me that points adjustment can be both negative or positive point values

### Actual Behavior
Current design has no help text that explicitly indicates that point adjustment value can be negative as well as positive.

mariehooper:
### Status
**IN DEVELOPMENT**

### Description
This PR is an update to form styling throughout GradeCraft as well as a few UX updates from QA testing. 

screen shot 2017-04-07 at 4 00 25 pm

### Migrations
NO

* Course setup form layouts
* Form styling everywhere

======================
Closes #3189

### DIFF
--- a/app/assets/javascripts/angular/templates/grade_scheme_elements/element.html.haml
+++ b/app/assets/javascripts/angular/templates/grade_scheme_elements/element.html.haml
@@ -1,14 +1,14 @@
 .grade-scheme-row
   .gse-form-container
-    %label.gse-label.form_label
+    %label.gse-label.form-hint
       Grade Letter
       %input.letter{'type'=>'text', 'name'=>'letter', 'ng-model'=>'gradeSchemeElement.letter'}
 
-    %label.gse-label.form_label
+    %label.gse-label.form-hint
       Level name
       %input.level{'type'=>'text', 'name'=>'level', 'ng-model'=>'gradeSchemeElement.level'}
 
-    %label.gse-label.form_label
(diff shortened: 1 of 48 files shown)


mariehooper:
@chcholman - would love to get your initial thoughts on the updates to form styling and layout especially on course set up forms. If this looks good to you, I will continue through the other forms and implement columnar layouts where it makes sense.

chcholman:
@mariehooper It's looking awesome! Loving the columnar layout, it's really helping things feel shorter. Please continue!

chcholman:
@mariehooper The work is currently breaking a number of forms (see the grade scheme elements creation form and the rubric edit form for two examples). I think we probably need to create a list of all forms, and if we're going to do stylistic changes this significant we need to go through each one and confirm that they're in good shape. Otherwise, we could break this up into smaller PRs, but you won't be able to do as large scale a refactor to start with...

mariehooper:
@chcholman Will make a list and keep working! Thanks for reviewing!

chcholman:
@mariehooper This is looking great. I've checked all the forms that I can think of and am going to go ahead and merge. I'm also going to alert Evan to this being a work in progress, and ask her to follow up with you directly as she sees spots that needs some refinement. One that I'd like us to attack sooner rather than later: getting the column format applied to the Assignment form.

PR #3199, opened 2017-04-11
Reroute Delete Staff Member to Staff Page

### LINKED ISSUE #2806: Route users back to the Staff Index page after removing a staff member
### Expected Behavior
As an instructor, if I have deleted a staff member from my list, I expect to stay on the Staff Index page.

### Actual Behavior
When I delete a staff member from my list, I am rerouted to the User Index page which is a very long list.

### Steps to reproduce
[Trello Card](https://trello.com/c/aiYJsENn/70-take-users-back-to-the-staff-index-page-after-removing-a-staff-member)

dovek:
### Status
**IN DEVELOPMENT**

### Description
As an instructor, if I have deleted a staff member from my list, I expect to stay on the Staff Index page.

### Related PRs
N/A

### Deploy Notes
N/A

### Gem dependencies
N/A

### Migrations
NO

As an Admin, navigate to the staff page and delete a staff member using the options cog. Deleting said staff member should remove the staff member from the staff roster as well as navigate the user back to the staff page.

* Staff page

======================
Closes #2806

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -9,7 +9,7 @@ class UsersController < ApplicationController
 
   before_action :ensure_staff?,
     except: [:activate, :activated, :edit_profile, :update_profile]
-  before_action :save_referer, only: [:manually_activate, :resend_invite_email]
+  before_action :save_referer, only: [:manually_activate, :resend_invite_email, :destroy]
   skip_before_action :require_login, only: [:activate, :activated]
   skip_before_action :require_course_membership, only: [:activate, :activated]
 
@@ -83,7 +83,7 @@ class UsersController < ApplicationController
 
     respond_to do |format|
       format.html do
-        redirect_to users_url,
+        redirect_to session[:return_to] || users_url,
         notice: "#{@name} was successfully deleted"
       end
     end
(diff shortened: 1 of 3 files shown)


app/controllers/users_controller.rb
```diff
@@ -83,7 +83,7 @@ def destroy
 
     respond_to do |format|
       format.html do
-        redirect_to users_url,
+        redirect_to session[:return_to] || users_url,
```
jonathoy: This seems like it should work great, but I wonder if it may be better to just hard-code a `redirect_to staff_index` in case there are other ways of deleting a user other than from the staff page.

chcholman:
@dovek How are you testing this? When I do this, I'm getting an error currently (`uninitialized constant CourseMembershipsController::Services`)

dovek:
@chcholman I did receive this error once while I was doing functional testing. When I stopped and restarted the server the error went away.

chcholman:
@dovek Awesome, restarting the server allowed me to get past that error. However, when I test this by navigating to the staff page, and delete a course membership, I am still redirected to the students index page. Looking at the code, it looks like you've edited the `UsersController`, but I believe we need to edit the `CourseMembershipsController` for this to work. Will you check it out when you get a chance?

dovek:
@chcholman Looking at this now.

PR #3205, opened 2017-04-12
3163 page canvas results

### LINKED ISSUE #3163: Page Canvas Grade Results
As an instructor
When I import grades from Canvas
Then I should get paged results
So I get results faster and I do not receive any timeouts

jonathoy:
### Status
READY

### Description

To address a timeout issue that we've been having when loading Canvas grades for import, we need to revise our current integration so that it fetches paginated results. By doing so, we can asynchronously load batches of grades all on a single page.

With these changes, the default grades per page are set to 25.

loading_grades

### Todos
- [ ] Update Documentation

### Deploy Notes
N/A

### Gem dependencies
N/A

### Migrations
NO

1. Go to a specific assignment and select from the Options menu to Import Grades from Canvas. Select an assignment from Canvas and click to import.

Note: The Canvas grade API (route name: `api_assignment_grades_importer_grades`) takes a `per_page` parameter that sets the batch size. To test the effect of varied batch sizes from the UI, add a `per_page` attribute to the `_gradeParams` literal in `GradeImporterService.coffee`.

Canvas grade import

======================
Closes #3163


jonathoy:
@jwright The front-end work is still a bit of a WIP but if you get the chance to look over the changes I made to your Canvas logic that would be great. I struggled with a neat way of returning the grades with a boolean indicator for indicating the existence of additional pages so if you have a better idea I'm all ears! You can see my approach in `canvas_syllabus.rb`.

I tried to maintain reusability here as much as possible since it seems as though you put a good deal of thought into the current implementation.

jonathoy:
Also, this one is for both @jwright and @acousticrobot - This may not be possible to replicate locally if you don't have the setup configured for the Canvas development environment but I'll try to give a high-level overview of the issue.

My page loads the grades for preview from Canvas so that the instructor can select the ones they want to import. Previously, upon submitting the user would be directed to a page that displays the results of the import. I wanted to retain this workflow by having my Angular form POST to `Grades::ImportersController#grade_import`, as it did before. However, coming from an Angular form I find that I am getting a CSRF error (Can't verify CSRF token authenticity.).

I'm still looking into this, but it seems as though our current gem for handling the token validation (ng-rails-csrf) does not work for this case. Any thoughts?

app/controllers/api/grades/importers_controller.rb
```diff
@@ -0,0 +1,29 @@
+class API::Grades::ImportersController < ApplicationController
+  include OAuthProvider
+
+  oauth_provider_param :importer_provider_id
+
+  before_action :ensure_staff?
+  before_action :require_authorization
```
jwright: This does a 302, redirect if the user is not authorized. I this going to work on the ajax side? Have you tested this?

app/controllers/grades/importers_controller.rb
```diff
@@ -37,11 +37,11 @@ def download
   def grades
     @assignment = Assignment.find params[:assignment_id]
     @provider_name = params[:importer_provider_id]
-    @provider_assignment = syllabus.assignment(params[:id], params[:assignment_ids])
-    @grades = syllabus.grades(params[:id], [params[:assignment_ids]].flatten)
```
jwright: Why were these instance variables removed?

lib/active_lms/syllabus/canvas_syllabus.rb
```diff
@@ -335,13 +339,13 @@ def assignment(course_id, assignment_id)
...
     # }]
-    def grades(course_id, assignment_ids, grade_ids=nil)
+    def grades(course_id, assignment_ids, grade_ids=nil, fetch_next=false, options={})
       grades = []
       params = { assignment_ids: assignment_ids,
                  student_ids: "all",
                  include: ["assignment", "course", "user"],
-                 per_page: 100 }
-      client.get_data("/courses/#{course_id}/students/submissions", params) do |data|
+                 per_page: options[:per_page] || 25 }.merge(options.except(:per_page))
```
jwright: This can be simplified with 

`per_page: options.delete(:per_page) || 25 }.merge(options)`

lib/canvas/api.rb
```diff
@@ -23,8 +26,11 @@ def get_data(path="/", params={})
         response = self.class.get(next_url, query: { access_token: access_token })
         raise ResponseError.new(response) unless response.success?
         yield response.parsed_response if block_given?
-        next_url = get_next_url response
+        next_url = get_next_url(response)
```
jwright: In order to eliminate the complexity of the branching logic, I would change this to:

```
loop do
  # ...
  break if !fetch_next || next_url.nil?
end
```

app/controllers/grades/importers_controller.rb
jonathoy: I no longer needed this data returned on page load since I'm now fetching this information via AJAX.

lib/active_lms/syllabus/canvas_syllabus.rb
jonathoy: 👍

lib/canvas/api.rb
jonathoy: Cool, I am in agreement

app/controllers/api/grades/importers_controller.rb
jonathoy: The promise resolves itself as a failure, but I will go ahead and make the overall error handling a bit more robust by adding some sort of message on the screen for the user.

jonathoy: I've changed the logic so that by setting the `redirect_path` here and having the `ExternalAuthorization` class respond according to the request type, I can have the Angular side of things redirect to `/auth/canvas` and have it follow the same pattern as it normally would for authenticating.

jonathoy: I've changed the logic so that by setting the `redirect_path` in the API endpoint and having the `ExternalAuthorization` class respond according to the request type, I can have the Angular side of things redirect to `/auth/canvas` and have it follow the same pattern as it normally would for authenticating.

jonathoy:
@chcholman Zhen gave us some new keys so if we can verify that they do solve our problem that would be awesome! I would be happy to make the config change if you are able to go over the necessary steps with me.

app/assets/javascripts/angular/templates/grades/import/form.html.haml
```diff
@@ -0,0 +1,88 @@
...
+%loading{"ng-if"=>"vm.loading"}
+  %h2
+    %i.fa.fa-spinner.fa-spin.fa-fw
+    Loading {{vm.provider}} grades..
+  %br
+
+%loaded{"ng-if"=>"!vm.loading"}
+  %without_grades{"ng-if"=>"!grades || grades.length < 1"}
+    %div There are no grades to import for these {{termFor("assignment")}}s.
+    %div
```
chcholman: Super picky language note: you're only ever importing grades for one assignment at a time, so let's make this 'this' rather than 'these', and can you downcase the term for assignment?

jonathoy: Yep no problem! I pretty much carried this stuff over verbatim from the old view.

jonathoy:
So I figured out what's going on with Tawfiq, and it turns out that the Canvas API will still return the submission if a grade was previously assigned to it but then is cleared out at a later point. For all intents and purposes, Canvas treats the assignment as being graded from then on.

I'm sure this was happening before as well but the behavior was probably unnoticed until now, since there is no explicit logic on our side to filter out these null grades. I can implement this in a separate PR if it's still desired.

chcholman:
@jonathoy Ah, that makes total sense. Yes, let's address that, but in a separate issue. Creating that now, and merging this!

PR #3206, opened 2017-04-12
Fix Edit User With Blank Password Bug

### LINKED ISSUE #3201: Editing a user without a password causes error
### Expected Behavior
After creating a brand new user and activating the user, you should be allowed to edit the user without password confirmation.

### Actual Behavior
After creating a brand new user and activating the user, you must set the new user's password before editing their account. Otherwise, the admin editing the account receives a "Password Confirmation can't be blank" error.

### Steps to reproduce
Create a new user. Edit the new user's account (change the user's role). Click "Update User" button. Should receive error message: 
1 error prohibited this user from being saved:
* Password confirmation can't be blank

dovek:
### Status
**IN DEVELOPMENT**

### Description
After creating a brand new user, operator should be allowed to edit the user without password confirmation, that is, user does not receive the "Password Confirmation can't be blank" error message when password fields are left blank.

### Related PRs
N/A

### Todos
- [ ] Add Tests

### Deploy Notes
N/A

### Gem dependencies
N/A

### Migrations
NO

Navigate to the staff page. Create a new user. Edit this user by making the user an instructor/student/etc... for the current course while leaving the password and password confirmation fields blank. Should receive message that the user was successfully updated.

* Users

======================
Closes #3201

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -59,7 +59,12 @@ class UsersController < ApplicationController
 
   def update
     @user = User.find(params[:id])
-    @user.assign_attributes user_params
+    up = user_params
+    if up[:password].blank? && up[:password_confirmation].blank?
+      up.delete(:password)
+      up.delete(:password_confirmation)
+    end
+    @user.assign_attributes up
     cancel_course_memberships @user
     if @user.save
       if @user.is_student?(current_course)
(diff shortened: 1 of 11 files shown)


chcholman:
@dovek I added tests to confirm that 1) admin users can edit a password and 2) the user can update other fields without needing to set a password. Can you confirm that this is ready to be merged?

PR #3222, opened 2017-04-18
3083 responsive charts

### LINKED ISSUE #3083: Add responsive containers to new analytics

mariehooper:
### Status
**READY**

### Description
This PR adds functions to resize Plotly.js charts on window resize end for both Angular and jQuery. jQuery functions will go away when those remaining charts are Angularized.

### Todos
- [x] DRY up repeated functions in directives
- [x] fix styling for initial load of assignment analytics page

### Migrations
NO

* Assignment analytics, analytics on dashboards

======================
Closes #3083

### DIFF
--- a/app/assets/javascripts/angular/directives/analytics/assignment_distribution.coffee
+++ b/app/assets/javascripts/angular/directives/analytics/assignment_distribution.coffee
@@ -67,15 +67,17 @@
           ay: -40
         }]
 
-      plotAssignmentDistro = ->
-        Plotly.newPlot('assignment-distribution-graph', data, layout, {displayModeBar: false})
+      Plotly.newPlot('assignment-distribution-graph', data, layout, {displayModeBar: false})
 
-      plotAssignmentDistro()
-
-      resizeTimer = undefined
       angular.element($window).on 'resize', ->
+        resizeTimer = undefined
+        assignmentDistributionGraph = document.getElementById('assignment-distribution-graph')
+
         clearTimeout resizeTimer
-        resizeTimer = setTimeout(plotAssignmentDistro, 250)
+        resizeTimer = setTimeout((->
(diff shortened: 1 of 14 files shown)


mariehooper:
Hi @acousticrobot - would love to get some feedback about how I could make this code more DRY. Particularly the Angular piece since that code will stay for the window resizing and the jQuery will be deprecated when the charts are angularized. Thank you!

mariehooper:
Hey @acousticrobot - I am unable to get the overlapping on initial draw to resolve itself. I thought it was a flexbox issue, but after some experimentation, I realized I am only seeing the issue inside jQuery UI tabs on the assignment page - which makes me wonder (1) if the angular loads before the jQuery UI, then its possible that the container divs don't exist when the charts are drawn and (2) this might not be an issue in production when there is caching. Does this sound reasonable? And if so, what would you recommend? 

@jonathoy - Jonathan and I are hoping you might take a look at this PR and see if you have any ideas about how to refactor the debounce to take the time as an option.  From Jonathan: I don’t see any issues with performance, only that the argument list for debounce is getting a little silly and I wonder if it could be simplified.

mariehooper:
@acousticrobot - After some digging, I figured out that it's not a race condition between Angular and jQuery, it's an issue with the charts being on a non-default tab. Since they start hiding on a tab, the containers holding the charts have no height or width. We need to draw the assignment charts for the first time on tab activation, rather than page load (`assignmentDistributionAnalytics` and `assignmentScoresEarnedAnalytics`). 

There are a couple of hacky ways I can think of to do this like using a function like this: 
```
      angular.element('.class-analytics-tab').on 'click', ->
        refreshGraph()
```
I think drawing on tab activation would be preferable though, because there would be no initial jump on the page.

mariehooper:
@acousticrobot - Should be good to go now!

PR #3230, opened 2017-04-20
User lastname

acousticrobot:
### Status
**READY, PENDING REVIEW OF COMMENT ON LINE 257**

### Description
  * This PR adds a unique last name to the user to avoid intermittent failures
  * It also isolates the specs for teams and groups, reducing the runtime for `user_spec.rb` by about a minute and a half

Cait, can you look at the comment on `it "returns 0 if the student has no score"...` and let me know what you want me to do?  I think this spec dates back to before we added non-null constraints on the db and probably could be removed, but we may want to take a look at `score_for_course(course)` and refactor this as well.

### DIFF
--- a/spec/models/user_spec.rb
+++ b/spec/models/user_spec.rb
@@ -1,15 +1,13 @@
 describe User do
   let(:course) { build(:course) }
   let(:student) { create(:user, username: "simple", last_name: "Oneofakind") }
-  let(:student_not_being_graded) { create(:user) }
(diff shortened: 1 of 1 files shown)


chcholman:
@acousticrobot It seems like we should both refactor the `score_for_course(course)` method, and can probably remove some of the tests here as well. We both added non-null contraints, and started storing students total values (previously we'd summed them on the fly), which I think is the legacy that caused the pieces we're seeing...

PR #3233, opened 2017-04-24
3186 validating dates

### LINKED ISSUE #3186: Update assignment due_dates via single page settings
As an instructor, I want to be able to rapidly update the due dates for assignments by editing them from directly from the settings page.

acousticrobot:
### Status
**READY**

### Description

Adds date fields to the assignments settings page, and validation.  If the dates are not valid, error explanation is presented and the API will not go through until they are resolved.

screen shot 2017-04-24 at 1 55 42 pm

Relies on a javascript date parsing library to format our date string format ("April 13, 2017 - 10:10am") back into a date in order to perform validations. I am open to other solutions for this.

Will not reliably handle all types of user error from manually entering date string (ex. "April 33, 2017"). Safe-guarding against this would be possible but non-trivial. If this is a requirement we should talk through the implementation and probably create a separate ticket.

url: `/assignments/settings`

### Related PRs

All Assignments settings work

### Gem dependencies

No gem dependencies. See comment above about `date.js`

### Migrations

NO

======================
Closes #3186


chcholman:
That looks great to me @acousticrobot! Merge when ready, and we'll ask @mariehooper to take a look at polishing the styles in the future!

PR #3234, opened 2017-04-24
Import Canvas Users

### LINKED ISSUE #2934: Import students from Canvas
As an instructor, I would like to be able to import the users currently associated with my Canvas site into my GradeCraft site, with the correct roles

A good pattern for this is importing assignments from the Canvas site

jonathoy:
### Status
READY

### Description
We currently can import various things such as grades and assignments from Canvas, but we cannot import only the users for a course. This PR adds this ability.

The user import process is structured in a way such that it can import all users for a particular course and not just the students. Since Canvas allows for the possibility of having multiple roles for a course, for our purposes only the most "important" role will be created. In order of precedence, the Canvas roles are as follows:

`TeacherEnrollment > DesignerEnrollment > TaEnrollment > StudentEnrollment > ObserverEnrollment`

Note that at this time, the designer role is mapped to a GSI role in Gradecraft.

### Todos
- [x] Add Documentation (Documented in #3263)

### Gem dependencies
N/A

### Migrations
NO

Head to the Users page where you will now notice the option to Import Users in the upper right-hand corner. Click and select to import users from Canvas.

Ensure that all users load via pagination and that upon submitting the form, the users are created successfully and their course roles properly assigned.

Also impacts Canvas grade import, since refactoring was required in order to reuse some existing user import logic.

======================
Closes #2934


jonathoy:
When I went back to make some changes I noticed that there's a lot of duplicated code that is pretty much already taken care of in `Services::CreatesOrUpdatesUser`. I tried to refactor out the `find_or_create_user` method in `CanvasUserImporter`, but as I had mentioned there's some issues with failures not being propagated to the parent contexts when the user creation fails in the `SavesUser` action. I would like to revisit this in a separate PR since it is not the first time this has become an issue.

Please take a look at the placement and wording of the tooltip in the situation where a user's role has changed in Canvas and let me know how and if you want me to make changes.

role_changed

chcholman:
@jonathoy Will you address the merge conflicts when you get a chance?

chcholman:
@jonathoy Thanks so much for dealing with the merge conflicts! This is looking great. I'm wondering about doing the following to make what's going on with the roles more apparent: 

- [ ] Rename the role column right now to "[provider name] role"
- [ ] Add a column next to it "GradeCraft role", and fill this in with the role they'll be assigned to in GC

Given that we don't perfectly match roles between the systems, and that it's not always clear what someone's current role is, this feels like it will help make things more transparent. 

In testing, I'm seeing the results page report users who didn't appear to have an account at all (according to the first import screen) show up under the "Unchanged" heading (perhaps because they do indeed have GC accounts, just not course memberships for this course). 

- [ ] If a user exists and the import adds them to the course, they should show as successfully imported. 
- [ ] Let's remove the username column from the final page, and instead show "Site Role" and their permissions for this course

PR #3236, opened 2017-04-26
3185 setting full points

### LINKED ISSUE #3185: Update assignment full_points via single page settings
As an instructor, I want to be able to rapidly update the point values for assignments by editing them from directly from the settings page.

acousticrobot:
### Status
**READY**

### Description

Adds an input field for `full_points` on the assignments settings page.
Renders term for pass/fail instead if assignment is pass/fail (non-interactive)
renames smart-number `gc-number-input` in line with date time (`gc-date-time-input`).

### Related PRs

Assignment Settings Branches

======================
Closes #3185


chcholman:
Looks great! Thanks @acousticrobot!

PR #3242, opened 2017-05-01
Filter out scoreless grades returned from Canvas

### LINKED ISSUE #3239: Filter out null grades from Canvas import
If a grade available to import from Canvas has no score and no feedback, we should not import the null object.

jonathoy:
### Status
READY

### Description
When importing grades from Canvas, it is possible to get submissions where the grade is scoreless. This happens in the scenario where a student has been graded on an assignment, but their grade is then cleared out later on. The Canvas API returns these types of submissions anyway since the grade status is considered to be "Graded".

For our case, we should omit these from the import.

### Gem dependencies
N/A

### Migrations
NO

Grade a student in Canvas, allowing the grade to persist. Then, clear their grade out and attempt to import grades in Gradecraft for that particular assignment. The scoreless grade should not appear in the selection list.

Canvas grade import

======================
Closes #3239

### DIFF
--- a/spec/helpers/canvas_api_helper_spec.rb
+++ b/spec/helpers/canvas_api_helper_spec.rb
@@ -1,0 +1,15 @@
+describe CanvasAPIHelper, focus: true do
+  subject { helper }
+
+  describe "#concat_submission_comments" do
+    it "returns nil if there are no comments" do
+      expect(subject.concat_submission_comments([])).to be_nil
+    end
+
+    it "returns comments parsed and separated by the default separator" do
+      comments = [{ "comment" => "good jorb" }, { "comment" => "excellent" }]
+      expect(subject.concat_submission_comments(comments)).to eq \
+        "Comment 1: good jorb; Comment 2: excellent"
+    end
+  end
+end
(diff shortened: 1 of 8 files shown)


chcholman:
@jonathoy There are use cases where instructors share intermediate feedback with students, but no scores. Can we refine the filter slightly to import grades that have feedback, but no scores?

jonathoy:
I've included `submission_comments` to the associations that are returned by the API. As discussed, if there are multiple comments for import they will be concatenated together.

spec/helpers/canvas_api_helper_spec.rb
```diff
@@ -0,0 +1,15 @@
+describe CanvasAPIHelper, focus: true do
+  subject { helper }
```
chcholman: Focus flag needs to be removed

chcholman:
@jonathoy I'm still seeing null objects? 

screen shot 2017-05-04 at 4 20 52 pm

jonathoy:
@chcholman I removed the `focus: true` flag but I'm not able to replicate what you're seeing. I generated a temporary access token for the production Canvas server so that I could pull from the same course, but when I set up the configuration I'm only seeing Tawfiq.

tawfiq

Any chance that you could double-check that you are able to replicate this?

chcholman:
Interesting, I now can't replicate this. Merging and will keep an eye out for it!

PR #3246, opened 2017-05-03
3240 Add individual assignments to google calendar

### LINKED ISSUE #3240: Add "Add to Google Calendar" to Individual Assignments
When a student or instructor looks at an individual assignment, they should be able to click an "Add to Google Calendar" button that works exactly as Events do. If there is no start time for the assignment (only a due at datetime), then set it to be 30 minutes prior to the due time.

dovek:
### Status
**READY FOR REVIEW**

### Description
When a student/instructor/observer looks at an individual assignment, they should be able to click an "Add to Google Calendar" button that works exactly as Events do. If there is no start time for the assignment (only a due at datetime), then set it to be 30 minutes prior to the due time.

### Related PRs
N/A

### Todos
- [x] Add Tests

### Deploy Notes
N/A

### Gem dependencies
N/A

### Migrations
NO

As a student/observer, go to the assignments menu and click any assignment. In the Description tab, the user should notice that the "Add to Google Calendar" button. Upon clicking the button, the assignment will be copied to the user's corresponding Google Calendar if the user has the User Authorization linking a Google Account to their GradeCraft account.

As an instructor, go to the assignments menu and click any assignment. In the top right corner of the screen the user should notice the "Add to Google Calendar" button. Upon clicking the button, the assignment will be copied to the user's corresponding Google Calendar if the user has the User Authorization linking a Google Account to their GradeCraft account.

* Assignments

======================
Closes #3240

### DIFF
--- a/app/controllers/google_calendars_controller.rb
+++ b/app/controllers/google_calendars_controller.rb
@@ -21,7 +21,7 @@ class GoogleCalendarsController < ApplicationController
     google_authorization = get_google_authorization(current_user)
     item = get_event_or_assignment(current_course, params[:class], params[:id])
     if item.due_at.nil?
-      redirect_to get_path(params[:class]), alert: "Google Calendar requires " + params[:class] + " have at least END time!" and return
+      redirect_to item, alert: "Google Calendar requires " + params[:class] + " have at least END time!" and return
     end
     begin
       google_event = create_google_event(item)
--- a/app/controllers/google_calendars_controller.rb
+++ b/app/controllers/google_calendars_controller.rb
@@ -31,9 +31,9 @@ class GoogleCalendarsController < ApplicationController
       calendar.authorization = secrets.to_authorization
       calendar.authorization.refresh!
       result = calendar.insert_event('primary', google_event)
-      redirect_to get_path(params[:class]), notice: params[:class].capitalize + " " + item.name + " successfully added to your Google Calendar"
+      redirect_to item, notice: params[:class].capitalize + " " + item.name + " successfully added to your Google Calendar"
     rescue Google::Apis::ServerError, Google::Apis::ClientError, Google::Apis::AuthorizationError, Signet::AuthorizationError
-      redirect_to get_path(params[:class]), alert: "Google Calendar encountered an Error. Your " + params[:class] + " was NOT copied to your Google calendar."
+      redirect_to item, alert: "Google Calendar encountered an Error. Your " + params[:class] + " was NOT copied to your Google calendar."
     end
   end
 
--- a/app/controllers/google_calendars_controller.rb
+++ b/app/controllers/google_calendars_controller.rb
@@ -45,17 +45,21 @@ class GoogleCalendarsController < ApplicationController
     end
     google_authorization = get_google_authorization(current_user)
     assignment = current_course.assignments.find(params[:id])
-    begin
-      google_event = create_google_event_from_assignment(assignment)
-      calendar = Calendar::CalendarService.new
-      client_id = Google::Auth::ClientId.new(ENV['GOOGLE_CLIENT_ID'], ENV['GOOGLE_SECRET'])
-      secrets = create_google_secrets(google_authorization)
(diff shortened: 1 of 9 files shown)


app/controllers/google_calendars_controller.rb
```diff
@@ -38,4 +38,29 @@ def add_event_to_google_calendar
     end
   end
 
+  def add_assignment_to_google_calendar
+    if !google_auth_present?(current_user)
```
chcholman: This feels like it could be DRYed up by creating a single `add_to_google_calendar` method that combines the procedural components of both `add_event` and `add_assignment` and then deals with specific object-based differences as necessary

app/controllers/google_calendars_controller.rb
```diff
@@ -38,4 +38,29 @@ def add_event_to_google_calendar
...
+      # rubocop:disable AndOr
+      redirect_to "/auth/google_oauth2" and return
+    end
+    google_authorization = get_google_authorization(current_user)
+    assignment = current_course.assignments.find(params[:id])
+    if assignment.due_at.nil?
+      redirect_to assignments_path, alert: "Google Calendar requires assignment have at least END time!"
```
chcholman: Is there any way that we can prompt the user to set the event times (in Google Calendar) if the GradeCraft version doesn't have one? I'm imagining that if you clicked the "Add to Calendar" button and there was no date: 
1) you'd get an alert in GradeCraft that the event had no related dates, would you like to set your own in Google Calendar and then
2) You'd get to the event create/edit page with the description and title in place but no dates, and you could manage the timing yourself

Thoughts?

app/helpers/google_calendars_helper.rb
```diff
@@ -33,6 +33,28 @@ def create_google_event(event)
     google_event
   end
 
+  def create_google_event_from_assignment(assignment)
+    google_event = Calendar::Event.new({
+      summary: assignment.name,
```
chcholman: This seems like it could also be combined with the event method. I'd like both to also: 

1) assign the description (right now the assignment does but the event does not)
2) set the start time to - 30 minutes if there is none (happening on assignment, and not on event)

spec/helpers/google_calendars_helper_spec.rb
```diff
@@ -1,8 +1,19 @@
...
+
+  class Helper
+    include GoogleCalendarsHelper
+
+    def current_course
+    end
+  end
+
+  subject(:helper) { Helper.new }
+  before {allow(helper).to receive(:current_course).and_return course}
```
dovek: @chcholman So here is where I tried to set up the allow.to receive... And for whatever reason, I'm still receiving the error:

'''
  1) GoogleCalendarsHelper#get_event_or_assignment returns the event of the corresponding event object
     Failure/Error: return current_course.events.find(id) if class_name == "event"
     
     NameError:
       undefined local variable or method `current_course' for #<RSpec::ExampleGroups::GoogleCalendarsHelper::GetEventOrAssignment:0x007fa74fafd040>
'''

spec/helpers/google_calendars_helper_spec.rb
```diff
@@ -13,6 +24,23 @@
...
   end
 
+  describe "#get_event_or_assignment" do
+    let(:user) { create :user }
+    let(:course) { build(:course) }
+    let(:event) {create(:event, course: course)}
+    let(:assignment_type) { create(:assignment_type, course: course) }
+    let(:assignment) { create(:assignment, assignment_type: assignment_type, course: course) }
+    it "returns the event of the corresponding event object" do
+      # allow(helper).to receive(:current_course).and_return course
```
dovek: I tried it here as well. Both returned the undefined local... error

spec/helpers/google_calendars_helper_spec.rb
dovek: @chcholman Jonathon suggested that I pass in current_course from the controller to the helper in order to avoid this. I'm going to try that next.

PR #3252, opened 2017-05-08
3190 badge dimensions

### LINKED ISSUE #3190: Ensure appropriate badge dimensions and maintained aspect ratio
### Expected Behavior
As an instructor, I would expect it to be clear what size images used for badges need to be.

### Actual Behavior
Currently there are no parameters for badge images and we don't (in most cases) account for the fact that they may not be round/square, so we need to maintain aspect ratio for sizing and add helper text for minimum size/pixel density.

### [Trello Card](https://trello.com/c/hHjKSoDt/96-implement-helper-text-specifying-badge-icon-dimensions)

to-do:
- [x] update badge icon component to be just badge icon instead of link, apply link where necessary (student dashboard)
- [ ] update styling for selected badge on student badge show page assuming that not all badges will have transparent background
- [x] add hint about badge dimension on creation form (150px is currently the largest version shown in UI)
- [x] review aspect ratio through app, determine if worth adding a clipping container

mariehooper:
### Status
**READY**

### Description
This PR adds some explanatory text to badge creation page for instructors along with a preview of how their uploaded images will look in GradeCraft. Also included in this PR is a crop-simulation container for all badges in the app to maintain aspect ratio.

### Migrations
NO

* badges

======================
Closes #3190

### DIFF
--- a/app/assets/javascripts/angular/templates/badges/main.html.haml
+++ b/app/assets/javascripts/angular/templates/badges/main.html.haml
@@ -8,7 +8,8 @@
     .badge-index-section
       .badge-icon-container{'ng-repeat'=>'badge in badges', 'ng-class'=>'{"focus-article": isFocusArticle(badge)}'}
         %a.mobile-click{:href => "badges/{{badge.id}}"}
-        %img{'ng-class'=>"(badge.earned_badge_count <= 0 ? 'badge-icon unearned' : 'badge-icon')", 'ng-src'=>'{{badge.icon}}', 'ng-click'=>'changeFocusArticle(badge)'}
+        .img-cropper.med-crop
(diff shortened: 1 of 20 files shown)


chcholman:
@mariehooper This is looking great, I really like the preview in general. Questions: 

- [ ] You can't really make use of the preview in the right way, because the image uploads when you submit the form. It feels like we should consider updating this to be dynamic in a different way - thoughts? 
- [ ] Will you take a look at all the places where we're using the badge and see if you think they're the right size/have the right spacing around them. Two that caught my eye: 

1) Abutting the text on the show page
screen shot 2017-05-09 at 8 33 18 am

2) Maybe *too* tiny on the earned badge show page? 

screen shot 2017-05-09 at 8 33 47 am

mariehooper:
@chcholman Thanks for the feedback! See if you like my solution for dynamic image previewing... and maybe try disabling your cache again to see if the styles will be applied in Chrome?

PR #3282, opened 2017-05-18
Fix breadcrumb-related issues

### LINKED ISSUE #3281: Canvas-Related Breadcrumbs Broken
Some of the Canvas-related routes are broken due to an error that is occurring in the breadcrumbs. The breadcrumbs expect a Course model to be set, but in the context of our Canvas controller methods the course is actually the Canvas course which is hash.

jonathoy:
### Status
READY

### Description
Some of the Canvas-related routes are broken due to an error that is occurring in the breadcrumbs. The breadcrumbs expect a Course model to be set, but in the context of our Canvas controller methods the course is actually the Canvas course which is hash.

### Migrations
NO

Ensure that Canvas import routes are accessible and that there are no breadcrumb errors.

Canvas user, assignment, and grade import

======================
Closes #3281

### DIFF
--- a/app/controllers/assignments/importers_controller.rb
+++ b/app/controllers/assignments/importers_controller.rb
@@ -19,9 +19,10 @@ class Assignments::ImportersController < ApplicationController
   # GET /assignments/importers/:importer_provider_id/courses/:id/assignments
   def assignments
     @provider_name = params[:importer_provider_id]
-    @course = syllabus.course(params[:id])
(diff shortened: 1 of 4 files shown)


chcholman:
@jonathoy I made one tiny update to remove the assigned `@course` in each method, and replace it with the before_action call @acousticrobot created that sets this. Let me know if you have any questions!

PR #3291, opened 2017-05-22
Set last_login_date When Authenticating Via LTI

### LINKED ISSUE #3269: Update LTI authentication so that it updates last login date for course

jonathoy:
### Status
READY

### Description
If the user is authenticating via LTI, the `last_login_date` should be updated on the course membership.

This PR also tidies up the associated specs and reduces the number of database transactions performed.

### Gem dependencies
N/A

### Migrations
NO

Ensure that it is still possible to authenticate to Gradecraft via LTI and that the user's `last_login_date` is updated accordingly.

Authentication via LTI

======================
Closes #3269

### DIFF
--- a/app/models/course_membership.rb
+++ b/app/models/course_membership.rb
@@ -27,21 +27,22 @@ class CourseMembership < ActiveRecord::Base
 
   def self.create_or_update_from_lti(student, course, auth_hash)
     return false unless auth_hash["extra"] && auth_hash["extra"]["raw_info"] && auth_hash["extra"]["raw_info"]["roles"]
-    course_membership = student.course_memberships.find_or_create_by(course_id: course.id)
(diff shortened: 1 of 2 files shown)


jonathoy:
@chcholman While I was looking over this code, I found a possible edge case that we may want to verify is correct.

If you look at `course_membership.rb` line 42 in a condition where the user already has an existing course membership, we see that the `instructor_of_record` value is being set to false if the role being returned from LTI is something other than instructor, teachingassistant, or learner.

Let's say that the user's role in GC was last set to GSI and now it's something that falls into the else block. Is it still safe to assume that the `instructor_of_record` value should be false, and if so should their role also be updated to something else like that of an observer?

chcholman:
@jonathoy That makes a lot of sense - if they fall into the else block, their role should be set to `observer`. This otherwise looks good to go!

PR #3293, opened 2017-05-23
3253 Add all asignments google calendar refactor

### LINKED ISSUE #3253: Add all course assignments to the calendar at once
### Expected Behavior
As a user, I want a quick way to add all assignments to my Google Calendar at once. This should be doable from the assignment index page.

dovek:
### Status
**IN DEVELOPMENT**

### Description
As a user, I want a quick way to add all assignments to my Google Calendar at once. This should be doable from the assignment index page.

### Related PRs
N/A

### Todos
- [x] Add Tests
- [x] Add Invisible Assignment Logic

### Deploy Notes
N/A

### Gem dependencies
N/A

### Migrations
NO

As a student with a google authorization record in the database, navigate to the assignments page and click the "Add All Assignments to Google Calendar" button and all assignments will be moved to the user's google calendar. As a professor with a google authorization record in the database, navigate to the assignments page and click the "Add All Assignments to Google Calendar" button and all assignments will be moved to the professor's google calendar.
 
* Assignments 

======================
Closes #3253


app/controllers/google_calendars/assignments_controller.rb
```diff
@@ -0,0 +1,36 @@
...
+  before_action do |controller|
+    controller.redirect_path request.referer
+  end
+
+  before_action :redirect_if_auth_not_present
+
+  Calendar = Google::Apis::CalendarV3
+
+  def add_assignment
+    assignment = get_item(current_course, "assignment", params[:id])
```
jonathoy: I'm not sure that there is much to be accomplished by having this `get_item` method in the helper since all it does is perform a lookup specific to the context of the caller. The complexity of `get_item` will also grow if there are more "items" in the future.

I think changing this line to just `assignment = current_course.assignments.find(id)` would suffice.

app/controllers/google_calendars/assignments_controller.rb
```diff
@@ -0,0 +1,36 @@
...
+    if assignment.due_at.nil?
+      # rubocop:disable AndOr
+      redirect_to assignment, alert: "Google Calendar requires Assignment to have at least END time!" and return
+    end
+    item_hash = add_single_item(current_user, assignment)
```
jonathoy: I know this is a hash, but maybe it would be more appropriately named if it were something like result.

app/controllers/google_calendars/assignments_controller.rb
```diff
@@ -0,0 +1,36 @@
...
+    process_hash(item_hash)
+  end
+
+  def add_assignments
+    assignment_list = get_all_items_for_current_course(current_course, "assignment", current_user)
```
jonathoy: What if you condensed lines 29-32 here to just something like `result = add_multiple_items(assignment_list, assignments_path)`? By doing so your controller method doesn't have to have all this logic. You might not need to pass around the `current_user` as you do on line 31 if you move back out the logic in `get_all_items_for_current_course`. Does that make sense?

app/helpers/google_calendars_helper.rb
```diff
@@ -47,6 +76,15 @@ def generate_open_date_if_one_does_not_exist(item)
     end
   end
 
+  def refresh_google_calendar_authorization(current_user)
+    calendar = Calendar::CalendarService.new
+    client_id = Google::Auth::ClientId.new(ENV['GOOGLE_CLIENT_ID'], ENV['GOOGLE_SECRET'])
+    secrets = create_google_secrets(get_google_authorization(current_user))
+    calendar.authorization = secrets.to_authorization
+    calendar.authorization.refresh!
```
jonathoy: I noticed that when you call `get_google_authorization` on line 82, the following happens:

```
refresh_if_google_authorization_is_expired(google_authorization)
```
Then on line 84 I see `calendar.authorization.refresh!`. I might be missing something but at a glance this seems redundant?

config/routes.rb
```diff
@@ -185,9 +185,24 @@
     resources :courses, only: [:create, :destroy, :index], module: :integrations
   end
 
-  resource :google_calendar, only: [ ] do
-    collection do
-      post "/:class/:id", action: :add_to_google_calendar, as: :add_to_google_calendar
+  # resource :google_calendar, only: [ ] do
```
jonathoy: Could you verify that this is no longer used?

spec/helpers/google_calendars_helper_spec.rb
```diff
@@ -13,20 +13,78 @@
...
     it "returns the assignment of the corresponding event object" do
-      expect(get_event_or_assignment(course, "assignment", assignment.id).id).to be assignment.id
-      expect(get_event_or_assignment(course, "assignment", assignment.id).class).to be Assignment
+      expect(get_item(course, "assignment", assignment.id).id).to be assignment.id
+      expect(get_item(course, "assignment", assignment.id).class).to be Assignment
+    end
+  end
+
+  describe "#get_all_items" do
+    let(:user) { create :user }
```
jonathoy: I'm not sure I see a reason why most of these `let`s  need to be duplicated x3 for each of the tests described here since there are no differences between them. Line 61, for instance, can stay where it's at since it is specific to `#retrieve_visible_assignments`, but things like `course` and `professor` only need to be declared once. If you put them at the beginning of the test, they should be lazy-loaded only as needed.

app/helpers/google_calendars_helper.rb
```diff
@@ -55,4 +93,40 @@ def create_google_secrets(google_authorization)
...
+      return {"message_type" => "notice", "message" => "#{item_list_filtered.count} item(s) successfully added to your Google Calendar"} unless item_list.count != item_list_filtered.count
+      return {"message_type" => "notice", "message" => "#{item_list_filtered.count} item(s) successfully added to your Google Calendar. #{item_list.count - item_list_filtered.count} item(s) were not added because of missing due date(s)."}
+    rescue Google::Apis::ServerError, Google::Apis::ClientError, Google::Apis::AuthorizationError, Signet::AuthorizationError
+      return {"message_type" => "alert", "message" => "Google Calendar encountered an Error. Your " + item.class.name + " was NOT copied to your Google calendar."}
+    end
+  end
+
+  def add(current_user, item)
+    calendar = refresh_google_calendar_authorization(current_user)
+    result = calendar.insert_event('primary', create_google_event(item))
```
jonathoy: It doesn't seem like this variable assignment is needed.

app/helpers/google_calendars_helper.rb
```diff
@@ -55,4 +93,40 @@ def create_google_secrets(google_authorization)
...
+
+  # note: if there is a server error during a batch process the items processed before the server error will be copied to the associated google calendar
+  def add_multiple_items(current_user, item_list, item_list_filtered)
+    begin
+      item_list_filtered.each do |item|
+        add(current_user, item)
+      end
+      return {"message_type" => "notice", "message" => "#{item_list_filtered.count} item(s) successfully added to your Google Calendar"} unless item_list.count != item_list_filtered.count
+      return {"message_type" => "notice", "message" => "#{item_list_filtered.count} item(s) successfully added to your Google Calendar. #{item_list.count - item_list_filtered.count} item(s) were not added because of missing due date(s)."}
+    rescue Google::Apis::ServerError, Google::Apis::ClientError, Google::Apis::AuthorizationError, Signet::AuthorizationError
```
jonathoy: My CodeClimate extension shows that this case might be untested.

spec/controllers/google_calendars/assignments_controller_spec.rb
```diff
@@ -0,0 +1,107 @@
+require "rails_spec_helper"
```
jonathoy: @chcholman Do you know if we need these anymore? Are they automatically included now in our RSpec configuration?

chcholman:
@dovek Is this ready to be reviewed again?

dovek:
@chcholman Not quite yet. Resolving some merge conflicts then it should be good to go.

dovek:
@chcholman Yes, this is ready for review again!

chcholman:
@dovek This is fantastic, thanks so much Kyle!

PR #3278, opened 2017-05-17
3244 ng refactor

acousticrobot:
### Status

**Ready**

### Description

This PR removes the last of the Angular controllers and factories.  This allows us to load more than one angular directive per page, so we can move forward with an interactive faculty on-boarding widget in the left hand column.

#### Badges Controller

The Badges Controller on the students view of the badges index page has been changed into the `studentBadgeIndex`, loaded from `/app/views/badges/_index_student.haml`. All past functionality should be the same.

#### Weights Controller

The interactive weighting on `/assignment_type_weights`, is now handled by the `assignmentTypeWeights` directive. All past functionality should be the same.

#### Rubrics Design

The majority of the PR changes the way Rubrics are designed, removing the controller and all factories, and managing it instead through the Rubric Service and Directives.

Updates to the RubricsController also affect editing a grade with a rubric.

### Related PRs

Weights need to be tested again once #3277 is merged into master and rebased against this branch.

### Migrations
NO

### Steps to Test or Reproduce

Test functionality on:

* Assignment Weights Page
* Badges Index Page
* Grade Edit Page (with rubric)
* Rubric Design Page

======================
Closes #3244

PR #3294, opened 2017-05-23
3023 criterion order

### LINKED ISSUE #3023: Rubric Design Page: Update Criterion Order
### Actual Behavior

Reordering Criteria on the rubric design page works, but it causes an Angular error, viewable in the console. This could be simply a matter of sending a better response from CriteriaController#update_order.

Possibly best would be to move this route to the `API::CriteriaController` and return the json used for the index. `@criterion = Criterion.update...` and then map the errors before returning success.

### LINKED ISSUE #3280: remove level badges route
### Expected Behavior

Unless there is a form I don't know about, Level badges should be handled through the API. We should remove these resources and routes: `resources :level_badges, only: [:create, :destroy]`

This requires #3278 to be merged.

This relates to #3087

acousticrobot:
### Status
**READY**

### Description
  * Fixes criteria ordering functionality on the rubric design page.
  * Adds a drag handle and limits drag and drop action to the handle
  * removes legacy routes for criteria and level badges

### TODO

- [x] Add UX improvement solution in this branch

======================
Follows #3278 
Closes #3280 
Closes #3023


chcholman:
Super nitpicky - can you make the z-index of the criterion border fall beneath the Rubric total tab at the top of the page? Otherwise ready to go!

chcholman:
@acousticrobot That makes sense. I'm okay with the pane hiding, but what about adding a white bkg to the box as well? It looks odd when you drag it over the sidebar, etc. and would feel more finished if it just travelled with a bkg I think? 

Have you tried dragging a criterion down? It's surprisingly hard to do. Is there any way we can support better here?

chcholman:
@acousticrobot This looks awesome! A much better solution. 

@mariehooper Will you take a final pass on design and accessibility?

PR #3304, opened 2017-05-31
Removed extraneous fields

chcholman:
### Status
**READY**

### Description
This is the second (final) pass at matching production data to the local schema.rb file


chcholman:
@acousticrobot Will you do a last check to make sure that all of these make sense to you?

chcholman:
@acousticrobot Great catch! As far as I can see the relationship was not one we were using anywhere - are you aware of anywhere I'm missing? (posting it to run the test suite next and see)

PR #3311, opened 2017-06-01
3121 quick grade style refactor

### LINKED ISSUE #3121: Refactor mass editing of grades
We should refactor the mass_edit page for grades in an assignment so that it's possible to dynamically change the quick grade style on the same page. Currently the option is on the Assignment edit page, and it has the potential to be unclear as to what the meaning of the setting is.

This would likely require us to redo the page in Angular.

See #2956 for additional details that were discussed as a result of the discussed bug.

jonathoy:
### Status
READY

### Description
The goal of this PR is to allow the dynamic changing of the quick grade styling from the mass edit page on individually graded assignments.

**At the moment, we will also need to refactor the group grading form as well in order to allow this capability.** See issue #3321 

In order to accomplish this, the mass edit form was rewritten in Angular along with the team selector filter that was previously rendered with it.

Refer to #2956 for previously discussed details on available quick grading options.

### Gem dependencies
N/A

### Migrations
NO

- Ensure that the students in the table are updated when the team filter selection changes
- Ensure that mass editing of grades works as expected for all individual assignment types

Mass grading

======================
Closes #3121


jonathoy:
@chcholman Feel free to take a look at this when you are ready. I came upon the realization that this work will need to be performed on the group mass grading page as well before we can start to remove references to quick grade style in places such as the assignment edit page and on the model. Maybe that work should be part of a separate PR? Thoughts?

chcholman:
@jonathoy Totally agree that it makes sense to do group mass grading in a separate PR and then remove all references!

jonathoy:
@chcholman Thoughts on @acousticrobot's comments?

chcholman:
@jonathoy @acousticrobot 

Thanks so much for the feedback and work both of you!

I totally agree re the UX of the "No Change", and am definitely up for this to receive more UX testing/another rethink. The key thing is making sure there's the ability to mark it as passed, failed, or no grade at all so we're not sending out information/notifications when we shouldn't be. 

If instructors want to 1) create rubrics and then 2) hack their way to the path for quick grading, I'm actually okay with that (realistically, it's the same as doing importing via a spreadsheet). I would *love* for us to get to do 3, which amounts to quick grading for rubrics, but yes, the concerns regarding a single page navigating 300 rubrics are intense. For now, I vote we skip concerns about 2, think about what we can do as we build towards 3, and continue talking about what on earth we can do to improve 1.

jonathoy:
@chcholman The changes you requested should be in now!

PR #3313, opened 2017-06-02
3155 autosave grade scheme elements

### LINKED ISSUE #3155: Autosave grade scheme elements on mass_edit page

jonathoy:
### Status
READY

### Description
This PR adds the ability to make grade scheme element changes and have them persist automatically in the background. Any change made on the form, whether it is an input change or the removal of a persisted element will save after a 3 second delay.

Once the AJAX call is initiated, the user will see the caption 'Saving..' on the affected element followed by either a success caption or an alert if the save was unsuccessful.

If the save is for the removal of a persisted element, an alert will inform the user regardless of whether the change was successful or not.

As part of this PR, the previous `mass_update` method on the `GradeSchemeElementsController` was moved to the API namespace and refactored for better error handling. Also, a fair amount of cruft was trimmed off some of the grade scheme element related specs.

Furthermore, onChange now triggers autosave as well as the dynamic reordering of grade scheme elements by lowest points.

### Migrations
NO

Ensure that all features work as expected when adding, updating, or deleting grade scheme elements. Changes should persist oChange after a 3 second day and messages should appear to inform the user on the status of the save. Grade scheme elements should also reorder dynamically onChange based on lowest points in descending order.

Mass editing for individually-graded assignments

======================
Closes #3155

### DIFF
--- a/app/assets/javascripts/angular/directives/common/maintain_focus.coffee
+++ b/app/assets/javascripts/angular/directives/common/maintain_focus.coffee
@@ -1,0 +1,15 @@
+# Keep focus on the input when the DOM is being updated or re-rendered
+# Inspired by: https://stackoverflow.com/a/31929381
+@gradecraft.directive 'maintainFocus', ['$timeout', ($timeout) ->
+  {
+    restrict: 'A'
+    require: 'ngModel'
+    link: ($scope, $element, attrs, ngModel) ->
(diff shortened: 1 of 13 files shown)


jonathoy:
@acousticrobot By changing it to just update onChange versus onBlur and adding a common directive to maintain focus after reordering, I think that your concerns should be addressed.

Issue #3314, opened 2017-06-02
Assignment boxplot no longer triggered by tabs


acousticrobot:
### Expected Behavior

[This code](https://github.com/UM-USElab/gradecraft-development/blob/master/app/assets/javascripts/angular/directives/analytics/assignment_distribution.coffee#L19) should trigger the boxplot when the tab is active, insuring it is drawn to the proper size.

### Actual Behavior

It appears the watch is no longer being triggered. I believe the class is no longer on the page and possibly the even is no longer triggered as well.

### Steps to reproduce

View an assignment as a student and click on the class analytics tab, the plot is only drawn to correct size on window resize.

@mariehooper I am assigning this to you since you know better than I how this was originally working, and also what html classes might have changed since then.

Issue #3110, opened 2017-03-14
Move analytics out to the API namespace


chcholman:

acousticrobot:
Examples:
[on assignment](https://github.com/UM-USElab/gradecraft-development/blob/master/app/models/assignment.rb#L291)
[on badges](https://github.com/UM-USElab/gradecraft-development/blob/master/app/models/badge.rb#L53)
[in presenter](https://github.com/UM-USElab/gradecraft-development/blob/master/app/presenters/info/dashboard_weekly_stats_presenter.rb)

acousticrobot:
Assignments handled in #3151

acousticrobot:
Course and Badges moved in #3414

acousticrobot:
Closing issue, can reopen or create new tickets for any other instances we find.

PR #3315, opened 2017-06-02
Course analytics

acousticrobot:
### Status
**READY**

### Description
 * Moves course scores boxplot into Angular directive
 * Adds reusable boxplot directive for both scores boxplots
 * removes unused JQuery Plot for points breakdown
 * removes a before action that was querying course scores on every controller action across the app
  * removes analytics methods from course model

### Related PRs

#3110 - moving analytics to API
#3314 - ticket to fix box that is nested in a tab

### Migrations
NO

View the boxplot on the dashboard and verify that it still works as expected.
View the boxplot for an assignment score, verify that it still works as expected.


chcholman:
@acousticrobot Fantastic cleanup, so nice to see.

PR #3324, opened 2017-06-05
update analytics redraw on tab activate function

### LINKED ISSUE #3314: Assignment boxplot no longer triggered by tabs

mariehooper:
### Status
**IN DEVELOPMENT**

### Description
This PR fixes a function that draws charts inside jQuery UI tabs that was originally targeting ng-app which has recently been consolidated to one initialization for GradeCraft.

### Todos
- [x] Remove unnecessary `$rootElement`
- [x] Replace '`plotGraph` function for `assignmentDistributionAnalytics` box plot.

### Migrations
NO

Checkout analytics on assignment for instructor or student.

* Assignment analytics

======================
Closes #3314

### DIFF
--- a/app/assets/javascripts/angular/directives/analytics/assignment_distribution.coffee
+++ b/app/assets/javascripts/angular/directives/analytics/assignment_distribution.coffee
@@ -1,12 +1,12 @@
 # box plot for assignment grade distribution
-@gradecraft.directive 'assignmentDistributionAnalytics', ['$q', '$window', '$rootElement', 'AnalyticsService', 'DebounceQueue', ($q, $window, $rootElement, AnalyticsService, DebounceQueue) ->
+@gradecraft.directive 'assignmentDistributionAnalytics', ['$q', '$window', 'AnalyticsService', 'DebounceQueue', ($q, $window, AnalyticsService, DebounceQueue) ->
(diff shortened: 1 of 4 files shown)


mariehooper:
Hey @acousticrobot - this is mostly resolved except the assignment boxplot - it looks like the `plotGraph` function was removed in [this commit](https://github.com/UM-USElab/gradecraft-development/commit/9f5633dad42e75b84bce26794ac2c37d744db8a1) so the boxplot is not currently drawing until the window refreshes... but everything else is now working as intended. It was the consolidation of ng-app to body instead of all the instances that cause need for update.

PR #3331, opened 2017-06-08
1976 no delete users

### LINKED ISSUE #1976: Don't delete deleted Users
As an instructor, if I delete a student, they should be removed from having an 'active' course membership, but the course membership should continue to exist, and they should be retrievable - including all grades, earned badges, submissions, etc. 

This is valuable for both data purposes (knowing the history of what happened), and managing user errors.

dovek:
### Status
**IN DEVELOPMENT**

### Description
As an instructor, if I delete a student, they should be removed from having an 'active' course membership, but the course membership should continue to exist, and they should be retrievable - including all grades, earned badges, submissions, etc.

This is valuable for both data purposes (knowing the history of what happened), and managing user errors.

### Related PRs
N/A

### Todos
- [ ] Add Tests
- [x] Finish Development

### Deploy Notes
Will need to run db:migrate

### Gem dependencies
N/A

### Migrations
YES

TO BE COMPLETED

* Users
* Course Memberships

======================
Closes #1976


dovek:
@chcholman During review, Jonathon noticed that there is a course_id field on both the badge and earned badge table. Is this necessary?

chcholman:
@dovek Yup! And we'll probably do some further work to add `course_id` to more tables over the next couple of months. Turns out: that's SUPER helpful when something gets deleted and we need to pull from a database backup. Otherwise, finding the deleted+associated objects is really hard.

app/models/course_membership.rb
```diff
@@ -16,8 +16,10 @@ class CourseMembership < ActiveRecord::Base
   end
 
   scope :auditing, -> { where( role: "student", auditing: true ) }
-  scope :being_graded, -> { where( role: "student", auditing: false) }
+  scope :being_graded, -> { where( role: "student", auditing: false, active: true) }
   scope :instructors_of_record, -> { where(instructor_of_record: true) }
+  scope :active, -> { where(active: true)}
```
jonathoy: I'm not seeing the `:active` or `deactivated` scopes that you added here being used anywhere. I would recommend either removing these to avoid adding unnecessary logic or using them in places such as lines 154, 158 on the courses model.

```
course_memberships.active.where(role: "student").count
...
course_memberships.active.where(role: "student", auditing: false).count
```

config/routes.rb
```diff
@@ -206,7 +206,13 @@
...
   end
 
-  resources :course_memberships, only: [:create, :delete, :destroy]
+  resources :course_memberships, only: [:create, :delete, :destroy] do
+    collection do
+      put "deactivate/:id", action: :deactivate, as: :deactivate
+      put "reactivate/:id", action: :reactivate, as: :reactivate
+    end
+  end
+  get :course_creation_wizard, to: "courses#course_creation_wizard"
```
dovek: I didn't add course_creation_wizard. I think this came through when I was working on my merge conflicts.

chcholman:
@dovek I posted a tiny refactor to leverage the scopes so that we can be sure we're being consistent with who is filtered out/when. 

Next piece to this: Take a look at the `app/models/concerns/assignment_analytics.rb` - we need to figure out how to get the deactivated students grades out of these calculations. You can see them displayed in the interface here: http://localhost:5000/per_assign (this page is currently on the InfoController).

Issue #3334, opened 2017-06-08
remove grade status field


acousticrobot:
Once we have up-to-date `complete` and `student_visible` fields on `Grade`, we can drop `status` and run this logic through these booleans.

PR #3303, opened 2017-05-30
3095 grade release process

acousticrobot:
### Status
**READY FOR REVIEW**

### Description

This PR is the logic implementation step of the new grade status refactor. It Simplifies the grading release process on the grade form and assignment show page.  Grades can be marked as "Complete", and when they are released to students, they are set to "Student Visible". These states are handled through boolean fields, `grade.status` becomes obsolete. It uses the same pattern for grading Challenges. It also removes the option to require release on assignments; all assignment are treated with the same logic.

### Deploy Notes

Must follow all migrations and rake tasks in:

#3327 (merged and tasks are run)
#3376 (merged and tasks are run)
#3393 (merged and tasks are run)

### Migrations

NO, all migrations should be added first as stated above.

### Steps to Test or Reproduce

Grade assignments and try setting them to either not complete, complete, or visible to students.

Review in progress grades and use the check boxes to release in progress grades on:
  *  grade status page (`/grading_status`)
  * individual assignment show page
  * group assignment show page

Test the same with group grades.
Test student released grades (release should be managed automatically).
Test the same with grading challenges.

### Impacted Areas in Application

* Grading and releasing grades for assignments
* Grading and releasing grades for challenges
* Importing grades

======================
Closes #3095

PR #3335, opened 2017-06-08
3327 grade submission fields

### LINKED ISSUE #3327: new grade release fields
Before we can update the grade release process, we want to:

  * add new fields to grades: `complete` and `visible_to_student`
  * update these fields as grades are saved
  * design rake tasks that will fill these fields in on existing grades

acousticrobot:
### Status
**READY**

#### New fields

Adds new boolean fields `complete` and `student_visible` to `Grade`. These will be used in future pull requests to manage the grade status.

The previous states and the new fields map as such:

| status | complete | student visible  |
| ----- | --------- | --------------- |
| "In Progress | false |   false   |
|  "Graded"   |  true  |   false  |
| "Released" | true | true |

It adds a temporary `before_save` action to maintain these fields on new grades.

#### New Sample Data

Cleans up and adds better sample data for grades in all grading states. 

#### Rake Task: `grades:update_status_fields`

A rake task is included to update the fields on existing grades. 

To test the rake task with the included sample data, comment out `before_save :update_status_fields` in `Grade` before running samples.

#### Removed Route

As far as I can tell the update route on the Grades controller is no longer in use, as the grade for updates entirely via the api.  I have removed this route and specs.

### Related PRs

Updating the status will require the following in order:

  * This PR( ticket #3327)
  * ticket #3334, A PR that transfers the logic over to these fields, and drops `grade.status`
  * PR #3303, which adds the new grading release UX, and removes `release_necessary` logic.

### Deploy Notes

  * run migration
  * run `rails grades:update_status_fields`

### Migrations
YES

### Steps to Test / Impacted Areas in Application

Because of all the different ways a grade can be created updated, I decided that the `before_save` was the safest way to make sure these fields are maintained.

  * grades via Grading form
  * Mass grading
  * Self grading
  * Importing Grades

======================
Closes #3327


app/models/concerns/grade_status.rb
```diff
@@ -60,11 +60,32 @@ def graded_or_released?
...
+        self.student_visible = false
+      else
+        self.complete = true
+        self.student_visible = true
+      end
+    elsif self.status == "Released"
+      self.complete = true
+      self.student_visible = true
+    end
+    return true
```
acousticrobot: If a `before_save` callback returns false, AR will raise an error. In progress grades will return false, since this is the last value assigned, so I have to ensure that we always return true.

jonathoy: Ah I see it - thanks for clarifying!

chcholman:
@acousticrobot Awesome! I'm up for that all happening tomorrow if you are.

PR #3337, opened 2017-06-09
alert for course creation errors

### LINKED ISSUE #3322: Better New Course Creation Form Needed
### Expected Behavior

If a new course does not have both a Number and a Title, the new course form button should be disabled. Errors messages should be returned describing what went wrong when a new course is not created.

### Actual Behavior

One can submit a course missing a number or title, but it will not be created, and no flash message we be displayed.

acousticrobot:
### Status
**READY**

### Description

Adds error messages to failed course creation. We could also add javascript validation but this seemed like a simpler solution to maintain for such a simple form.

### Related PRs

May affect #3182 - alerts refactor, on rebase. This removes the surrounding format block from the controller methods, and adds the `render partial: "layouts/alerts"` that won't be needed once this partial is centralized.

Add a new course without a number, name, or both, and you will now see an error message on redirect.

======================
Closes #3322

### DIFF
--- a/app/controllers/courses_controller.rb
+++ b/app/controllers/courses_controller.rb
@@ -77,6 +77,7 @@ class CoursesController < ApplicationController
       redirect_to edit_course_path(@course),
       notice: "Course #{@course.name} successfully updated"
     else
+      @institutions = Institution.where(has_site_license: true)
       render action: "edit"
     end
   end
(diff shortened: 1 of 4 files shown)


chcholman:
Will you take a look at the failing specs here?

chcholman:
@acousticrobot Adding a course without the requisite items gets proper validation, but updating a course to not having some required hits a hard error. Would it make sense to fix that in this branch or a separate one?

PR #3363, opened 2017-06-13
3042 limit pass condition

### LINKED ISSUE #3042: Block users from selecting pass/fail if the assignment isn't pass/fail
When I'm creating an unlock condition, I shouldn't be able to select "Passed" as an option unless the assignment selected as the condition is pass/fail

acousticrobot:
### Status
**READY**

### Description

When selecting an assignment as a condition for another assignment, this PR limits the state Achieved to include "Grade Earned" for assignments with points, or "Passed" for pass/fail assignments.

### Migrations
NO

  * Add an assignment condition on an assignment edit page.
  * Choose an graded assignment and verify the options for "State Achieved"
  * Choose a pass/fail assignment and again verify the options
  * Save a condition and verify that the state is preserved when returning to the edit page
  * Verify the options for the saved condition are proper on page load.

======================
Closes #3042

### DIFF
--- a/app/assets/javascripts/assignments.js
+++ b/app/assets/javascripts/assignments.js
@@ -36,11 +36,15 @@ function filterAssignmentConditions(select) {
   assignment = _.find(data,{id: parseInt(id)});
   if (assignment.pass_fail) {
     $("#assignment-state-achieved option[value='Grade Earned']").remove();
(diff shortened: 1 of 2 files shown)


chcholman:
This looks great, thanks so much @acousticrobot!

PR #3367, opened 2017-06-14
3346 group resubmissions bug

### LINKED ISSUE #3346: Group resubmissions not showing up as resubmissions
As an instructor, I expect groups to be able to resubmit assignments, and for these to appear on the grading_status resubmissions section

jonathoy:
### Status
READY

### Description
This PR address a bug where a group submission was not showing on the grade status page if it was a resubmission. The issue was in the `AssociatesSubmissionWithGrade` service, which was not properly linking the grade with the submission since the group grades were being updated iteratively in the Angular service.

**This also includes a fix for a preexisting bug in master where the group was not within the context of the assignment guidelines partial.**

### Gem dependencies
N/A

### Migrations
NO

Create a submission on a group-graded assignment, and grade it as an instructor. Resubmit and ensure that the submission appears on the grading status page.

`AssociatesSubmissionWithGrade` service and group grading

======================
Closes #3346

### DIFF
--- a/app/assets/javascripts/angular/directives/student/submission/text_comment_input.coffee
+++ b/app/assets/javascripts/angular/directives/student/submission/text_comment_input.coffee
@@ -5,10 +5,12 @@
     vm = this
     vm.loading = true
     vm.submission = StudentSubmissionService.submission
+
     vm.queueSaveDraftSubmission = () ->
       StudentSubmissionService.queueSaveDraftSubmission(vm.assignmentId)
 
     StudentSubmissionService.getDraftSubmission(vm.assignmentId).then(() ->
+      vm.submission.text_comment_draft = vm.submission.text_comment if !vm.submission.text_comment_draft?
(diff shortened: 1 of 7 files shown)


jonathoy:
@acousticrobot Thanks for the feedback! I seem to be having some trouble replicating what you're describing with the submission text not showing up though. You are talking about the froala text area not showing the submission text, correct?

jonathoy:
5603f11 should fix #3375

jonathoy:
@chcholman My workflow for testing this functionally still seems to work for me, so we can walk through it together at some point to confirm.

I did notice that the button text never changes to say "Resubmit" however, and this appears to be because the text is conditional based upon the method `will_be_resubmitted?` in the Submissions model. The method relies on the `submission_grade`, which at a glance does not seem to account for group grades. A fix for this would potentially impact other portions of the app as well. Thoughts?

PR #3386, opened 2017-06-19
fix specs that fail for unique key violation from database

acousticrobot:
### Status
**READY**

### Description

Reverts failing specs to using doubles. Doubles run much faster for these specs than hitting the database, but if there is a reason why we want these specs to run with real AR models, the other solution would be to remove the hard-coded ids from the calls to the factories.

See: https://app.codeship.com/projects/106957/builds/26122135?pipeline=db3badd7-664d-4ab2-9fed-2a90c324381b as an example of how these specs fail due to hard-coded ids.

### DIFF
--- a/spec/presenters/submissions/show_presenter_spec.rb
+++ b/spec/presenters/submissions/show_presenter_spec.rb
@@ -7,12 +7,12 @@ describe Submissions::ShowPresenter do
     { course: course }
   end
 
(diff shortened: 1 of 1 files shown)


chcholman:
@acousticrobot I'm *all* for fixing the new quite-regularly failing specs! But I'm not convinced this actually speeds up the test suite - this came in at the current average time of just over 20 minutes...

chcholman:
Oh, that is awesome! Thanks for checking!

PR #3396, opened 2017-06-21
3348 import assignments from csv

### LINKED ISSUE #3348: Import assignments from CSV
### Feature
Just like with users and grades, I expect to be able to import assignments via a csv import. The fields should be: Assignment Name, Assignment Type Name, Point Total, Description, Due Date. The interaction is going to need to both: 1) create an assignment type if one doesn't already exist, and 2) help figure out what due date you meant.

jonathoy:
screen shot 2017-06-21 at 3 41 18 pm

### Status
READY

### Description
This PR adds the ability to import new assignments via CSV. It will make an attempt to do a case-insensitive match on assignment type and try to parse the date.

After uploading the file, the user will be presented with a table to review the imported results. If the assignment type does not match one that currently exists for the current course, it can be mapped to an existing one or set to the blank option to create new. The date must match the displayed format to correctly parse. However, the user can choose to change the date and time using the date time picker component.

Upon importing, the successful and unsuccessful rows will be displayed.

### Gem dependencies
N/A

### Migrations
NO

Download the sample template and attempt to create assignments using various permutations of inputs.

CSV Assignment Import

======================
Closes #3348


jonathoy:
The brakeman warning in Codeclimate is a false positive, since the before filter ` ensure_allows_file_imports` should whitelist the possible templates that can be rendered.

jonathoy:
@acousticrobot Could you take a look at this? Since it's a CSV upload, there's just so many edge cases to account for when testing.

chcholman:
@jonathoy Functionally, this is working beautifully. I'm going to pass it to @mariehooper for more thorough design refinement next!

mariehooper:
Hey @jonathoy I think this looks good! The piece I would change would be have the text for Assignment Type and Date appear in the fields for each row representing an assignment that has been uploaded rather than below. I was a bit confused as to what I was looking at when I first uploaded an assignment, and I think a lot would be solved if the info appeared in the field. It also might be helpful to add an asterisk to the names of the assignment types that will be created since that will link your asterisked note below the list a little more clearly.

screen shot 2017-06-27 at 5 01 27 pm

jonathoy:
@mariehooper Would you mind taking another look at the latest changes? One change that I did make was I removed the null or "empty" option from the assignment type selector if the value in the csv matches an existing type in Gradecraft. Not selecting an assignment type if one is matched would be considered an invalid form submission anyway.

PR #3400, opened 2017-06-22
3398 refine gse ordering

### LINKED ISSUE #3398: Reorder Grade Scheme Elements on Blur
Evan's concern is that when editing existing grade scheme elements, it becomes jarring to have the elements reorder dynamically on change especially if the user wants to keep reference of where the element is currently in relation to others.

To improve this, we can change the reordering event to trigger on blur. We should also ensure that the text input maintains focus after the save event.

jonathoy:
### Status
READY

### Description
Per Evan's experience, the grade scheme elements reordering dynamically on change is a bit jarring. The elements now reorder once they have been saved. Save still occurs after a 3 second delay on change.

Change the point threshold on a grade scheme element and ensure that the ordering only happens after save.

======================
Closes #3398

### DIFF
--- a/app/assets/javascripts/angular/directives/grade_scheme_elements/element.coffee
+++ b/app/assets/javascripts/angular/directives/grade_scheme_elements/element.coffee
@@ -48,5 +48,10 @@
         DebounceQueue.addEvent(
           'gradeSchemeElement', 'saveChanges', _save, [scope, isRemoval], 3000
         )
+
+      # If lowest_points changes, reorder the elements accordingly
+      scope.$watch('lowest_points', (newValue, oldValue) ->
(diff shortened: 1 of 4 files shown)


jonathoy:
@acousticrobot Minor inconvenience, but on save the input field loses focus. I looked briefly into it but I didn't want to spend too much time trying to figure out how the `maintain-focus` directive would need to change to accommodate this. Not sure if you had had quick thoughts on this?

jonathoy:
@chcholman What are you thoughts on this? It's otherwise ready to go.

chcholman:
@jonathoy @acousticrobot I agree some animation here could be lovely, but this addresses the jarring nature really nicely. I'm going to merge, and include this as one of the areas we may start to layer animation in a bigger convo with @mariehooper!

PR #3405, opened 2017-06-22
3395 resubmit bug for groups

### LINKED ISSUE #3395: Update buttons for group resubmission workflow to properly reflect "resubmitting" state

jonathoy:
### Status
READY

### Description
This bug addresses an issue where the submission buttons were not displaying the correct text due to the fact that they were relying on a method which did not account for group submissions.

### Migrations
NO

As staff, if you attempt to edit a submission and the submission will be resubmitted, then the caption on the edit submission button should read "Resubmit" instead of "Edit Submission".

As a student, if you are editing a submission that has draft content, the edit submission caption should be "Edit Draft". Otherwise, it will read either "Resubmit" or "Edit Submission" according to the correct condition.

Submissions

======================
Closes #3395

### DIFF
--- a/app/models/submission.rb
+++ b/app/models/submission.rb
@@ -174,6 +174,16 @@ class Submission < ActiveRecord::Base
     end
   end
 
+  def term_for_edit(user_is_staff)
+    if !user_is_staff && text_comment_draft.present?
+      "Edit Draft"
+    elsif will_be_resubmitted?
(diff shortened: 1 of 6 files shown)


jonathoy:
@acousticrobot This PR might be of potential interest to you since I noticed that you had opened 3401, which is related.

PR #3419, opened 2017-06-28
3370 gradebook refactor

### LINKED ISSUE #3370: Gradebook refactor
The revived gradebook page is still plagued by pageload times. We need to do whatever we can to speed this up - including potentially loading the content dynamically after hte page is loaded

jonathoy:
### Status
READY

### Description
This PR is a refactor of the current gradebook so that the data can be loaded asynchronously via AJAX. This addresses the issue where the page would time out when the resultset was too large.

Students are currently batched out into groups of 25.

### Migrations
NO

Attempt to load the gradebook for any given course, ensuring that the data is presented correctly and that the links are valid and columns sortable.

To test that results are batched correctly, the batch size can be changed temporarily in the `GradebookService#getStudents` argument.

Gradebook

======================
Closes #3370

### DIFF
--- a/app/controllers/api/gradebook_controller.rb
+++ b/app/controllers/api/gradebook_controller.rb
@@ -16,20 +16,20 @@ class API::GradebookController < ApplicationController
     end
   end
 
-  # Returns all grade ids for the current course; for fetching by batch
-  # GET api/gradebook/grade_ids
-  def grade_ids
-    render json: MultiJson.dump(current_course.grades.pluck(:id))
+  # Returns all student ids for the current course; for fetching by batch
+  # GET api/gradebook/student_ids
+  def student_ids
+    render json: MultiJson.dump(current_course.students.pluck(:id))
   end
 
-  # Returns gradebook data for the grades in the current course
-  # Optionally returns a subset of grades if provided an array of grade ids
-  # GET api/gradebook/grades
-  def grades
-    @grades = current_course.grades
+  # Returns gradebook data for the students in the current course
+  # Optionally returns a subset of students if provided an array of student ids
+  # GET api/gradebook/students
+  def students
+    @students = current_course.students
 
-    if params[:grade_ids].present?
-      @grades = @grades.where(id: params[:grade_ids])
+    if params[:student_ids].present?
+      @students = @students.where(id: params[:student_ids])
     end
   end
 end
--- a/app/controllers/api/gradebook_controller.rb
+++ b/app/controllers/api/gradebook_controller.rb
@@ -1,9 +1,35 @@
 class API::GradebookController < ApplicationController
   before_action :ensure_staff?
 
-  # Retusn all assignment names sorted by assignment type order, assignment order
-  # GET api/gradebook/assignment_names
-  def assignment_names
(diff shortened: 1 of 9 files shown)


jonathoy:
@acousticrobot If, while reviewing, you see some areas where the performance can be improved when loading these entries please let me know. I tried to do optimizations where possible, but this mainly fixes the issue of the page not loading at all and timing out for large datasets.

app/views/info/gradebook.html.haml
```diff
@@ -1,37 +1,6 @@
...
-                      - if grade
-                        = link_to "#{points grade.score}", edit_grade_path(grade)
-              - if current_course.valuable_badges?
-                %td= points student.earned_badge_score_for_course(current_course)
-              %td= points student.score_for_course(current_course)
-              %td= student.grade_level_for_course(current_course)
-              %td= link_to student.name, student_path(student)
+  %gradebook{"data-term-for-badge"=>"#{current_course.badge_term}",
+             "data-term-for-student"=>"#{current_course.student_term}",
+             "data-has-badges"=>"#{current_course.valuable_badges?}"}
```
acousticrobot: All of the other api calls that require term_for functionality send the terms in the meta data and rely on the api service helper `termFor`. Unless this doesn't work for your needs I would suggest we do that here too.

app/controllers/api/gradebook_controller.rb
```diff
@@ -0,0 +1,32 @@
...
+  # GET api/gradebook/student_ids
+  def student_ids
+    render json: MultiJson.dump(current_course.students.pluck(:id))
+  end
+
+  # Returns gradebook data for the students in the current course
+  # Optionally returns a subset of students if provided an array of student ids
+  # GET api/gradebook/students
+  def students
+    @students = current_course.students
```
chcholman: And here?

app/controllers/api/gradebook_controller.rb
```diff
@@ -0,0 +1,32 @@
...
+          name: assignment.name
+        }
+      end
+    end
+  end
+
+  # Returns all student ids for the current course; for fetching by batch
+  # GET api/gradebook/student_ids
+  def student_ids
+    render json: MultiJson.dump(current_course.students.pluck(:id))
```
chcholman: It looks like this might need some test coverage?

jonathoy: I think this is one of the situations where I was wondering about whether we needed coverage? Would the test not just be ensuring that the JSON body contains the IDs for all the students we create in our test course? The logic here is pretty much just plucking Active Record values and using MultiJson to return them. Happy to add them in if you feel it's important for regression testing.

app/controllers/api/gradebook_controller.rb
jonathoy: Same here - testing that the instance variable is set?

chcholman: I was less concerned about that and more about the next line where we dynamically set/override who exactly is part of current users, but I agree with you about generally testing less. I think it might ultimately make sense to have some feature testing here, but am going to merge this for now!

PR #3424, opened 2017-06-29
Restore Canvas state when launched via LTI

### LINKED ISSUE #3423: Restore State For Canvas Users When Authenticated Via LTI
Not sure if restoring the state from the database as opposed to the session will solve [this](https://rollbar.com/gradecraft/gradecraft-development/items/1986/) issue, but worth a shot since it continues to occur frequently.

More details regarding the handling of state by `omniauth-canvas` [here](https://github.com/atomicjolt/omniauth-canvas#state). This will affect users authenticating via LTI.

jonathoy:
### Status
IN DEVELOPMENT

### Description
While reading through the `omniauth-canvas` docs some time ago during our refactoring, I noticed [this](https://github.com/atomicjolt/omniauth-canvas#state).

It appears that we still frequently notice this error occurring in production:

```
OAuth::Signature::UnknownSignatureMethod: OAuth::Signature::UnknownSignatureMethod
  File "/usr/local/bundle/gems/oauth-0.5.1/lib/oauth/signature.rb" line 17 in build
  File "/usr/local/bundle/gems/ims-lti-1.1.13/lib/ims/lti/request_validator.rb" line 20 in valid_request?
  File "/usr/local/bundle/gems/ims-lti-1.1.13/lib/ims/lti/request_validator.rb" line 42 in valid_request!
  ...
```

See [here](https://rollbar.com/gradecraft/gradecraft-development/items/1986/) for the Rollbar details.

The hope is that this change will fix these issues that we have yet to fully understand or reproduce. It involves only including some middleware to handle the storing of the state differently. It would seem that there is little risk to implementing this change.

### Migrations
NO

?

Canvas via LTI

======================
Closes #3423


jonathoy:
@chcholman This may actually require us to update our `omniauth-canvas` gem to the latest version. We would be going from 0.1.1 to 1.01 which is a major version up. If possible, we should do some testing in staging before we push this to production

jonathoy:
I did a bit more digging and I'm a little bit concerned about the lack of documentation going on. The GitHub wiki gives instructions for preserving state, but it does not come with any migrations.

For the code to work, it appears to need at least the `OauthState` table and the `OauthStateMiddlewareException`, but the only info I can find on these two are in some related projects from the same owner.

https://github.com/atomicjolt/adhesion
https://github.com/atomicjolt/lti_starter_app

It's disappointing but for now, it may be best not to spend too much time on this.

jonathoy:
I've implemented some changes based on the recently updated documentation for the omniauth-canvas gem, but we will need to test that we can still authenticate via LTI. Perhaps this can be pushed to the staging environment?

jonathoy:
Note: In updating to the latest version of omniauth-canvas, you might find that there are some occasional errors that look like this:

```
You are setting a key that conflicts with a built-in method OmniAuth::AuthHash::InfoHash#name defined at /Users/jonathoy/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/omniauth-1.3.2/lib/omniauth/auth_hash.rb:34. This can cause unexpected behavior when accessing the key via as a property. You can still access the key via the #[] method.
```

According to [this](https://github.com/omniauth/omniauth/issues/872), it's because of an issue in the hashie dependency. Unfortunately, the only way to get rid of this is to remove the version limit in the gemfile for omniauth and upgrade to latest. This has the potential to affect our other omniauth strategies such as google-oauth2 and kerberos, so I'm going to suggest we hold off on this for now until we can at least determine if the changes proposed in this PR are beneficial.

chcholman:
@jonathoy This is ready to be tested in staging - let me know if you need help identifying courses to test LTI launch from staging, we have at least one already set up

jonathoy:
I'm closing this PR for now since it's going to require us to do a deeper dive - it's not been a well-documented issue. I will leave the branch for now in case we need to refer back to it.

PR #3425, opened 2017-06-29
3110 user analytics

acousticrobot:
### Status
**READY**

### Description

This PR moves the analytics off of the User model and into a conern.

It also renames the methods returning earned badges used for unlock conditions to `awarded`: `awarded_badges_for_badge` and `awarded_badges_for_badge_count`. This is in line with the existing `awarded_badges` relationship, and aims to make sure that information sent to the student only uses the `visible_earned_badges_for_badge` type methods.

### Related PRs

Other PRs addressing ticket #3110

### DIFF
--- a/app/models/concerns/analytics/user_analytics.rb
+++ b/app/models/concerns/analytics/user_analytics.rb
@@ -37,7 +37,7 @@ module Analytics
     end
 
     def earned_badge_score_for_course(course)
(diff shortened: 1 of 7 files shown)


chcholman:
@acousticrobot Weird - the file that you marked isn't a partial, it's a view that we do indeed use (It's the Awarded Badge Chart, third item down on the left hand navigation for staff). I tested to confirm that it does indeed still work with the changes.

PR #3432, opened 2017-07-03
Course Creation Refactor

chcholman:
### Status
**READY**

### Description
This fixes an issue where users who had failed to enter data correctly on the course form were redirected to the new form - thereby losing all of the data they'd entered. It also rearranges the structure of the course form files a little bit so that there's a single core form that gets rendered in both `new` and `edit`, and breaks out the basic settings components into a separate partial.

### Migrations
NO

* Course Creation and Editing


chcholman:
@mariehooper Will you take a look at this? I would like to use simple_form since it makes writing form views so much lighter weight, but I don't want to lose the accessibility that we've been careful to add. Let me know if there's something that I've missed and should be checking in implementing accessibility support!

chcholman:
@mariehooper Awesome, thanks so much for the quick check on this!

PR #3437, opened 2017-07-05
3110 weekly stats

acousticrobot:
### Status
**READY**

### Description

Replaces the weekly stats partial (on faculty dashboard, student dashboard, and faculty student view) with an Angular widget.

Removes entirely the weekly stats presenter.

I believe this will close 3110, but we should check for other random analytics methods first before closing.

### Related PRs

PRs addressing #3110 (moving analytics into the API)

### Migrations
NO

  * As faculty, view weekly stats on dashboard, should replicate data seen on master
  * As student, same on dashboard
  * As faculty, view student page, should be the same as the students weekly stats dashboard data
  * If nothing earned in the week (replicate with stale sample data, or delete earned grades/badges, etc.), a message to this effect should present, rather than empty lists of assignments, badges, points, etc.


app/controllers/api/courses_controller.rb
```diff
@@ -1,5 +1,5 @@
 class API::CoursesController < ApplicationController
-  before_action :use_current_course, only: :analytics
+  before_action :use_current_course, only: [:analytics, :one_week_analytics]
   before_action :ensure_staff?, only: :course_creation
```
acousticrobot: The rest are used by both faculty and staff, I'll remove the before_action while we have it in our sights.

app/assets/javascripts/angular/directives/analytics/weekly_stats.coffee
```diff
@@ -0,0 +1,67 @@
+# box plot for course grade distribution
+@gradecraft.directive 'weeklyStatsAnalytics', ['$q', 'AnalyticsService', ($q, AnalyticsService) ->
+    weeklyStatsCtrl = [()->
+      vmWeeklyStats = this
+      vmWeeklyStats.loading = true
+
+      services(vmWeeklyStats.studentId).then(()->
```
acousticrobot: Good call, I like this pattern better!

PR #3481, opened 2017-07-13
don't assign id in specs

acousticrobot:
### Status
**READY**

### Description
Removes the direct assignment of an id in a factory create (left over from the double). This was causing errors on codeship like [this one](https://app.codeship.com/projects/106957/builds/26770886?pipeline=b156c71f-a9a2-4525-a0f6-ab353da63587).


chcholman:
Awesome, thank you for this @acousticrobot!

PR #3483, opened 2017-07-13
3093 attendance assignment type

### LINKED ISSUE #3093: Add attendance assignment type
- Add boolean that marks an assignment type as attendance 
- Add analytics heatmap to student presence
- Create form entry for fast class event creation
- Add # of attended/past classes pane to dashboard

jonathoy:
### Status
READY

### Description
We need to create an assignment type that streamlines attendance taking.

### Todos
- [x] Allow marking existing assignment types as attendance
- [x] Include options dropdown on index table for quick grading, editing, etc.
- [x] Include ability to go back and mass edit existing attendance assignments
- [ ] #3484: Add ability to import attendance assignments from CSV

### Migrations
YES

On a new course, an instructor should be able to bulk create attendance events (assignments) via the newly added Attendance link in the staff nav bar. On this setup page, they can choose a date range and select specific days of the week to check for attendance with specific `open_at` and `due_at` times. The ability to add points is also available.

Once properly configured, the instructor can then click on Configure Events where they will be presented with a mass edit type view for each of the events. Events are automatically persisted once the required name field has been populated. Once an event is saved, they can also choose to upload a media image for the assignment.

Once one or more attendance events are saved, any subsequent visits to the Attendance link will bring you to an index page where you can create new individual events and mass edit existing. Quick grade, edit, copy, and delete are also available via the options dropdown.

As of right now, an attendance event is treated the same as an assignment for all intents and purposes.

Assignments, analytics

======================
Closes #3093


jonathoy:
Going to hold off on the last remaining changes to this until @acousticrobot's branch `3450-ng-condition` is finished and merged. Per our discussion, there is a lot of potential duplication in logic here that we can avoid when it comes to media uploads and assignment creation.

chcholman:
Functional testing notes for later work: 

- [ ] Deleting an event deletes the top one from the list, not the one clicked on
- [ ] the "Go back" button appears to be disabled? 
- [ ] Add Submit or Save button

jonathoy:
The go back button is disabled once there are one or more persisted attendance "events", although I am more than happy to discuss the functionality or need for this button. It became more complex once we allowed going back and having dates and times change post-save.

The rest of the UX was originally modeled after the Trello card. I'll put this in staging and maybe @evanstraub and others can give their thoughts on how to improve workflow. In the meantime, I have addressed the majority of the issues listed here.

magyarn:
UX observations:

1. When I create a level and click the name field, the text cursor does not appear. It's missing the indication that I can start typing.

2. After I'm done editing a learning objective, I expected to return to the accordion category where I last was. Instead I'm brought to the LO index page with all accordions closed.

3. The empty state should offer a brief explanation/distinction of learning objectives used for assessment purposes, as well as a link to a set of example LOs.

4. The form for creating a new LO could be more consistent with other forms used in the app.

5. Consider integrating "categories" as a step in creating a new LO, rather than a prerequisite action (this would help users advanced towards their goal of creating a LO more readily).

6. "Add Levels" button could be more visible/prominent.

7. Consider making the "Options" button for each LO look more secondary to the primary action of adding a new LO. That way there's one clear primary action for the page. Secondary style = buttons with only blue outline and transparent background color.

jonathoy:
@chcholman I believe that this work can be merged pending any additional testing/feedback.

chcholman:
_Feedback_ from @acousticrobot, @cait is going through and checking to see where each piece is
- [ ] When start date and end date are incompatible, the error shows up on whichever one of the two was last entered rather than as a pair, and with generic “invalid” error. I think it would be better to have a notice similar to the assignments setting page and rails validation: “end date must be after start date” or similar.
- [ ] Same with the start/end time

- [ ] Clicking in the form anywhere to the left of the date/time pickers brings up the UX time picker widget. This was confusing to me when I tried to click out of the widget and it would pop back up again.

- [ ] edge case: user can’t use the time picker to pick 12:00am, they have to change the time for it to show up in the field.

- [ ] edge case: validation does not allow a night class to cross the 12am barrier.

Page 2:

- [ ] When the assignment persists, there is no visual indication that it has saved (left bar is still yellow).

- [ ] When attempting to delete an attendance event down in the list it deletes the first event.

 - [ ] I’m not sure about the “Go Back” option, and the workflow it supports. I see that as soon as one event has persisted there is no options to configure events based on a timeline. I would think this would either always be an option, or this option would cease to exist and I would have my persisted attendance events as soon as I hit “Configure Events”, not the first time I enter more information into an event.

- [ ] This is a use case decision, but does it make sense to require a title for each attendance event? If I wanted a generic title like “Attendance” and my class met twice a week, I would have to fill out this field 30 times for a typical college course.

- [ ] Each event is displayed with a calendar widget and a date/time at top example “Wednesday, August 9 2017, 12:00 AM - 1:00 AM”. The first time I used the form, I thought this was the Title for the assignment, and didn’t realize that the assignments had not yet persisted.

 - [ ] I would expect this date information to be presented like this beyond this page, but it is never shown this way again to the faculty or student user, (it shows up partially where the start and end dates are shown). This make me wonder if this should be the default title for the assignment, which could then be replaced with something more specific on this page.

dovek:
@jonathoy I think this looks really cool. A few suggestions though

1. On the mass_edit view, there is only a submit button. It probably doesn't matter much, but if a user goes there by accident, maybe a cancel button would be nice.

2. I found an error for validation. To replicate it, do the following:

A. Create some attendance events (not sure if this is necessary).
B. Create new attendance event.
C. Make the open date after the close date and submit.

The error I received is: NoMethodError at /attendance
undefined method `valid?' for nil:NilClass

And occurs at line 9 of app/views/layouts/_alerts.haml
```
- if !request.get? && local_assigns.key?(:model) && !model.valid?
```

3. When configuring for the first time, I see that you can have different start and end times. Do these really matter? I imagine if you are taking attendance it is for that day. IF you find that time is important, do we maybe want to make them auto-fill. For instance, if I'm instructor for a k-12 class I will have to put in 8am to 3pm 5 times. I think it would be cool if I only had to put it in once and have it auto-fill to Friday.

4. Maybe make "Start Date" and "End Date" on the configuration view "Course Start/End Date" to be more clear.

5. I found a weird error. When I create a Tuesday/Thursday course from January 22nd to February 9th starting at 8am and ending at 10am (6 attendance days altogether), upon submission only 5 show up in the list in the attendance view. If I refresh the list, all 6 are there. Strange?

app/assets/javascripts/angular/directives/attendance/event_attribute_card.coffee
```diff
@@ -0,0 +1,39 @@
...
+      event: '='
+      eventIndex: '='
+    restrict: 'EA'
+    templateUrl: 'attendance/_event_attribute_card.html'
+    link: (scope, el, attr) ->
+      scope.formErrors = []
+
+      scope.queuePostAttendanceEvent = () ->
+        AttendanceService.queuePostAttendanceEvent(@event)
+        $timeout(
```
acousticrobot: Maybe just a comment here that states "clear out error messages" or something like that.

app/assets/javascripts/angular/services/AttendanceService.coffee
```diff
@@ -0,0 +1,146 @@
...
+        GradeCraftAPI.logResponse(response)
+      , (response) ->
+        GradeCraftAPI.logResponse(response)
+    )
+
+  editAttendanceEvent = (attendanceEvent) ->
+    window.location.href = "/assignments/#{attendanceEvent.id}/edit"
+
+  deleteAttendanceEvent = (attendanceEvent) ->
+    return GradeCraftAPI.deleteItem(events, attendanceEvent) if !attendanceEvent.id?
```
acousticrobot: GradeCraftAPI.deleteItem requires an id to work, this line should be replace with something like:

`return _.remove(events, {name: attendanceEvent.name}) if !attendanceEvent.id?`

jonathoy:
I've addressed the main concerns documented here with the exception of some of the UX-related bullet points. This can be reviewed one final time or merged once #3557 is in master.

chcholman:
@jonathoy This looks fantastic, I'm so excited for this to come in. The one thing I'd change - no need to require instructors to complete the description for each assignment. That means that when the events are created initially, they're valid and so should be marked as green.

jonathoy:
@chcholman Currently the events are marked green to signify that they have been persisted. The current workflow is that there is a button to go back to the setup and make changes if you have not yet persisted any events. The user then has to hit the final submit button to save all changes.

Would you prefer that events are persisted as soon as they fill in and submit the initial setup form? I feel that this would result in some unnecessary creates and destroys.

PR #3490, opened 2017-07-18
3450 ng conditions

### LINKED ISSUE #3436: Error setting the date on the unlock conditions
Error observed when trying to set the date attached to an unlock condition:

screen shot 2017-07-05 at 10 00 18 am

Visible on the badges form

### LINKED ISSUE #3450: angularize the unlocks widget
There are a number of errors to fix in the unlocks widget present on the Assignment, Badge, and GSE forms.  I have proposed that these widgets be rebuilt in Angular, so that we have more control over the logic gates that control them, and so they don't clobber each other when there is more than one per item.

#### Limitations:

We will have to have an id for the object we are assigning the Condition to. To start we can rebuild the widget on a separate page for an existing item.  It will be transferable to the edit pages, but we will have to decide if we want to make the create form a two step process, or if there is a better solution for the `/new` pages.

### LINKED ISSUE #3489: Add assignment last updated at date to assignment settings page

acousticrobot:
### Status
**IN DEVELOPMENT**

### Description

  * Adds Unlock Condition as angular directive
  * Adds Assignment new and edit forms as Angular directive
  * Adds Badge new and edit forms as Angular directive
  * Adds the Unlock Condition portion to the Grade Scheme Elements edit page

### Todos
- [x] Assignment Updated At
- [x] Assignment Submit Button
- [x] Assignment with grades*
- [x] Badges form
- [x] Assignment Types conditions

*In api/assignments/update, we need to check for changes in important fields, and update grades when in these instances.

### Migrations
NO

 * Test create and edit an Assignment, and add unlock conditions
 * Test create and edit a Badge, and add unlock conditions

 Assignment, Badges, Unlocks

======================
Closes #3450 
Closes #3489
Closes #3436


jonathoy:
On the badge edit page, if I try to add an unlock condition for assignment type I see `assignmentType` as the label.

screen shot 2017-07-27 at 9 59 16 am

app/assets/javascripts/angular/directives/assignments/edit_basics.coffee
```diff
@@ -0,0 +1,16 @@
+@gradecraft.directive 'assignmentEditBasics', ['AssignmentTypeService', 'AssignmentService', (AssignmentTypeService, AssignmentService) ->
```
acousticrobot: If this doesn't get converted into React and we find a refactor useful, I would break the form up into more generic inputs, similarly to how I handled the booleans on the `settings` page. For now though, I think we should get this in as flat as possible.

app/assets/javascripts/angular/directives/assignments/edit_unlocks.coffee
```diff
@@ -0,0 +1,17 @@
...
+    }
+    templateUrl: 'assignments/edit_unlocks.html',
+    link: (scope, el, attr, ngModelCtrl)->
+      scope.termFor = AssignmentService.termFor
+
+      scope.updateAssignment = ()->
+        AssignmentService.queueUpdateAssignment(@assignment.id)
+
+      scope.hasUnlocks = ()->
+        UnlockConditionService.unlockConditions.length
```
acousticrobot: Yes, since 0 returns false in js it's a convention for checking if an array has any elements.

app/assets/javascripts/angular/directives/grade_scheme_elements/edit_unlocks.coffee
```diff
@@ -0,0 +1,10 @@
+@gradecraft.directive 'gradeSchemeElementEditUnlocks', ['GradeSchemeElementsService', 'UnlockConditionService', (GradeSchemeElementsService, UnlockConditionService) ->
+
+  return {
+    scope: {
+      gradeSchemeElementId: "="
+    }
+    templateUrl: 'grade_scheme_elements/edit_unlocks.html',
+    link: (scope, el, attr, ngModelCtrl)->
```
acousticrobot: 👍

app/assets/javascripts/angular/directives/grade_scheme_elements/grade_scheme_element_edit.coffee
```diff
@@ -0,0 +1,15 @@
+# Entry point for editing an grade_scheme_element
+@gradecraft.directive 'gradeSchemeElementEdit', [() ->
+  GradeSchemeElementEditCtrl = [()->
```
acousticrobot: Good catch here, I started adding a full edit form for the GSE, and then realized that it wouldn't be good to allow points editing outside of checks on all the other GSEs. I can remove this directive and template entirely.

app/assets/javascripts/angular/directives/assignments/edit_unlocks.coffee
jonathoy: Yep. I think this works for your case. I personally am extra wary about type coercion in JS. If you had done something like `if (UnlockConditionService.unlockConditions.length == true)` in Coffeescript and length was 1, it would evaluate to false since Coffeescript treats == as ===.

app/assets/javascripts/angular/directives/unlock_conditions/unlock_conditions.coffee
```diff
@@ -0,0 +1,27 @@
+@gradecraft.directive 'unlockConditions', ['UnlockConditionService', (UnlockConditionService) ->
+  unlockConditionsCtrl = [()->
+    @loading = true
```
acousticrobot: I'm good with that, if for no other reason than keeping the code consistent, since there are ~20 other files we have set up using the `vm = this` pattern.

app/controllers/api/badge_files_controller.rb
```diff
@@ -0,0 +1,46 @@
...
+    file = BadgeFile.where(id: params[:id]).first
+    if file.present?
+      file.delete_from_s3
+      file.destroy
+
+      if !file.exists_on_s3? && file.destroyed?
+        render json: { message: "Badge file successfully deleted", success: true },
+        status: 200
+      elsif file.destroyed?
+        render json: {message: "Badge file deleted, error removing remote file", success: true},
```
acousticrobot: Mike set up this pattern initially when he redid our s3 uploads. I've wondered before about the frequency of these errors and what we actually would catch in this condition chain. Feel free to make a issue to examine it further if you want, there are three files uploaders that use this approach which should all use the same pattern. (I guess students can't delete submission files?)

app/controllers/api/unlock_conditions_controller.rb
```diff
@@ -0,0 +1,72 @@
+class API::UnlockConditionsController < ApplicationController
+  before_action :ensure_staff?
+  before_action :use_current_course
+
+  # GET /api/assignments/:assignment_id/unlock_conditions
```
acousticrobot: I believe it is because they are nested in the `assignment` resource, so the id becomes explicit for the outside model. Here is another example: 
`/api/assignments/:assignment_id/criteria`. 

This pattern comes from:
```
resources :assignments
  resources :unlock_conditions, only: :index
```

if it was:
```
resources :assignments
  resources :unlock_conditions, only: [:index, :show]
```
We would also have `/api/assignments/:assignment_id/unlock_conditions/:id`
The generic `:id` param is assigned to the innermost resource.

app/controllers/api/badge_files_controller.rb
chcholman: Students should definitely be able to delete submission files - if that isn't in place we need to investigate in a different ticket.

app/assets/javascripts/angular/templates/unlock_conditions/unlock_condition.html.haml
```diff
@@ -0,0 +1,52 @@
+.unlock-condition-requirement(ng-class='{"valid" : conditionIsValid(condition)}')
+  %label Requirement
+  %select(ng-model="condition.condition_type" name="condition_type" ng-change="changeConditionType(condition)")
+    %option(ng-repeat="type in conditionTypes" ng-selected="conditionsTypeTranslation(type) == termFor(condition.condition_type)" value="{{conditionsTypeTranslation(type)}}")
```
chcholman: The options here should be title-cased (currently it's showing "Quest type" rather than "Quest Type"

chcholman: "Badges" should not appear as an option if the course doesn't have badges turned on

app/assets/javascripts/angular/templates/unlock_conditions/unlock_condition.html.haml
```diff
@@ -0,0 +1,52 @@
...
+.unlock-condition-parameters(ng-if="condition.condition_type=='AssignmentType'")
+  %label {{ termFor("assignment_type") }}
+  %select(ng-model="condition.condition_id" name="condition_id" ng-change="updateCondition()")
+    %option(ng-repeat="assignmentType in assignmentTypes" ng-selected="assignmentType.id == condition.condition_id" value="{{assignmentType.id}}")
+      {{assignmentType.name}}
+  %label State Achieved
+  %select(ng-model="condition.condition_state" ng-change="updateCondition()")
+    %option(ng-repeat="state in assignmentTypeStates"  ng-selected="state == condition.condition_state" value="{{state}}")
+      {{state}}
+  %label Number (Points or termFor(""))
```
chcholman: This is currently rendering exactly and not being translated

app/assets/javascripts/angular/templates/unlock_conditions/unlock_condition.html.haml
```diff
@@ -0,0 +1,52 @@
...
+  %input(ng-model="condition.condition_value" gc-number-input ng-change="updateCondition()")
+  %label Completed By Date
```
chcholman: I can't select a date - getting an "Uncaught Missing instance data for this datepicker" message in the console

app/assets/javascripts/angular/templates/badges/edit_basics.html.haml
```diff
@@ -0,0 +1,22 @@
+%section.form-section
+  .form-subsection
+    .form-item
+      %label
+        {{termFor("badge")}} Name
```
chcholman: Needs to be title cased

chcholman:
If I add an unlock condition, go back to the main tab, and go back to the unlock conditions tab, I've suddenly got two of the condition I created. If I delete one of the multiple, it deletes all of them.

app/assets/javascripts/angular/templates/unlock_conditions/unlock_condition.html.haml
acousticrobot: :+1 Both fixed.

app/assets/javascripts/angular/templates/badges/edit_basics.html.haml
acousticrobot: fixed

chcholman:
Awesome, glad that worked! I'm still hitting the date issue, any ideas?

PR #3493, opened 2017-07-20
Grading Status Updates

### LINKED ISSUE #3488: Cross-assignment list of "Ungraded Submissions"
- Sorted by recency, add assignment column to table
- Keep search utility

chcholman:
### Status
**READY**

### Description
This PR combines all grades and submissions into a single table for each section, rather than a table for each assignment. 

### Migrations
NO

* Grading Status page

======================
Closes #3488

### DIFF
--- a/app/controllers/grades_controller.rb
+++ b/app/controllers/grades_controller.rb
@@ -97,6 +97,24 @@ class GradesController < ApplicationController
       notice: "Thank you for letting us know!"
   end
 
+  # PUT /grades/:ids/release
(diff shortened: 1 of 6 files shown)


app/controllers/grades_controller.rb
```diff
@@ -97,6 +97,24 @@ def feedback_read
       notice: "Thank you for letting us know!"
   end
 
+  # PUT /grades/:ids/release
```
jonathoy: Nitpicky comment - I think this route is actually `/grades/release`

PR #3495, opened 2017-07-20
3349 import earned badges

### LINKED ISSUE #3349: Import earned badges
### Feature
As an instructor, I want to be able to import earned badges, just the way we import grades. The pattern to match here is that of pass/fail import. If a student is marked with a [1], then they earn the badge on import. Otherwise, no. With a grade, if a user has a score then we override it on the import. Here, we're going to do the opposite - if it's possible to earn the badge (some badges can only be awarded once), then the student earns it *each time* the instructor imports a spreadsheet with a 1 by their name. This also impacts the structure of the sample file for download - *don't* display a 1 for students who have earned it (we might consider adding a column to the file that we don't import that displays the count of time the student has earned the badge). 

Tasks: 
- [ ] Create import
- [ ] Add tests
- [ ] Create sample file for download to give instructors the right pattern to work off of

dovek:
### Status
**IN DEVELOPMENT**

### Description
As an instructor, I want to be able to import earned badges, just the way we import grades. The pattern to match here is that of pass/fail import. If a student is marked with a [1], then they earn the badge on import. Otherwise, no. With a grade, if a user has a score then we override it on the import. Here, we're going to do the opposite - if it's possible to earn the badge (some badges can only be awarded once), then the student earns it each time the instructor imports a spreadsheet with a 1 by their name. This also impacts the structure of the sample file for download - don't display a 1 for students who have earned it (we might consider adding a column to the file that we don't import that displays the count of time the student has earned the badge).

### Related PRs
N/A

### Todos
- [x] Add Tests

### Deploy Notes
N/A

### Gem dependencies
N/A

### Migrations
NO

In Progress

* Badges

======================
Closes #3349


app/importers/badge_importers/csv_badge_importer.rb
```diff
@@ -0,0 +1,141 @@
...
+        end
+
+        earned_badge = @current_course.earned_badges.where(student_id: student.id, badge_id: badge.id).first
+        earned_badge_count = @current_course.earned_badges.where(student_id: student.id, badge_id: badge.id).count
+
+        if Integer(row.earned) == earned_badge_count
+          unchanged << earned_badge
+        end
+
+        while earned_badge_count < Integer(row.earned)
```
dovek: @chcholman and I discussed this during development and decided that we could do a loop, so for example: if current awarded count is 3 and the new awarded amount is 5, the logic should loop through and award the badge 2 more times (assuming that the badge is multi-awardable).

chcholman:
@dovek This functions perfectly. Let's brainstorm ways to make the outcomes/response page a little bit more intuitive for users when you get back into the office on Wednesday, but otherwise this is ready to go!

chcholman:
The UX notes we discussed earlier this morning: 

- [ ] Don't show 'nil' imports at all - if there's nothing imported for the student, that should be neither an error nor a success
- [ ] Break the components of the badge import errors out into separate columns
- [ ] Remove the email column from the success and add the feedback (I realized that the number makes sense, because in this display each badge has it's own row) 

Thanks @dovek!

chcholman:
@dovek So much closer! I just reviewed this with Evan, and she suggested two more changes: 
- [ ] switch the column "Current Earned Count" to the very end
- [ ] The "New Awarded Count" should NOT include the count already earned - whatever the instructor enters here is the number that should be awarded on import

Happy to talk through if any of this doesn't make sense!

chcholman:
Thank you @dovek!

PR #3496, opened 2017-07-20
2765 papertrail

### LINKED ISSUE #2765: Switch to PaperTrail for logging
Since [PaperTrail](http://papertrailapp.com) has significantly better pricing plans, we could save money by switching to them

jwright:
### Status
**READY**

### Description
This changes the system and job logger to use [Papertrail](https://papertrailapp.com) instead of [Loggly](https://loggly.com).

In order to log the events in the event loggers, without a large refactor, is to replace `LogglyResque` with a new module `PapertrailResque`.

Of note is the fact that the Papertrail logger does not handle [QuietAssets](https://github.com/rails/sprockets-rails/pull/355) in Sprockets. This is because the logger does not respond to the `#silence` method [which is used when the assets are quieted](https://github.com/rails/sprockets-rails/issues/376). Several attempts were made to include the [`LoggerSilence` module](https://github.com/rails/rails/blob/master/activesupport/lib/active_support/logger_silence.rb) but it was not straight forward as the `RemoteSyslogLogger` is a module and not a class.

This should not be an issue as Papertrail is only used on `staging`, `beta`, `production-testing`, and `production` environments where the assets are not quieted. If it becomes an issue in the future, I believe we can implement the silence method.

### Todos
- [x] Replace `LogglyResque`
- [x] Ensure that the correct tags are being logged for jobs
- [x] Remove loggly references

### Deploy Notes
There is further need for the `LOGGLY_TOKEN` environment variable. That can be removed after deployment.

Gradecraft owns the Papertrail account and the credentials can be retrieved from @chcholman if needed.

I would recommend putting this on staging first and testing. I tested this locally on development to push to Papertrail temporarily.

### Gem dependencies
- REMOVED [logglier](https://github.com/freeformz/logglier)
- ADDED [remote_syslog_logger](https://github.com/papertrail/remote_syslog_logger)

### Migrations
NO

```sh
git pull --prune
git checkout 2765-papertrail
bundle; script/server
```

1. Visit https://papertrailapp.com and login
1. Visit a page in GradeCraft and ensure that the activity is being logged with the appropriate format
1. Ensure that the job information is being logged (such as `pageview_event_log`) is being logged with the appropriate format

* Logging

======================
Closes #2765


jwright:
@jonathoy I changed the `config/environments/development.rb` file and added basically the same settings for `config.logger` and checked Papertrail to ensure the logs were being sent there.

For jobs, I had to create the `PaperTrailResque` logger and then enqueued the `PageView` event right away and saw the data there.

PR #3522, opened 2017-07-28
run grade jobs only on submit

### LINKED ISSUE #3519: Grade Notifications
### Expected Behavior
Either a or b:

Editing a Grade that has already been released to a student should either send out a notification:
a) only on hitting submit
b) not at all.

### Actual Behavior

Every autosave to a visible grade sends a new notification to the student

acousticrobot:
### Status
**READY**

### Description

This PR limits the notifications sent when changes are made to a student visible grade.

If a faculty member alters a student's grade once it has been released, we need to limit notifications and updates for the autosave feature. But when they click submit, the background processes will still fire, cashing scores and sending email notifications.

#### 1:  The faculty grades an Assignment for student. Until the grade is submitted and released, no notifications are sent.

screen shot 2017-07-28 at 12 57 52 pm

#### 2. As soon as the grade is submitted as `student_visible`, a notification is sent to the student that the assignment has been graded.

screen shot 2017-07-28 at 12 58 18 pm

#### 3. Faculty goes back and alters the student's released grade. Incremental updates are autosaved (see in console below) but the student does not receive new notifications.

screen shot 2017-07-28 at 12 59 29 pm

#### 4. When the faculty hits the submit button, the student receives a new notification.

screen shot 2017-07-28 at 12 59 59 pm

### Limitations to this approach:

Faculty really shouldn't be altering grades that are student visible, but we probably need to continue to allow for this. 

In this solution, if a faculty member alters a released grade multiple times, we will still end up with a student receiving many notifications.

If faculty rely on the autosave feature and back out of the form without hitting submit, no background process will be run.

One alternative would be for faculty to manually trigger notifications for an email, something that 99% of the time would be an additional and unnecessary step. 

Another alternative would be to only send an email when the grade is first released, but then students would not be informed if a grade is changed, which I think is something they should know.

One feature improvement might be to put a notification into the form that warns faculty they are about to change a grade that has already been released to the student.

======================
Closes #3519

### DIFF
--- a/app/controllers/api/grades_controller.rb
+++ b/app/controllers/api/grades_controller.rb
@@ -19,11 +19,12 @@ class API::GradesController < ApplicationController
   # POST api/grades/:grade_id
   def update
     grade = Grade.find(params[:id])
-
-    # Only send notifications etc. when handling the form submission
-    run_jobs = params[:submit]
-    result = Services::UpdatesGrade.update grade, grade_params, current_user.id, run_jobs
+    result = Services::UpdatesGrade.update grade, grade_params, current_user.id
     if result
+      # Only send notifications etc. when handling the form submission
+      if params[:submit] && grade.student_visible?
+        GradeUpdaterJob.new(grade_id: grade.id).enqueue
+      end
       @grade = result.grade
       render "api/grades/show", success: true, status: 200
     else
--- a/app/controllers/api/grades_controller.rb
+++ b/app/controllers/api/grades_controller.rb
@@ -19,7 +19,10 @@ class API::GradesController < ApplicationController
   # POST api/grades/:grade_id
   def update
     grade = Grade.find(params[:id])
-    result = Services::UpdatesGrade.update grade, grade_params, current_user.id
+
+    # Only send notifications etc. when handling the form submission
+    run_jobs = params[:submit]
+    result = Services::UpdatesGrade.update grade, grade_params, current_user.id, run_jobs
     if result
       @grade = result.grade
       render "api/grades/show", success: true, status: 200
--- a/app/services/creates_grade.rb
+++ b/app/services/creates_grade.rb
@@ -12,8 +12,8 @@ module Services
 
     aliases raw_params: :attributes
 
-    def self.create(raw_params, graded_by_id, run_jobs)
-      with(raw_params: raw_params, graded_by_id: graded_by_id, run_jobs: true)
+    def self.create(raw_params, graded_by_id)
+      with(raw_params: raw_params, graded_by_id: graded_by_id)
         .reduce(
           Actions::VerifiesAssignmentStudent,
           Actions::BuildsGrade,
--- a/app/services/creates_grade.rb
+++ b/app/services/creates_grade.rb
@@ -12,8 +12,8 @@ module Services
 
     aliases raw_params: :attributes
 
-    def self.create(raw_params, graded_by_id)
-      with(raw_params: raw_params, graded_by_id: graded_by_id)
+    def self.create(raw_params, graded_by_id, run_jobs)
+      with(raw_params: raw_params, graded_by_id: graded_by_id, run_jobs: true)
         .reduce(
           Actions::VerifiesAssignmentStudent,
           Actions::BuildsGrade,
--- a/app/services/creates_grade/runs_grade_updater_job.rb
+++ b/app/services/creates_grade/runs_grade_updater_job.rb
@@ -4,11 +4,10 @@ module Services
       extend LightService::Action
 
       expects :grade
-      expects :run_jobs
 
       executed do |context|
         grade = context[:grade]
-        if grade.student_visible? && context[:run_jobs]
+        if grade.student_visible?
           grade_updater_job = GradeUpdaterJob.new(grade_id: grade.id)
           grade_updater_job.enqueue
         end
--- a/app/services/creates_grade/runs_grade_updater_job.rb
+++ b/app/services/creates_grade/runs_grade_updater_job.rb
@@ -4,11 +4,11 @@ module Services
       extend LightService::Action
 
       expects :grade
+      expects :run_jobs
 
       executed do |context|
         grade = context[:grade]
-
-        if grade.student_visible?
+        if grade.student_visible? && context[:run_jobs]
           grade_updater_job = GradeUpdaterJob.new(grade_id: grade.id)
           grade_updater_job.enqueue
         end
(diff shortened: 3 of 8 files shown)


app/controllers/api/grades_controller.rb
```diff
@@ -19,7 +19,10 @@ def show
   # POST api/grades/:grade_id
   def update
     grade = Grade.find(params[:id])
-    result = Services::UpdatesGrade.update grade, grade_params, current_user.id
+
+    # Only send notifications etc. when handling the form submission
+    run_jobs = params[:submit]
```
jwright: I like the rename here to make it more intention-revealing.

app/services/creates_grade.rb
```diff
@@ -12,8 +12,8 @@ class CreatesGrade
 
     aliases raw_params: :attributes
 
-    def self.create(raw_params, graded_by_id)
-      with(raw_params: raw_params, graded_by_id: graded_by_id)
+    def self.create(raw_params, graded_by_id, run_jobs)
+      with(raw_params: raw_params, graded_by_id: graded_by_id, run_jobs: true)
```
jwright: UI default is `false` here for `run_jobs` but backend is `true`. Should we make these consistent?

app/services/creates_grade/runs_grade_updater_job.rb
```diff
@@ -4,11 +4,11 @@ class RunsGradeUpdaterJob
       extend LightService::Action
 
       expects :grade
+      expects :run_jobs
 
       executed do |context|
         grade = context[:grade]
-
-        if grade.student_visible?
+        if grade.student_visible? && context[:run_jobs]
```
jwright: I don't know if we should do this in this PR but I feel we should use the `GradeProctor.viewable?` here instead of just checking for `student_visible`.

app/services/creates_grade_using_rubric.rb
```diff
@@ -18,7 +18,7 @@ class CreatesGradeUsingRubric
     aliases raw_params: :attributes
 
     def self.create(raw_params, graded_by_id)
-      with(raw_params: raw_params, graded_by_id: graded_by_id)
+      with(raw_params: raw_params, graded_by_id: graded_by_id, run_jobs: true)
```
jwright: Same comment re: default as above.

spec/services/creates_grade/runs_grade_updater_job_spec.rb
```diff
@@ -1,12 +1,27 @@
...
-  let(:context) { { grade: grade } }
+  let(:context) { { grade: grade, run_jobs: true } }
 
   it "expects grade passed to service" do
-    expect { described_class.execute({ })}.to \
+    expect { described_class.execute({ run_jobs: true })}.to \
       raise_error LightService::ExpectedKeysNotInContextError
   end
 
+  it "expects a confirmation boolean" do
```
jwright: I think this is description is not descriptive enough. I feel like we should say `it "expects a boolean to run the update jobs"` or something similar.

spec/services/creates_grade/runs_grade_updater_job_spec.rb
```diff
@@ -1,12 +1,27 @@
...
+    expect { described_class.execute({ grade: grade })}.to \
+      raise_error LightService::ExpectedKeysNotInContextError
+  end
+
+  context "wihout confirmation" do
```
jwright: `context "without confirmation to run the update jobs" do`?

app/services/creates_grade/runs_grade_updater_job.rb
acousticrobot: Interesting, the GradeProctor is something we might need to revisit. `student_visible` is now a field on Grade and not just a method, I removed the release logic that used to live on `Assignment`. This greatly simplifies the permissions but I think the Proctor is still useful since it also checks in the context of the `current_user`.

In this case I think we want to run this based directly on the state of boolean attribute, as the grade is always visible to the current user who is a grader, but we only want to send notifications if the grade is visible to the student. It seems circuitous to me to check `GradeProctor.new(grade).viewable?(grade.student)` when what we really want is the `grade.student_visible` field. What do you think?

jwright: Ok. If this has changed, I apologize. 

The original concept for the proctors was to have all authorization logic go through those instead of having it spread all throughout the application in different ways (a problem I remember we had and bugs coming from). The proctor contains different logic depending on who was trying to view the grade.

I am fine with whatever since I have not been attached to the project lately but I hope we just don't have authorization logic spread throughout the application again in different ways. I agree it is more intention revealing.

app/services/creates_grade.rb
acousticrobot: Yeah I was having a hard time with this. I had a lot of unintentional failures without default true on the ruby side. The biggest problem is we have services that iterate over other services that completely obscure the calls to the background jobs.  But then on the front end, I really don't want the API to default to running background jobs since there is potentially 20 or thirty AJAX calls where it should be `false`, followed by one explicit `true`.  Ideally this boolean could be handled in `Services::UpdatesGrade` and not passed all the way down to `Actions::RunsGradeUpdaterJob` but I don't think it's possible to run a step conditionally. This is one of the things I much prefer about `Porch` over `LightService` btw.

What do you think about separating the background job out of the service, and then running `RunsGradeUpdaterJob` through a second service call on submit true? Then these changes wouldn't affect the services in the rest of the app. I was initially hesitant to have two services in one controller action, but it might actually be cleaner.

spec/services/creates_grade/runs_grade_updater_job_spec.rb
acousticrobot: I'll change this if we decide to keep this boolean in the Action.

spec/services/creates_grade/runs_grade_updater_job_spec.rb
acousticrobot: I'll change this if we decide to keep this boolean in the Action.

app/services/creates_grade.rb
jwright: I feel like right now, I think this little inconsistency is fine. I did not realize all the places this is used and therefore, the inconsistency is fine since it won't be consistent everywhere.

acousticrobot: 👍

app/services/creates_grade/runs_grade_updater_job.rb
acousticrobot: OK cool, since this is not actually authorization logic, but about background processes that need to run on released grades (sending emails and updating cached scores) I'll leave it as is, but I agree with you that viewing permissions should continue to run through proctors.

chcholman:
@acousticrobot Interestingly, this successfully held back the grade notification, but the earned badge announcements associated with the regrade were sent immediately. Do you want to handle that here or in a different ticket?

PR #3528, opened 2017-07-31
update styling for student panel pages

### LINKED ISSUE #3510: Add ability to scroll sidebar content on grade predictor and badge page
### Expected Behavior
As a student, I expect to be able to view the entire description of my selected assignment or badge on the grade predictor and badge page, not a truncated version that may be covered by other page elements.

mariehooper:
### Status
**READY**

### Description
This PR removes the description length constraint from student side panel in the predictor and badge page and adds ability to scroll content. 

### Migrations
NO

======================
Closes #3510

### DIFF
--- a/app/assets/javascripts/angular/templates/badges/student_index.html.haml
+++ b/app/assets/javascripts/angular/templates/badges/student_index.html.haml
@@ -1,7 +1,5 @@
 .student-panel-wrap
-  %student-panel.student-panel.badge-panel{'context'=>'"badges"'}
-
-  #student-main-column
+  #student-main-column.student-flex-col
 
     %loading-message{'loading'=>'vm.loading', 'message'=>"Loading Badges..."}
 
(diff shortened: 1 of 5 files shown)


mariehooper:
@chcholman - I have another branch that implements independent scrolling, but it doesn't feel seamless in this app since we have sticky footer implemented. It becomes 3 different scroll bars then rather than simply one for each column. I almost would prefer to consider handling the assignment descriptions in modals rather than the sidebar design. In user testing, we have found that students sometimes don't realize content is even changing as they click around. This addresses Jaee's heuristic eval, but wanted your opinion since it's not currently a great solution.

chcholman:
@mariehooper I agree that this isn't ideal, but I actually think this is about the best version of it that I've seen. I like your idea of exploring modals to handle this, but am going to merge this in the meantime. Thank you!

PR #3565, opened 2017-08-16
3547 weights regrade

### LINKED ISSUE #3547: Assignment Weighting in the Grade Predictor not updating actual grades
As a student, if I update my weighting choices in the Grade Predictor, I expect that my grades will be updated to reflect this. These updates are currently not being applied.

acousticrobot:
### Status
**READY**

### Description
  * This PR adds updating to existing grades and course score from the controller when a student changes their weights on an AssignmentType. 
  * It adds a description of the callback stack that operates when a course score is updated, so we can think about how to refactor this.
  * It changes some `caching_...` method names to `update_...` since there is no caching in these methods
  * It fixes a calculation bug in the predictor that was multiplying weights on earned points twice.

======================
Closes #3547

### DIFF
--- a/app/assets/javascripts/angular/services/PredictorService.coffee
+++ b/app/assets/javascripts/angular/services/PredictorService.coffee
@@ -57,8 +57,7 @@
       subset = assignmentsForAssignmentType(assignments,assignmentType.id)
       total = AssignmentService.assignmentsSubsetPredictedPoints(subset)
     else
-      # Use weighted calculation sent from API
-      total = weightedEarnedPoints(assignmentType)
+      total = assignmentType.final_points_for_student
 
     if includeWeights
       total = weightedPoints(assignmentType, total)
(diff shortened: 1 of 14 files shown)


chcholman:
screen shot 2017-11-05 at 2 20 52 pm

- [ ] Something's up with the name at the top of the section?
- [ ] I think the number placement may be offcenter by a couple of pixels?

chcholman:
If I clear out all weights, the grade (no longer included) still stays green. The reverse is also true - if I weight an assignment type, the grades stay greyed out until I refresh. 
screen shot 2017-11-05 at 2 22 59 pm

chcholman:
This looks awesome @acousticrobot - I've got 2 really small design bugs noted above and one slightly larger issue (if you want to address the color change in a separate PR that's okay by me), but otherwise this should get merged asap, we've got a class that is using this feature and we need to help them get scores straightened out.

chcholman:
@acousticrobot 👍

PR #3593, opened 2017-08-27
Fix for submissions editing and the tab redirect

chcholman:
### Status
**In Development**

### Description
This PR fixes the editing of submissions and redirects users to the proper tab


chcholman:
Hey @jonathoy can you investigate this first thing Monday morning? Editing a submission currently produces an error, and while commenting out the reference to Deleting the Draft gets it to go through I'm hesitant to do that without knowing the exact repercussions of that

PR #3617, opened 2017-08-30
3586 Fix Google Rate Limit Issue

### LINKED ISSUE #3586: Google Rate Limit Exceeded
Users are attempting to add so many assignments at once that the API's rate limit is exceeded. We need to either implement batching or a backoff strategy. https://developers.google.com/drive/v3/web/batch

dovek:
### Status
**IN DEVELOPMENT**

### Description
Users are attempting to add so many assignments at once that the API's rate limit is exceeded. We need to either implement batching or a backoff strategy. 

### Related PRs
N/A

### Todos
- [ ] Add Tests

### Deploy Notes
N/A

### Gem dependencies
N/A

### Migrations
NO

On the Dashboard, if a user spams the "Add All Assignments to Calendar" link

1. The link text should change to "Adding Assignments to your Calendar..."

2. The button, although it will seem active will be disabled, so no matter how many times a user hits it, it will only make the API call once.

* Dashboard
* Assignments
* Google Calendar

======================
Closes #3586

### DIFF
--- a/app/controllers/google_calendars/assignments_controller.rb
+++ b/app/controllers/google_calendars/assignments_controller.rb
@@ -28,7 +28,7 @@ class GoogleCalendars::AssignmentsController < ApplicationController
   def add_assignments
     assignment_list = get_all_items_for_current_course(current_course, "assignment", current_user)
     assignment_list_filtered = filter_items_with_no_end_date(assignment_list)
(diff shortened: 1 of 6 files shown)


chcholman:
This is awesome, and I found the same as you noted - it felt much faster!

PR #3626, opened 2017-08-31
working submission form

chcholman:
### Status
**IN DEVELOPMENT**

### Description
This addresses a bug where: 
- for assignments that ONLY allow attachments
- if a student is editing (not creating) one
- the update submission button sends them to a page not found error

### DIFF
--- a/app/views/submissions/_autosave_form.html.haml
+++ b/app/views/submissions/_autosave_form.html.haml
@@ -1,7 +1,7 @@
 = render partial: "assignment_guidelines", locals: { assignment: presenter.assignment }
 
 %student-submission{class: "ng-cloak"}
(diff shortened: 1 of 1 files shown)


chcholman:
@jonathoy or @acousticrobot Do either of you have a moment to take a look at this? We've got an ongoing bug where if the assignment *only* allows attachments to be uploaded (text comment and links are turned off) the submission id isn't properly being passed into the Angular controller ONLY in the `Edit` condition. Removing the call to the angular form in the form setup fixes this, but I am trying not to break other functionality. Thoughts about the best way to fix?

chcholman:
@jonathoy that works perfectly! Thank you so much

PR #3633, opened 2017-09-05
Automatically fetch additional pages

### LINKED ISSUE #3632: Importing students from Canvas - only possible to import first 10 students
As an instructor, when I go to import students from Canvas I can only see and import the first 10 on the roster

jonathoy:
### Status
READY

### Description
This PR addresses the bug where users were not being automatically fetched from Canvas when there were additional pages.

### Migrations
NO

Canvas user import

======================
Closes #3632


jonathoy:
@chcholman Was able to test this locally, albeit with some weirdness, and I appear to be able to load 10+ users. They should still be batched according to the default size of 10 per page/request.

chcholman:
@jonathoy Awesome, thanks so much for the quick fix!

PR #3635, opened 2017-09-06
3588 Activate All Students per Course

### LINKED ISSUE #3588: Button to activate all students in a course in bulk for admin users
As an admin user, occasionally we hit a space where students have accounts that need to be activated. We should add a button that allows an admin user to do this from within the interface.

dovek:
### Status
**IN DEVELOPMENT**

### Description
As an admin user, occasionally we hit a space where students have accounts that need to be activated. We should add a button that allows an admin user to do this from within the interface.

### Related PRs
N/A

### Todos
- [x] Add Tests

### Deploy Notes
N/A

### Gem dependencies
N/A

### Migrations
NO

Create a single new user or multiple new users for a single course who has/have not been activated. Navigate to 'My Courses'. In the drop-down options per course, notice the new "Activate All Students" option. Selecting this will loop through the not-yet-activated students in a course and activate their account (not to be confused with activating their course membership). You will notice an alert at the top of the page when the process is complete.

* Courses
* Users

======================
Closes #3588

### DIFF
--- a/app/controllers/courses_controller.rb
+++ b/app/controllers/courses_controller.rb
@@ -139,9 +139,9 @@ class CoursesController < ApplicationController
       student.activate!
     end
     if total != 1
-      redirect_to session[:return_to] || dashboard_path, notice: "#{total} students have been activated!" and return
+      redirect_to session[:return_to] || students_path, notice: "#{total} #{(term_for :student).downcase}(s) have been activated!" and return
     else
-      redirect_to session[:return_to] || dashboard_path, notice: "#{total} student has been activated!" and return
+      redirect_to session[:return_to] || students_path, notice: "#{total} #{t(term_for :student).downcase}(s) has been activated!" and return
     end
   end
 
--- a/app/controllers/courses_controller.rb
+++ b/app/controllers/courses_controller.rb
@@ -18,6 +18,7 @@ class CoursesController < ApplicationController
                                      :recalculate_student_scores]
   skip_before_action :verify_authenticity_token, only: [:change]
   before_action :ensure_not_impersonating?, only: [:change]
+  before_action :save_referer, only: [:activate_all_students]
 
   def index
     @courses = current_user.courses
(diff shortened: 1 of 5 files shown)


app/controllers/courses_controller.rb
```diff
@@ -130,6 +131,20 @@ def unpublish
...
 
+  def activate_all_students
+    course = current_user.courses.where(id: params[:id]).first
+    students = User.accounts_not_activated(course)
+    total = students.count
+    students.each do |student|
+      student.activate!
+    end
+    if total != 1
+      redirect_to session[:return_to] || dashboard_path, notice: "#{total} students have been activated!" and return
```
chcholman: We're only going to do this from the students index page, so we don't need the `session[:return_to] || dashboard_path` component - just sent them back to the students_path

chcholman: Let's reference the `term_for student` in the return notification

spec/controllers/courses_controller_spec.rb
```diff
@@ -184,6 +184,25 @@
...
+      end
+
+      it "redirects to referer url if present" do
+        request.env["HTTP_REFERER"] = "http://some-referer.com"
+        post :activate_all_students, params:{id: course.id}
+        expect(response).to redirect_to("http://some-referer.com")
+      end
+
+      it "redirects to dashboard if referer url is not present" do
+        post :activate_all_students, params:{id: course.id}
```
chcholman: this will change as a result of switching the dashboard_path to students_path

app/controllers/courses_controller.rb
```diff
@@ -130,6 +131,20 @@ def unpublish
...
+    else
+      redirect_to session[:return_to] || dashboard_path, notice: "#{total} student has been activated!" and return
+    end
```
chcholman: same note as above

chcholman:
Just a few small tweaks @dovek and this is ready to go, thanks so much!

PR #3639, opened 2017-09-12
Fix issue caused by condition_value being nil

### LINKED ISSUE #3640: Bugs in unlock conditions
Not entirely sure when this happens but because of the possibility for an `condition_value` to be null on an UnlockCondition, there are potential points of failure. See Rollbar issue [2445](https://rollbar.com/gradecraft/gradecraft-development/items/2445/) as an example.

Also if a custom term has been set for an assignment and as assignment has been set as an unlock requirement, then any subsequent editing of the condition will appear as if the requirement type is Assignment Type.

jonathoy:
### Status
READY

### Description
In response to Rollbar issue [2445](https://rollbar.com/gradecraft/gradecraft-development/items/2445/), it appears that there is a nil comparison failure at line 122 in the UnlockCondition model.

Per the stack trace, line 122 seems to refer to the following line in `#check_assignments_completed_condition`:

```
assignment_completed_count >= condition_value
``` 

As observed in Evan's case of Sophomore Honors course - course 295, the `condition_value` is null.

The simple fix to this is to override the accessor method for `condition_value` such that 0 is returned if null. The better solution is probably to update the field with a migration and a rake task so that the field is non-nullable and defaults to 0.

This PR addresses the former.

### Migrations
NO

======================
Closes #3640

### DIFF
--- a/app/assets/javascripts/angular/templates/unlock_conditions/unlock_condition.html.haml
+++ b/app/assets/javascripts/angular/templates/unlock_conditions/unlock_condition.html.haml
@@ -57,8 +57,12 @@
   %select(ng-model="condition.condition_state" ng-change="updateCondition()")
     %option(ng-repeat="state in assignmentTypeStates"  ng-selected="state == condition.condition_state" value="{{state}}")
       {{state}}
-  %label(ng-if="condition.condition_state == 'Assignments Completed'") Number of {{termFor("assignments")}}
-  %label(ng-if="condition.condition_state != 'Assignments Completed'") Amount
+  %label(ng-if="condition.condition_state == 'Assignments Completed'")
+    Number of {{termFor("assignments")}}
+    %span.required-label(ng-if="condition.condition_value == null") *required
+  %label(ng-if="condition.condition_state != 'Assignments Completed'")
+    Amount
+    %span.required-label(ng-if="condition.condition_value == null") *required
   %input(ng-model="condition.condition_value" gc-number-input ng-change="updateCondition()")
   %label Completed By Date
   %input(id="{{datePickerId(condition)}}" ng-model="condition.condition_date" gc-date-time-input ng-change="updateCondition()")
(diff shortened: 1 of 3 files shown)


chcholman:
@jonathoy This makes a lot of sense as a solution. One reason we're seeing this (essentially bad data) though is because we haven't quite managed to get the necessary validations for the data back into places since the revised form. Could you possibly add a validation to the PR that checks for: 

- Unlock Conditions that are based on Assignment Types
- Both the "Assignments Completed" and "Minimum Points Earned" need to require the user to enter numbers in the field below, and we're currently not alerting them to that or requiring that they do so

jonathoy:
@chcholman Could you clarify what you mean when you say that you would want validations for unlock conditions that are based on assignment types?

chcholman:
@jonathoy If you load up a form where you're adding an unlock condition, the first select form allows you to choose the core condition - Assignment Types are one option, and it's the two that inherit from that that are currently not being checked (if you click through others you'll see various conditions that are getting checked, with the "required" call out right next to them).

jonathoy:
I put in a quick fix to display the required validations, but I think that the overall logic in this form can be simplified greatly by using Angular form validations similar to how they're being used in grade scheme elements. In doing so, it might also help in making this page more accessible by handling some of the aria and required tags better.

Something else that I've noticed on this form that will take a little bit more effort to fix is the fact that the Submit button on the page does not persist all changes. If you're on the badge edit page in the unlock tab for instance, notice that if you quickly change an input value and click on Submit Badge the changes will not be persisted. This is due to the fact that the directives are not currently talking to each other.

jonathoy:
Would you like me to try to address this issue in this PR or separately?

chcholman:
@jonathoy Great catch on the form not catching all of the changes. Yes, let's plan to address this in a separate branch since not validating the data is currently causing issues in production. Just finished testing, and merging now, will create a separate issue next.

PR #3672, opened 2017-09-27
change debounce for unlocks to .75 seconds

### LINKED ISSUE #3653: Unlocks form does not persist when hitting submit
The Submit button on the assignment edit page does not persist all changes. If you're on the badge edit page in the unlock tab for instance, notice that if you quickly change an input value and click on Submit Badge the changes will not be persisted. This is due to the fact that the directives work independently of one another.

Based on how we have been setting up these forms, we may want to check that the Submit button works in all other places that the unlocks form is being used.

acousticrobot:
### Status
**READY**

### Description
Quick bugfix to shorten the debounce time on Unlock Conditions.

======================
Closes #3653

### DIFF
--- a/app/assets/javascripts/angular/helpers/debounceQueue.js.coffee
+++ b/app/assets/javascripts/angular/helpers/debounceQueue.js.coffee
@@ -22,18 +22,29 @@ angular.module('helpers').factory('DebounceQueue', ['$timeout', ($timeout)->
       event.apply(null, args)
     else
       cancelEvent(type, id)
-      queueStore[type + id] = $timeout(() ->
-        event.apply(null, args)
-      , delay)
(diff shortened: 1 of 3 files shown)


jonathoy:
I think this is a great way to address the need that we have for having a submit all feature when our forms become more complicated. We should go over the other places where we use debounce and make sure that there are submit buttons that work as intended. Moving forward, I will try to make sure that we're following this strategy for all autosaved forms.

@acousticrobot I've updated the original ticket describing the bug that was linked to this. Since this PR is intended to close the issue, could you make the change in the UnlocksService as well?

jonathoy:
@acousticrobot Scratch that! I saw the change nested in the second commit. Merging!

PR #3690, opened 2017-10-05
Possible fix to check trial create for environment

### LINKED ISSUE #3689: Possibility for a non-umich user to be created on umich.gradecraft.com
Per Camille, a user with a non-umich email was created for the umich instance. They received an activation email that pointed them to the umich domain and upon activation they were automatically logged in, bypassing Shibboleth.

jonathoy:
### Status
READY

### Description
See issue for background. I traced back how this might be happening and it seems that the user received an activation email that pointed them to the umich domain for activation. The only way I can see this happening is if they were somehow able to get to the create a free trial page while on the umich domain.

I wasn't able to figure out exactly where the origin of this happening might be, but I did find one partial view where it wasn't protected when the environment was production. I added two extra checks to ensure that this isn't possible in the future.

Create free trial

======================
Closes #3689

### DIFF
--- a/app/controllers/user_sessions_controller.rb
+++ b/app/controllers/user_sessions_controller.rb
@@ -1,3 +1,4 @@
+# rubocop:disable AndOr
 class UserSessionsController < ApplicationController
 
   before_action :ensure_staff?, only: [:impersonate_student]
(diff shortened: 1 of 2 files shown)


jonathoy:
@chcholman If you get a chance, could you double-check that this change is sound and if you have any additional thoughts regarding a possible point of origin for this scenario?

chcholman:
I think this makes a lot of sense to try! I will be curious to see what sorts of errors we continue to get around this, my guess is we've got old links in a few places, but this is a great effort to redirect them.

PR #3711, opened 2017-10-12
3563 onboarding

### LINKED ISSUE #3563: Onboarding Partials only when needed
The onboarding slideshow should only be called when it will be presented, and otherwise skipped.
Additionally, there is a listener that fails on the faculty onboarding when it doesn't find the id.

acousticrobot:
### Status
**READY**

### ADDED

   * ability to view manually through "View Tour" button

### Description

Reimplements the student and staff onboarding slideshow with the following improvements:

  * Only loads partials if they are going to be displayed.  
  * Moves the logic for determining if a user has seen the onboarding for a course out of the view and into a method.
  * Fixes a javascript error when a close button wasn't present in the slideshow

======================
Closes #3563

### DIFF
--- a/app/assets/javascripts/angular/helpers/exceptionOverride.js.coffee
+++ b/app/assets/javascripts/angular/helpers/exceptionOverride.js.coffee
@@ -1,5 +1,6 @@
 angular.module('helpers')
   .factory('$exceptionHandler', () ->
+
     (exception, cause) ->
       exception.message += 'Angular Exception: "' + cause + '"'
       throw exception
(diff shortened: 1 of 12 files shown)


chcholman:
Thank you for this fix @acousticrobot! The "View Tour" button on both instructors and students doesn't appear to be working to pull this up on demand (possible that it is, so if it's something you've tested let me know and I'll investigate more), could you possibly take a look?

chcholman:
I don't love this - it feels really strange to have the whole page reload. What are other options we could consider?

chcholman:
Thanks for this @acousticrobot! I'm thinking through the options and will get back to you asap!

PR #3713, opened 2017-10-16
Added has_many relationships to allow for easier querying

chcholman:
### Status
**READY**

### Description
This PR adds a variety of inheritance relationships to make it easier to query for rubric related objects in the Rails console. 

* Assignments, rubrics, criteria, criterion grades, levels, level badges


jonathoy:
@chcholman I think this all makes sense!

PR #3714, opened 2017-10-16
When copying a course, level badges reference the badges in the original course

jonathoy:
### Status
NOT READY

### Description
When copying a course, assignments with rubrics and level badges are copied but level badges refer to the original badge in the original course.

Copy a course that has an assignment with rubrics + level badges. The level badges on the copied course for the copied rubric should be invalid due to the badge ids referencing the wrong badge.

Course copy

======================
Closes #[ISSUE NUMBER]

### DIFF
--- a/app/controllers/courses_controller.rb
+++ b/app/controllers/courses_controller.rb
@@ -91,6 +91,7 @@ class CoursesController < ApplicationController
         duplicated.course_memberships.create(user: current_user, role: current_role)
       end
       duplicated.recalculate_student_scores unless duplicated.student_count.zero?
+      Services::ResolvesCopiedLevelBadges.for_course course # some sort of service to resolve badges ids by matching on attributes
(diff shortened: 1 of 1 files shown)


jonathoy:
@acousticrobot Looking for some feedback here. So i'm looking at the `Course` model in the `def copy` method. Level badges are okay the way things currently are set up if you copy on assignment, because the badge ids don't need to be updated.

However, I'm not seeing an easy way of resolving the level badges on the Course copy level as there isn't really context of both the new level badges and the new badges at the same time during the copy process.

It seems like we would have to introduce some sort of Service or Job to basically perform the translations and update the old badge id to the new. Maybe we could translate the badges by matching on many attributes instead of just the name in the event that there aren't unique badge names in the course.

Am I missing a better solution here and if not, do you think that this is a reasonable solution? It's not the most efficient solution, that's for sure.

PR #3722, opened 2017-10-17
Allow student impersonation on unpublished course

### LINKED ISSUE #3476: Student preview available in unpublished state
As an instructor, I often want to be able to preview a course as a student while I'm still working on it (it is unpublished). It would be nice for instructors to be able to use the impersonate feature to see the student side...even if it's not yet published.

### LINKED ISSUE #3811: Allow impersonation of a student even if no students are in the course

jonathoy:
### Status
IN PROGRESS

### Description
This seems suspiciously easy but me and @acousticrobot think that to allow student impersonation on an unpublished course, all that we need to do is change one line. To my knowledge, this seems to work as intended.

Create a new course or un-publish an existing one. Try to preview the course as a student.

Preview as Student

======================
Closes #3811

### DIFF
--- a/app/assets/javascripts/angular/templates/courses/creation_checklist.html.haml
+++ b/app/assets/javascripts/angular/templates/courses/creation_checklist.html.haml
@@ -12,3 +12,8 @@
       )
       .title
         %a{"ng-href"=>"{{item.url}}"} {{ vmChecklist.termFor(item.title) }}
+
+
+  .preview-student-bar
+    .center.italic.small
+      %a(href="/impersonate_student/0") Preview Course as Student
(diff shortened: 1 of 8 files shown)


chcholman:
That does indeed work @jonathoy! One UX tweak I think we should probably consider - how can we call out for the instructors that *they* can see this (which they want) but students can't (which they will then get freaked by, since they can see it).

Longterm, I'm curious how to approach letting the instructor preview the student side...but without a 'real' student to test against. Let me know if you have any ideas for how to approach that!

jonathoy:
@chcholman What are your thoughts on this? I added a generic John Doe student who is used for impersonating. I then put an additional option on the Course Creation pane to allow previewing. Each time you impersonate/unimpersonate it creates or destroys the course membership for the current course, which isn't the most performant thing to do.

The alternative would be to have some sort of type for user that maybe is "System" or something, and we ignore them from grading and rosters.

screen shot 2017-10-18 at 2 23 50 pm

PR #3724, opened 2017-10-18
3723 duplicate score level bug

### LINKED ISSUE #3723: Adding score levels accidentally creates a duplicate of the last level created
### Expected Behavior
As an instructor I want to be able to create score levels - names attached to fixed point values that are awarded. 

### Actual Behavior
When creating score levels the last level I add gets duplicated when I click the Update assignment button

acousticrobot:
### Status
READY

### Description

Fixes a bug where the last score level added was duplicated on submit. This exposes a limitation to the debounce queue `runAllEvents` method: timeouts that have been triggered will stay in the `queueStore` and be retriggered on submit.  For updates this is a performance issue, and for creates this will result in duplicate models.

The solution proposed here is to cancel the events form the timed-out method before making the http call.

======================
Closes #3723


chcholman:
I have confirmed that this works locally. Is this good to merge @acousticrobot and @jonathoy?

jonathoy:
I think we are both good with the changes

PR #3725, opened 2017-10-18
3686 Authenticate through Google

### LINKED ISSUE #3686: Google Authentication
As an instructor, I would like to be able to declare that I and my class would like to use Google Authentication to access GradeCraft

dovek:
### Status
**IN DEVELOPMENT**

### Description
As an instructor, I would like to be able to declare that I and my class would like to use Google Authentication to access GradeCraft

### Related PRs
N/A

### Todos
- [ ] Add Tests

### Deploy Notes
N/A

### Gem dependencies
N/A

### Migrations
NO

As an instructor, create a new learner. That new Learner should receive an email.

From the Learner's email, click the activate account link. This will take the Learner to the set password page. Instead of setting the password, use the Sign in with Google button. This should create a Google specific authorization for the new Learner and successfully log the Learner in to GradeCraft.

If successfully logged into GradeCraft, log out. Now, when the Learner is taken back to the standard login page, the Learner can press the "Sign in with Google" button and be logged into GradeCraft.

If I have never authenticated through Google, then I should be taken to the Authorization error page. This path will be modified in a later edition to instead offer to create an account via Google.

* Sign In

======================
Closes #3686


jonathoy:
@dovek Was there anything in particular that you wanted to to know or were you just ready for a general review of this PR?

app/controllers/authorizations_controller.rb
```diff
@@ -1,9 +1,22 @@
...
+
   def create
     UserAuthorization.create_by_auth_hash request.env["omniauth.auth"], current_user
 
     return_to = session[:return_to]
     session[:return_to] = nil
     redirect_to return_to || root_path
   end
+
+  def log_me_in
```
jonathoy: Let's make this private if it doesn't need to be public.

app/controllers/authorizations_controller.rb
```diff
@@ -1,9 +1,22 @@
 class AuthorizationsController < ApplicationController
+
+  skip_before_action :require_login, only: [:create]
```
jonathoy: I think that we need to keep in mind that this AuthorizationsController is shared amongst all OmniAuth strategies. By making the `log_me_in` method required for the create method, this now has the possibility of affecting how the OAuth portion works when connecting to Canvas. The logic for the method is simple right now, but I can see this being a concern if we introduce additional logic.

I also have concerns about the workflow here. Although it would not be simple to do, I think that there's a potential security issue here where any arbitrary person could login as someone else in GC assuming they know their email and have formatted the request properly to match the OmniAuth structure. I think that you ideally need to `require_authorization` before logging someone in so that we know that the person has a valid OAuth token.

jonathoy:
I think we're nearly there! Fundamentally I think that the workflow is good. I outlined a few suggestions below if you wanted to implement them, but the only concern that you might want to address is #5. I don't see in your logic anywhere that you are doing a variation of `.activate!` and I would ask that you just make sure that this isn't an issue.

**Notes**

1. This may be just an unfinished feature, but what if I click the Authenticate Through Google button and I don't have an account created in Gradecraft yet? Maybe the error message should be more descriptive if this is not a workflow that is supported.

2. After checking out the workflow for `AuthorizationsController#create_for_google`, it seems like maybe you could tidy things up by not having to have that extra route override and controller method. I think you should just be able to add a `before_action :log_me_in, if: Proc.new { |controller| controller.params[:provider] == "google_oauth2" }`, since the logic in `create` and `create_for_google` are very similar. If you wanted to maintain the notice specific to Google authentication, you could maybe set that in some sort of `attr_accessor` as part of the `log_me_in` method. As a side note, the message "GradeCraft now has access to your Calendar, please try to add your item again." might need to be more generic since this callback can be accessed if the user is just trying to log into Gradecraft via Google and not necessarily importing events.

3. Not seeing a huge need for `GoogleController#launch_from_login`. It seems the only place where it's being called is from the button on the login page (`_gradecraft.html.haml`). You could instead just set the button to go directly to "/auth/google_oauth2" to avoid the extra redirect and then get rid of the route and method.

4. At a glance, line 10 in `GoogleController` seems to be redundant with line 16.

5. I think you can simplify `GoogleController#launch_from_activation_token`. If I'm following correctly, you get there from clicking on the link in the activation email. With your current logic, I'm not entirely sure if the account gets activated in GC. Could you check this? I think that the method overall can just be refactored to the following:

```
before_action do |controller|
  controller.redirect_if_auth_not_present current_user
end

# don't launch_from_activation_token if auth is not present
def launch_from_activation_token
  user = User.load_from_activation_token(params[:id]) if current_user.nil?
  user.activate!
  auto_login user and redirect_to dashboard_path
end
```

dovek:
Google Branding Guidelines: https://developers.google.com/identity/branding-guidelines

chcholman:
Current error message if you log in with an email that doesn't have an account: 

```Ack!
We're Having Some Trouble Logging You In.
Are you sure you've created your GradeCraft account? Students must log in through CTools/Canvas/Google before they can access GradeCraft directly.

Follow these directions to create your GradeCraft account, and get in touch if that doesn't work.```

chcholman:
@dovek Will you remove this text as well from the error message? 

```%p #{ link_to "Follow these directions", "http://help.gradecraft.com/"} to create your GradeCraft account, and #{mail_to "cholma@umich.edu", "get in touch"} if that doesn't work.```

PR #3744, opened 2017-10-30
reset grade to incomplete on edit

acousticrobot:
### Status
**READY**

### Description
Assures any grade loaded onto a grade edit page is reset to incomplete and not student visible, so that incremental updates through the autosave are not visible to students.

### Caveats

I think this solution should be implemented quickly, but I can think of a couple improvements that could be created as separate tickets:

1) Grade history currently displays changes to grade complete and student visible to students. We may want to restrict these fields from the history, or other solution (requires digging into how grade history works).

2) Accidentally clicking "Edit" on a visible grade and then "Cancel" will not return to the previous state.  We could send a warning to the professor that they have changed a released grade and must hit resubmit with the appropriate status (easy), or add a "edit-mode" boolean onto a grade that temporarily blocks visibility but preserves the previous state (harder).


chcholman:
@acousticrobot There's a sporadic (I've triggered it 2/4 times) bug where if I have it marked as student visible, click Submit, then click "Cancel" and go back to editing, the grade is actually set to student visible. Can you investigate?

PR #3746, opened 2017-10-31
3739 Duplicated Announcements

### LINKED ISSUE #3739: Quick Grade sending out grade notifications to students multiple times
### Expected Behavior
As a student, I anticipate that if I have gotten a grade for an assignment, I will get one notification about its release 

### Actual Behavior
Currently, if an assignment has been quick graded (this is the confirmed scope, but it may be beyond this), students are getting many Grade Announcements but at increasing intervals. This is an example of one student's announcement center after they had an assignment quick graded. 

screen shot 2017-10-28 at 7 11 12 pm

jonathoy:
### Status
READY (Pending Codeship Build)

### Description
This bug is likely the cause of the service degradations that we've been experiencing in production.

**Findings**:

A recent commit that was merged into master, relating to the ability to rename the Grade Predictor, included a change to reference the custom `term_for` in the grade released mailer template. This helper method is not available for use within the scope of this template.

Since we still had a broken reference to the Papertrail remote logger in the base Resque class implementation, the exceptions that were occurring were being swallowed.

In the `GradeUpdaterJob`, the order of notifying a student is that the announcement will be created followed by the email being sent. The result of this bug was that the job would retry and announcements would get duplicated but no email would ever be sent.

**Additional Fixes:**

Because of the way that the Resque jobs were currently set up, it was also found that the email for challenge grades being released was not sending. This was due to two reasons - that it was also referencing `term_for` as well as referencing improperly the students and course associations on a `ChallengeGrade`.

**Key takeaways:**

- It may be a good idea to go over the rest of the emails to ensure that they haven't been failing silently, although it appears that most of the other job classes have spec tests
- `challenge_grade_update_performer` needs a spec test
- We need to decide on a long-term logging strategy and implement an interim solution, to catch these sorts of problems
- Stretch goal would be to add documentation on how the Resque jobs work, to reduce the headache if this sort of issue ever happens again
- Quick grading one person looks like it still triggers the grade updater job for all graded students, versus just the ones who's grades have changed

### Deploy Notes
All the current grade updater jobs in prod will continue to run and people will continue to receive duplicate announcements until the queue is cleared.

- Quick grade a team challenge and ensure that the email sends for each student in the team
- Quick grade a student on any assignment and ensure that they get one announcement and one email

Grade release announcements + email, challenge grade released email

======================
Closes #3739


jonathoy:
There's also some commented out lines for removal in production.rb, beta.rb, etc. that refer to our previous Papertrail logging configuration.

Gemfile.lock
```diff
@@ -801,4 +797,4 @@ RUBY VERSION
    ruby 2.2.2p95
 
 BUNDLED WITH
-   1.13.7
+   1.15.3
```
chcholman: I'm going to revert this back to 1.13.7

app/performers/challenge_grade_update_performer.rb
```diff
@@ -40,8 +40,10 @@ def fetch_challenge_grade_with_challenge
   end
 
   def notify_challenge_grade_released
-    if @challenge_grade.student.email_challenge_grade_notifications?(@challenge_grade.course)
-      NotificationMailer.challenge_grade_released(@challenge_grade).deliver_now
+    @challenge_grade.team.students.each do |student|
```
chcholman: awesome, this bug has been overdue for a fix

PR #3763, opened 2017-11-03
3649 group submissions on grading status

### LINKED ISSUE #3649: Group submissions not appearing on the grading status page

jonathoy:
### Status
READY

### Description
This PR addresses missing group submissions on the grade status page. The issue is due to an oversight in the underlying Submission scope, `by_active_students`.

Create a group submission and ensure that it shows on the grade status page in all of the various statuses possible (e.g. In progress, Resubmitted, etc.)

Grade status

======================
Closes #3649

### DIFF
--- a/app/models/submission.rb
+++ b/app/models/submission.rb
@@ -28,12 +28,16 @@ class Submission < ActiveRecord::Base
       "grades.student_id = submissions.student_id)")
   end
 
-  scope :by_active_students, -> do
-    joins("INNER JOIN course_memberships ON "\
-      "course_memberships.course_id = submissions.course_id AND "\
-      "course_memberships.user_id = submissions.student_id")
-      .where("course_memberships.active = true")
-      .references(:course_membership, :submission)
+  scope :by_active_individual_students, -> do
+    individual
+      .joins(student: :course_memberships)
+      .where(student: { course_memberships: { active: true }})
+  end
+
+  scope :by_active_grouped_students, -> do
+    where.not(group_id: nil)
+      .joins(group: { students: :course_memberships })
+      .where(group: { students: { course_memberships: { active: true }}})
   end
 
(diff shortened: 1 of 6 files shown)


jonathoy:
@chcholman The proposed solution here is maybe not the neatest solution, but it became incredibly complex to chain together the two queries needed to form the one singular scope that we would have needed for this purpose.

app/models/submission.rb
```diff
@@ -69,7 +74,7 @@ class Submission < ActiveRecord::Base
   multiple_files :submission_files
 
   def self.submitted_this_week(assignment_type)
-    assignment_type.submissions.submitted.by_active_students.where("submissions.submitted_at > ? ", 7.days.ago)
+    assignment_type.submissions.submitted.where("submissions.submitted_at > ? ", 7.days.ago)
```
chcholman: So this no longer filters out students who have been deactivated - we need to address that

app/models/submission.rb
```diff
@@ -28,12 +28,16 @@ class Submission < ActiveRecord::Base
...
+  scope :by_active_individual_students, -> do
+    individual
+      .joins(student: :course_memberships)
+      .where(student: { course_memberships: { active: true }})
+  end
+
+  scope :by_active_grouped_students, -> do
+    where.not(group_id: nil)
+      .joins(group: { students: :course_memberships })
+      .where(group: { students: { course_memberships: { active: true }}})
```
chcholman: So a concern I have here is what happens if one student is deactivated? The submission should still show if anyone in the group is active

app/models/submission.rb
jonathoy: The active students requirement has been moved up into the view.

chcholman: Awesome

PR #3766, opened 2017-11-03
Gradebook - use grade score instead of full_points

jonathoy:
### Status
READY

### Description
Change the gradebook export to use the grade score instead of the `full_points`

======================
Closes #[ISSUE NUMBER]


jonathoy:
Were you able to figure out a good way to view the export locally? I believe this is the only change that needs to be made since the Gradebook view in the app uses score.

jonathoy:
@chcholman I'm seeing in the support ticket that it was suggested that the opposite was true - that the export is correct and the gradebook view is incorrect. Which should we be using? Score or full_points?

chcholman:
@jonathoy `full_points` is actually a different field (and not the one we want to use here). I propose`final_points` (not `score`) in both places (not least because it's the values that are actually summed by instructors, and when they compare it to students grades, it will make sense).

PR #3774, opened 2017-11-06
Refactor RollbarJS Configuration

### LINKED ISSUE #3773: RollbarJS Refactoring
We are currently receiving many requests from RollbarJS that may or may not be erroneous. We should refactor the current method for configuring RollbarJS so that the token is a bit harder to determine and include some user data for logged exceptions.

jonathoy:
### Status
IN DEVELOPMENT

### Description
This PR makes the client-side token for RollbarJS a little bit more difficult to obtain by having it pull from ENV variables.

It also includes some user information for the `current_user` in the form of their id, username, and email.

### Deploy Notes
For each environment, the following strings in `.env` need to be configured:

ROLLBAR_ENV = staging, production, or beta
ROLLBAR_CLIENT_TOKEN = Under Rollbar project settings, use the new client-key labeled `post_client_item_new`.

### Migrations
NO

RollbarJS generic + Angular error logging

======================
Closes #3773


jonathoy:
Added ENV variables to staging, production, and beta. Once this feature is merged, the old client token can be deleted. We should continue to tweak the rate limiting on the new token in case we get lots of erroneous requests again.

PR #3775, opened 2017-11-06
Remove errant dynatable config

jonathoy:
### Status
READY

### Description
Not sure if this is a rebase artifact, but at some point (231012519ea86ddce552284d57e5902bdb06ab77) it appears that a line was removed but is now back.

The Dynatable bug was first noticed on the Manage Courses page, but has the possibility of affecting other Dynatable uses.


jonathoy:
@chcholman Since you last worked on these changes, could you double-check that there isn't anything else out of place?

PR #3776, opened 2017-11-08
Course copy review

### LINKED ISSUE #3770: Issues in Course Copy
### Expected Behavior
As a user, I anticipate being able to copy a course and all associated objects except the staff, students, groups, grades, submissions, events, challenge_grades, assignment_type_weights, announcements, earned_badges, and analytics will come forward (there may be others - these are the initial ones) along with it. 

### Actual Behavior
The copy feature is currently failing to properly copy assignments into new assignment types. It's possible this bug is fixed, and the course we're copying (175 in app) has some bad data that is causing the issue to persist, but before moving forward I would like to do thorough testing in the staging environment to confirm that things are good. 

### Steps to reproduce
In staging, trying copying one of the Videogames & Learning courses (Not the UX version, which doesn't have real data associated with it). Confirm that: 

- [ ] Course settings come through properly, including the syllabus file if attached, and connecting to the right institution if present
- [ ] Assignment Types copy properly
- [ ] Assignments come through properly, including assignment files, rubrics, their criteria and level badges
- [ ] Unlock Conditions copy properly
- [ ] Badges copy properly, including Badge files
- [ ] Team Challenges, including Challenge files

jonathoy:
### Status
HOLD

### Description
Unlock conditions need to be copied. However, the associations will refer to the wrong models just as they do with level badges when a course is copied.

The best way of fixing this issue will depend on the work being documented and continued in https://github.com/UM-USElab/gradecraft-development/pull/3717.

======================
Closes #3770


jonathoy:
@acousticrobot Any additional thoughts on unlock conditions being copied? It sounds like the work that you had started will help to address the issue of us having errant associations.

PR #3781, opened 2017-11-20
User search enhancements for admins

### LINKED ISSUE #3438: Fuzzy search for users
Currently the user search requires you to enter all of a user's full first or last name in order to match the response. It would be fantastic for the search to be able to match pieces of names instead.

jonathoy:
### Status
READY

### Description
As part of a refactor for the admin user search feature, the Angular form was pulled out and replaced with a form backed by the Ransack gem. The reason for this was that it provides more flexibility out-of-the-box when searching and it may prove useful in the future for features that require more complex querying (this was investigated for the use of auto-awarded badges).

Currently pagination is not implemented, therefore the default resultset size is 50. I would be happy to tweak this value, although a user can also do this on a per request basis by changing the `max_results` value in the generated search url.

For example:
```
/users/search?utf8=✓&max_results=50&q[username_cont]=jonathoy&commit=Search
```

I have also added two new indices to the `Users` table - one for the `first_name` and another for the `last_name`. This allowed for a drastic improvement in performance.

### Gem dependencies
- ADDED Ransack 1.8.4

### Migrations
YES

As an admin, ensure that there are no observed issues when searching for users

User search

======================
Closes #3438


jonathoy:
@evanstraub In staging, for your viewing pleasure!

PR #3783, opened 2017-11-21
Only set the submission_id on the grade if not nil

jonathoy:
### Status
HOLD

### Description
Observation as reported by Evan:

From the assignment show page, there appear to be three resubmissions. However, from the grading status page there is only one for that assignment. The two places differ because they are based off of two different resubmission methods on the Submission model.

Essentially, one of the grades had a `submission_id` that was correctly set and the other two had a `submission_id` of `nil`, even though a submission existed for the user + assignment.

I wasn't able to reproduce this locally, but I know that the assigning of the `submission_id` to the grade happens in the `AssociatesSubmissionWithGrade` service. For this PR, I changed the last line in that action to only update the `submission_id` if it is not `nil`.

======================
Closes #[ISSUE NUMBER]

### DIFF
--- a/app/services/creates_grade/associates_submission_with_grade.rb
+++ b/app/services/creates_grade/associates_submission_with_grade.rb
@@ -17,7 +17,7 @@ module Services
           s = Submission.find_by(assignment_id: assignment.id,
                                  student_id: student.id)
         end
-        context[:grade].submission_id = s.nil? ? nil : s.id
+        context[:grade].submission_id = s.id unless s.nil?
       end
     end
   end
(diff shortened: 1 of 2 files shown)


jonathoy:
@chcholman Any thoughts on this? I have the specific details for this occurrence but I've already performed manual updates in the database to link the submission to the grade. Again, I wasn't able to replicate this locally so this is somewhat of a shot at the dark in terms of a fix.

chcholman:
@jonathoy I'm fine merging this, it seems logical. But, should we be thinking about iterating the method used on the grading status page as well?

PR #3789, opened 2017-12-04
Autosave notifications

### LINKED ISSUE #3788: Earned badge notifications being sent on autosave
If an assignment has been graded previously and an instructor goes back to edit, any time that they award a badge they will receive a notification. This should only happen after the changes have been submitted and not when the change is autosaved.

jonathoy:
### Status
IN DEVELOPMENT

### Description
When editing an existing grade, if you click on a badge and it autosaves, a notification will be generated for the student even if you have not submitted yet.

### Migrations
NO

Grade an assignment and observe notifications behavior on autosave when awarding additional badges

Grading

======================
Closes #3788

### DIFF
--- a/app/assets/javascripts/angular/services/BadgeService.coffee
+++ b/app/assets/javascripts/angular/services/BadgeService.coffee
@@ -186,7 +186,8 @@
     $http.post('/api/earned_badges/', requestParams).then(
       (response)-> # success
         if response.status == 201
-          GradeCraftAPI.addItem(earnedBadges, "earned_badges", response.data)
+          result = GradeCraftAPI.addItem(earnedBadges, "earned_badges", response.data)
+          _.last(earnedBadges).pending_notify = true if result? and !notify
         GradeCraftAPI.logResponse(response)
         setBadgeIsUpdating(badgeId, false)
       ,(response)-> # error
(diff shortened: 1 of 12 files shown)


app/assets/javascripts/angular/services/BadgeService.coffee
```diff
@@ -174,18 +174,20 @@
...
+      "badge_id": badgeId
       "grade_id": gradeId
+      "notify": notify
     }
 
     $http.post('/api/earned_badges/', requestParams).then(
       (response)-> # success
         if response.status == 201
-          GradeCraftAPI.addItem(earnedBadges, "earned_badges", response.data)
+          result = GradeCraftAPI.addItem(earnedBadges, "earned_badges", response.data)
```
acousticrobot: I'm not sure why you need to assign this to a variable, am I missing where this is used?

jonathoy: I wanted to mark the earned badges as needing to be notified. I looked at `GradeCraftAPI.addItem` and noticed that it will either return undefined if the item isn't added, or it will return a value if it was added. Either way it does a push to the modelArray, so the last item that was added should be marked with `pending_notify = true`.

Hence: 
```
_.last(earnedBadges).pending_notify = true if result? and !notify
```

jonathoy:
For the meantime, the work in #3777 should stop these emails from being generated. If we continue to encounter issues related to email notifications, I would want to look at a more thorough refactor of the notification logic.

PR #3798, opened 2017-12-11
Copy files

### LINKED ISSUE #3770: Issues in Course Copy

jonathoy:
### Status
READY

### Description

These changes are an extension of the work performed in #3794, which involves resolving copy lookups. The changes in this PR allow the following to be copied as well:

1. Badge icon
2. Badge files
3. Assignment media
4. Assignment files
5. Challenge media
6. Challenge files
7. Course syllabus

The process is that the files themselves will be copied on S3, rather than just having the file in the copied model reference the file in the original.

Notes: Two notable changes were made to the `ModelCopier` class. One involves switching the order of operations for performing lookups and performing overrides and the other involves making the original model readonly so as to reduce any unintended side-effects.

**Future changes**

- Need to address polymorphic unlock conditions, which are not currently able to be copied via #3794 
- Need to access files always using presigned urls so that we can remove public ACLs

**Todo**

- Ensure that remote files in S3 are deleted on model destroy, if we want to be mindful of storage usage
- Ensure that when copying a specific assignment, the assignment files are actually being replicated in S3 rather than simply being linked to the original file.

### Deploy Notes
**As an interim solution, public ACLs on the applicable buckets in S3 must be created to ensure that badge icons are visible post-copy**

1. Badge icon
2. Assignment media
3. Challenge files

See current bucket policy for gradecraft-staging for additional details.

### Gem dependencies
- UPDATED `acts_as_list` to latest (0.9.10), to allow suppressing of unwanted behavior that happens behind the scenes on model updates.

### Migrations
NO

- May need to ensure that there are no errant `BadgeFile`s in production, since this adds a new validation to ensure that a `BadgeFile` must have a file.

Course copy

======================
Closes #3770


jonathoy:
Observation made while testing:

In staging, course 201 (Teaching with Technologies in 6-12) has an invalid `assignment_type` due to the fact that it fails the `max_points` greater than 0 rule. Most of the course will copy (minus the errant assignment type), but the message returned post-save will say that it failed.

This is likely a data anomaly, but since incremental saving happens throughout the copy process now what happens if there is invalid data somewhere downstream in the process? Should the entire course be destroyed?

jonathoy:
Need to determine status of our infrastructure support, specifically with database backups and restores. Although the developers have tested the features with confidence on staging, it would be ideal to have a failsafe since the breadth of the data changes when performing course copy is vast.

evanstraub:
Testing course copy on staging with many things.

-Assignment structure and order - YES
-Assignment file attachments - NO
-Rubric level badges - YES
-Badge icons - YES
-Assignment unlocks/locks - YES
-Badge unlocks - YES

jonathoy:
@evanstraub I pushed a change to staging that hopefully should fix assignment files, if you have a chance to take a look again.

PR #3818, opened 2018-01-23
Nightly score recalculations

### LINKED ISSUE #3817: Nightly score recalculation
We should add a recurring job that will run the score recalculator on a nightly basis.

jonathoy:
### Status
IN DEVELOPMENT

### Description
Currently triggers score recalculation jobs for each unique combination of `student_id` and `course_id`, based on all grades and earned badges that are `student_visible` and updated within the last 24 hours.

Anybody with `admin=true` on their User model can access the job panel by going to '/jobs'.

### Gem dependencies
Updated multiple gems in order to add `resque-web` interface for admin users. See Gemfile.lock for further details

### Migrations
NO

Ensure that the job is occurring daily at 2:00 am (this can be changed)

Grades

======================
Closes #3817


jonathoy:
Okay. So.

I've determined that the timeout error that we are experiencing in Quay during the build is not related to any sort of Gem update, but rather it happened when I added line 28 in the `config/initializers/resque.rb` in commit ae6744701178c351a2bf7f8d7987f35c628181a0.

I've found that it was necessary to add this line to the initializer rather than the scheduler task in the Rakefile. What it does is it loads the nightly schedule from the YAML file into Redis. Since we had never used scheduled jobs before in the app, we never really experienced this issue.

So the issue..

From my investigating, I've found that this timeout has happened in the past due to a large number of expired keys. Being not that familiar with Redis, I'm still looking into what constitutes an expired key. What I do know is that when I run the command `REDIS.keys.count` in umich prod, I get 21512 which seems like a lot.

I'm documenting my findings here for now while I continue this investigation, but I may need someone better versed in the infrastructure side of things to help with this. This feature will be useless if we can't figure out a resolution for it. We appear to have a complex, and possibly overly so, Resque setup going on.

jonathoy:
Update: The number of Redis keys is limited by only the amount of memory available. Plus, I don't think that any of our keys have an expiration therefore they would not be considered volatile.

I took particular note of [this](https://github.com/resque/resque-scheduler/issues/541#issuecomment-218894599) comment in an issue that was opened for the `resque-scheduler` gem and it seems to be the same issue that we are experiencing.

I can load the schedule in the initializer as it was done in ae67447, but the Docker build will fail. If I load it in the Rakefile, it will only show up on the schedules tab in the `resque-web` interface if I set `persist: true`. If we have no interest or need for the web interface, I'm happy to rollback the Gem updates and the commits to get rid of it. In this case, the schedule will just run quietly in the background.

jonathoy:
I made a change to the staging Dockerfile in order to properly test this work. For some reason, there was a configuration line that wasn't starting up `resque-scheduler` so I added it back in. There seem to be no ill-effects from doing this and it now more closely resembles what is happening in the production and app deployments.

PR #3819, opened 2018-01-24
updated rollbar version

chcholman:
### Status
**READY**

### Description
This PR updates the Rollbar gem to the latest version


jonathoy:
👍  We should update the rollbar-js client as well, but that is less important right now. Want this merged into production now?

chcholman:
@jonathoy I don't think it's a rush - just getting it in the next time you post should be fine! Merging now

PR #3859, opened 2018-02-05
3821 copy unlocks

### LINKED ISSUE #3821: Copy Unlock Conditions
Course copy should copy the unlock conditions in the original course. This requires new functionality in the copy procedure to overcome these two hurdles:

1) Unlock conditions are polymorphic
2) Models can be cross conditions -- i.e. assignments for badges and badges for assignments. We need to have all models copied and the new id's stored for reference before we can copy conditions.

acousticrobot:
### Status
**READY**

### Description

Part III of fixing the course copy bugs.

This PR builds upon the lookup store and file copy work, and brings in copying for conditions.

A final copy procedure has been added, called `:cross_references`, which acts like associations, but copies after all associations have been copied. This assures that the copy_log is full of all references.

It also adds a check for polymorphic relationships, so if the reference is not found in the lookup, the model is checked for a reference type.

It also saves the final copy log as a model, for debugging purposes. Currently only available in the console, this could also be made accessible via the api.

### Related PRs

#3794
#3798

### Deploy Notes

This rake task must be run before copying will work:

`bundle exec rake conditions:add_course_ids

### Migrations
YES

======================
Closes #3821

### DIFF
--- a/app/models/copy_log.rb
+++ b/app/models/copy_log.rb
@@ -1,0 +1,14 @@
+class CopyLog < ActiveRecord::Base
+  belongs_to :course
+
+  validates_presence_of :course
+  validates_presence_of :log
+
+  def to_hash
+    eval self.log
+  end
+
+  def parse_log(hash_log)
+    self.log = hash_log.to_s
+  end
+end
--- a/app/models/model_copier.rb
+++ b/app/models/model_copier.rb
@@ -164,9 +164,18 @@ class ModelCopierLookups
 
       # :courses => :course_id
       id_key = "#{class_type.to_s.singularize}_id".to_sym
+      id_type = "#{class_type.to_s.singularize}_type".to_sym
 
       if original.respond_to?(id_key) && lookup_has_key?(class_type, original.send(id_key))
         h[id_key] = lookup(class_type, original.send(id_key))
+
+      # handle polymorphic relationships
+      # In order for this to work for unlock conditions, they have to be run after all other
+      # models have been copied -- If badges can be conditions for assigment and assignment for badges,
+      # we have to copy all the models first, then go back and lookup the associations.
+      # TODO: add course_id to conditions, and then copy them last from the course copy.
+      elsif original.respond_to?(id_type) && lookup_has_type?(original.send(id_type), original.send(id_key))
+        h[id_key] = lookup(original.send(id_type), original.send(id_key))
       end
       h
     end
--- a/app/models/model_copier.rb
+++ b/app/models/model_copier.rb
@@ -16,6 +16,10 @@ class ModelCopier
     copied.save
     lookup_store.store(original, copied)
     copy_associations options.delete(:associations) {[]}, attributes
+    # copy associations that cross-reference other models, and need to
+    # be run after all associations have already been copied.
+    copy_associations options.delete(:cross_references) {[]}, attributes
+    copied.log_copy(lookup_store) if copied.respond_to?(:log_copy)
     copied
   end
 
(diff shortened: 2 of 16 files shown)


app/models/model_copier.rb
```diff
@@ -74,23 +99,85 @@ def initialize(association)
...
+
+      # :courses => :course_id
+      id_key = "#{class_type.to_s.singularize}_id".to_sym
+      id_type = "#{class_type.to_s.singularize}_type".to_sym
+
+      if original.respond_to?(id_key) && lookup_has_key?(class_type, original.send(id_key))
+        h[id_key] = lookup(class_type, original.send(id_key))
+
+      # handle polymorphic relationships
+      elsif original.respond_to?(id_type) && lookup_has_key?(original.send(id_type).underscore.pluralize.to_sym, original.send(id_key))
```
jonathoy: I hope that I will not have to touch this 😬

app/models/copy_log.rb
```diff
@@ -0,0 +1,14 @@
+class CopyLog < ActiveRecord::Base
+  belongs_to :course
+
+  validates_presence_of :course
+  validates_presence_of :log
+
+  def to_hash
+    eval self.log
```
jonathoy: I wonder if a safer way to store and restore data in the form of hashes would be to use [ActiveRecord::Store](http://api.rubyonrails.org/classes/ActiveRecord/Store.html)? I've not used it before but it looks very straightforward to implement.

app/models/unlock_condition.rb
```diff
@@ -1,9 +1,14 @@
...
+  include Copyable
+
+  belongs_to :course
   belongs_to :unlockable, polymorphic: true
   belongs_to :condition, polymorphic: true
 
   validates_presence_of :condition_id, :condition_type, :condition_state
   validates_associated :unlockable
+  # TODO: add course validation after rake task is run
+  # validates_presence_of :course
```
jonathoy: I'm going to point this out so that I remember to do it whenever we are in agreement that the course copy work can be merged.

db/schema.rb
```diff
@@ -221,6 +221,11 @@
     t.boolean  "include_in_timeline", default: false, null: false
   end
 
+  create_table "copy_logs", force: :cascade do |t|
+    t.text    "log"
+    t.integer "course_id"
+  end
```
jonathoy: Should we maybe add
```
t.timestamps
```
so that we have some time information available?

jonathoy:
Noting here that I ran the rake task in staging, which is pretty close to production, and one record failed to update due to some bad data on the associated unlockable assignment. When running this in production, we should double check that there are no 

```
UnlockCondition.where(course_id: nil)
```

jonathoy:
If I uncomment the validation that you put into `unlock_condition.rb` to validate the presence of the course, I get a whole lot of spec failures which makes sense.

app/models/unlock_condition.rb
acousticrobot: It will be a two step process -- get everything merged into production, and then a second PR with this uncorked, and updates to the u_c factory and specs.

app/models/copy_log.rb
acousticrobot: Feel free to give this a try if you want, I just spent a half hour playing with adding `store :log` to the model and couldn't get it to work.  I don't actually use the `to_hash` for anything, so we can also just take it out and leave this solution for another time.

db/schema.rb
acousticrobot: 👍

app/models/copy_log.rb
acousticrobot: Update: I just added an api route to view the log, and it worked much better to actually use the `to_hash` method, so we can think about alternatives in this PR. I do think you were right when you said if someone can get in to the database and put in a damaging hash, then they would have access to do much worse, but I also know the robots are complaining about the `eval`.

jonathoy:
@acousticrobot I rebased against the latest changes in #3798 and noticed a duplicate commit. Your unlock branch should be good now, but if you need to continue to work off of this I would recommend scrapping your local one and resetting to the current state in origin.

PR #3893, opened 2018-03-05
Reset multiplier-related fields if unused

### LINKED ISSUE #3892: Multipliers display after disabling
In the scenario where an instructor has previously used and set up multipliers for their course, if their course is copied and they try to uncheck using multipliers in the copied course, then as a student they will continue to see coins in the predictor.

jonathoy:
### Status
READY

### Description
This PR clears out the `total_weights`, `weights_close_at`, `max_weights_per_assignment_type`, and `max_assignment_types_weighted` fields if the `has_multipliers` value changes to false via a callback on the Course model.

======================
Closes #3892

### DIFF
--- a/app/assets/javascripts/angular/templates/predictor/main.html.haml
+++ b/app/assets/javascripts/angular/templates/predictor/main.html.haml
@@ -21,8 +21,6 @@
           .predictor-title-right-span
             Weights are Closed for This Course
 
-
-
     %predictor-section-assignment-types
 
     %predictor-section-badges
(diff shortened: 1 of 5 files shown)


jonathoy:
Off the top of your head, do you know of other places in the app where this might cause any sort of visual bugs @acousticrobot? I am not all that familiar with multipliers and so I only verified that this hides it from the student predictor.

jonathoy:
I'm going to merge this and suggest that we take a deeper look as part of #3895

PR #3894, opened 2018-03-05
3863 Badges Earned Export Glitch

### LINKED ISSUE #3863: Final Grade export is reporting all student badges earned, not just for that course
### Expected Behavior
As an instructor, I expect that the final grades export will report the number of earned badges the student has earned for that course 

### Actual Behavior
Instructors are seeing the count of all badges ever earned by the student

### Steps to reproduce
Download final grades export

dovek:
### Status
**IN DEVELOPMENT**

### Description
Final Grade export is reporting all badges earned by a student for all courses. It should be reporting specifically for the course that the export is being ran against.

### Related PRs
N/A

### Todos
- [ ] Add Tests

### Deploy Notes
N/A

### Gem dependencies
N/A

### Migrations
NO

As an instructor, run the export final grades export. Then in a rails c console, run the Course.find(x).earned_badges.where(student_id: [student id your checking against]).count. These numbers should match.

* Exports

======================
Closes #3863

### DIFF
--- a/app/exporters/course_grade_exporter.rb
+++ b/app/exporters/course_grade_exporter.rb
@@ -25,7 +25,7 @@ class CourseGradeExporter
       student.score_for_course(course),
       student.grade_letter_for_course(course),
       student.grade_level_for_course(course),
-      student.earned_badges_for_course.count,
+      student.earned_badges_for_course(course).count,
       student.id,
       student.created_at,
       student.last_course_login(course),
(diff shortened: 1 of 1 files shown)


app/exporters/course_grade_exporter.rb
```diff
@@ -25,7 +25,7 @@ def student_data(student, course)
       student.score_for_course(course),
       student.grade_letter_for_course(course),
       student.grade_level_for_course(course),
-      student.earned_badges.count,
+      course.earned_badges.where(student_id: student.id).count,
```
chcholman: I would use the `earned_badges_for_course(course)` method instead - it's on the `User` model

chcholman: (You'll still need to concatenate `count` onto it)

chcholman:
I would switch to using the established method for querying the count here, but otherwise looks good!

PR #3897, opened 2018-03-06
Add copy validation

### LINKED ISSUE #3889: Run Validator Before Copy
Course copy will have errors if any of the models are invalid, since the duplicate won't save.

We need to validate all the following models before copy :
  * assignment
  * assignment_type
  * badge
  * challenge
  * course
  * course_membership
  * criterion
  * event
  * grade_scheme_element
  * level
  * level_badge
  * rubric
  * team
  * team_membership
  * unlock_condition
  * concerns

jonathoy:
### Status
READY

### Description
Prior to copying a course **or** assignment, all associations about to be copied should be validated so that we don't end up with incorrectly linked models when save is called in the `ModelCopier`.

**Caveat: The validation process in this PR relies on the copy associations being defined in the `COPY_ASSOCIATIONS` hash. If we want to copy additional associations in the future, this needs to be updated as needed.**

### Migrations
NO

Force one or more model errors (possible via `update_columns`) and attempt to do a copy. The process should fail and you should receive a raw JSON response which displays some error validation results.

Download a JSON prettifier for your browser if you don't have one!

Course copy
Assignment copy

======================
Closes #3889

### DIFF
--- a/app/controllers/assignments_controller.rb
+++ b/app/controllers/assignments_controller.rb
@@ -55,9 +55,13 @@ class AssignmentsController < ApplicationController
   # Duplicate an assignment - important for super repetitive items like
   # attendance and reading reactions
   def copy
-    assignment = current_course.assignments.find(params[:id])
-    duplicated = assignment.copy_with_prepended_name
-    redirect_to edit_assignment_path(duplicated), notice: "#{(term_for :assignment).titleize} #{duplicated.name} successfully created"
+    begin
+      assignment = current_course.assignments.find(params[:id])
+      duplicated = assignment.copy_with_prepended_name
+      redirect_to edit_assignment_path(duplicated), notice: "#{(term_for :assignment).titleize} #{duplicated.name} successfully created"
+    rescue InvalidAssociationError => e
+      render json: { message: e.message, details: e.details }, status: :internal_server_error
+    end
   end
 
   def destroy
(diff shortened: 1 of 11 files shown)


jonathoy:
If there are one or more errors on a course, you will instead get a JSON response with a 500 status. Looks like this:

```json
"message":"Failed to copy Jon's Test Course due to validation errors",
"details":{
   "courses":[
      {
         "type":"course",
         "id":878,
         "valid":true,
         "errors":null
      }
   ],
   "badges":[
      {
         "type":"badge",
         "id":1312,
         "valid":true,
         "errors":null
      }
   ],
   "assignment_types":[
      {
         "type":"assignment_type",
         "id":1551,
         "valid":true,
         "errors":null
      },
      {
         "type":"assignment_type",
         "id":1894,
         "valid":true,
         "errors":null
      }
   ],
   "assignments":[
      {
         "type":"assignment",
         "id":11812,
         "valid":true,
         "errors":null
      },
      {
         "type":"assignment",
         "id":13833,
         "valid":false,
         "errors":[
            "The full points for the assignment must be less than the cap for the whole assignment type."
         ]
      },
      {
         "type":"assignment",
         "id":15529,
         "valid":true,
         "errors":null
      },
      {
         "type":"assignment",
         "id":15528,
         "valid":true,
         "errors":null
      }
   ],
   "rubrics":[
      {
         "type":"rubric",
         "id":4516,
         "valid":true,
         "errors":null
      },
      {
         "type":"rubric",
         "id":5044,
         "valid":true,
         "errors":null
      }
   ]
}
}
```

jonathoy:
@acousticrobot I'm getting sporadic test failures on this work and aren't sure why. Locally it's fine so I think it must be something persisted across tests. If you have the time to look while you're reviewing that'd be awesome.

https://app.codeship.com/projects/106957/builds/32719191?pipeline=54ef631d-2e67-4624-bc75-e64763953bf1

PR #3901, opened 2018-03-07
Hide assignment types on the student index until assignmsn visible

### LINKED ISSUE #3835: Hide assignment types if there are no visible assignments within them

chcholman:
### Status
**READY**

### Description
Addresses a glitch where students would still see assignment type categories even when all of the assignments within it were not visible to them (#3835)

### Migrations
NO

* Student index

======================
Closes #3835

### DIFF
--- a/app/models/assignment_type.rb
+++ b/app/models/assignment_type.rb
@@ -150,6 +150,16 @@ class AssignmentType < ActiveRecord::Base
     score
   end
 
+  # checks to see if the assignment type has any assignments within it that the
+  # student can see - used to hide assignment type's external list item on the
+  # student index until at least one assignment within it has been made visible
+  # to that student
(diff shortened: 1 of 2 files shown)


app/models/assignment_type.rb
```diff
@@ -150,6 +150,16 @@ def max_points_for_student(student)
     score
   end
 
+  # checks to see if the assignment type has any assignments within it that the
+  # student can see - used to hide assignment type's external list item on the
+  # student index until at least one assignment within it has been made visible
+  # to that student
+  def visible_assignments_for_student?(student)
```
jonathoy: I think this can be simplified to this:

```
def visible_assignments_for_student?(student)
  assignments.any? { |a| a.visible_for_student? student }
end
```

Issue #3880, opened 2018-02-20
Improve Gradecraft Performance


jonathoy:
Gradecraft slow. Make better.

jonathoy:
In order of priority for async rewrites in Angular:

1. Dashboards for all users (everyone hits them on login)
2. Assignment index
3. Assignment Show page 
4. Student Index
5. Grading Status
6. User index
7. Course index

To be done as part two of this issue

acousticrobot:
#2 Assignments Index page is now in it's own ticket: #3910

jonathoy:
Noting that for our largest courses, Talent Gateway for example, we're approaching the point where we may want to consider taking the refactor work one step further by paginating results. We are starting to display way too much HTML on certain pages. This should be fairly straightforward to do with the batch loading that has been put in.

PR #3911, opened 2018-03-15
Angularize dashboard due this week component

jonathoy:
### Status
READY

### Description
This PR Angularizes the due this week component that is displayed on the dashboard. After investigating via the use of `rack-mini-profiler`, it was found that this component takes the longest to render.

Step one is to load this component asynchronously and step two is to see if we can get rid of the presenter and maybe improve upon some of the back-end querying.

### Migrations
NO

This one is a difficult one to test. This widget obviously displays assignments that have been predicted by the student as well as those that are due within the week.

The view also differs for an instructor versus a student. We should also ensure that with the submission button refactor, there is not an errant scenario where the button is showing when it isn't supposed to or it is showing the wrong text or links to the wrong page.

screen shot 2018-03-15 at 11 09 54 am

======================
Relates to #3880


jonathoy:
After having gone through the process of rewriting this component, I realized how complex the backend logic really was. I think this work is going to require some extensive testing.

If while you're reviewing this, you see places where things can be further improved please let me know. For now, I've left the presenter logic mostly the same.

jonathoy:
screen shot 2018-03-21 at 12 16 34 pm

jonathoy:
Need to check where the helper was swapped out in place of the old template. Evan reported that she was seeing the submit button on assignments that did not accept submissions.

jonathoy:
Noticing an error with this component on Talent Gateway course 110

PR #3934, opened 2018-03-26
Add recaptcha to trial creation form

### LINKED ISSUE #3933: Better restrict trial account creation process

jonathoy:
### Status
**READY**

### Description
**Requires site keys to be generated**

This PR adds an additional layer of security to the form for creating trial accounts that is accessible on app.gradecraft.com.

Also generates additional keys for future use for production and staging.

If additional security is needed, we can up the verification requirement from the recaptcha dashboard.

### Gem dependencies
- ADDED recaptcha

### Migrations
NO

### Deploy Notes
Keys must be generated per environment and added to the respective Dockerfiles as ENV variables.

Ensure that for a non-beta environment, it is not possible to visit `/users/new_external` or create a trial account.

If the environment is beta, ensure that the reCaptcha verification is a requirement for account creation.

Trial account creation

======================
Closes #3933


chcholman:
Looks good to me @jonathoy! Merge whenever ready.

PR #3935, opened 2018-03-28
Fix lo bugs

### LINKED ISSUE #3927: Learning Objective Improvements

jonathoy:
### Status
**READY**

### Description

- Fixes some visibility issues related to assessed outcomes
- Learning objective index view was always displaying a count to achieve, even if the course was configured to use points
- Adds additional tests that were missing as well as some much needed sample data 🎉 
- Restrict API for learning objective outcomes to only the student, if the current user is a student

Learning objectives

======================
Closes #3927

### DIFF
--- a/app/assets/javascripts/angular/directives/learning_objectives/components/_objectives_table.coffee
+++ b/app/assets/javascripts/angular/directives/learning_objectives/components/_objectives_table.coffee
@@ -22,6 +22,7 @@
   {
     scope:
       category: '='
+      objectivesAwardPoints: '='
     bindToController: true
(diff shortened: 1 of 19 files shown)


jonathoy:
@evanstraub If either you or Michael have the time to do another round of testing of learning objectives, that would be great. This work should already be in staging.

It's not urgent, but just as an FYI there is a situation that Laurie has created with her staging course where she has since deleted a competency level that was linked to an assessed outcome. This is a discussed edge case that is currently causing the outcome to now show on the student view. We should communicate with Laurie at some point what she might want to happen in this case.

PR #3942, opened 2018-04-02
3899 simplify resubmissions

### LINKED ISSUE #3899: Simplify Resubmissions
### Expected Behavior
As an instructor, when I create an assignment that accepts submissions and resubmissions, it won't matter if a student has submitted before or not, once the assignment has been graded the first time then every thing submitted afterwards will be considered a resubmission. This will be reflected on the Grading Status page.

### Actual Behavior
Not the above.

### Steps to reproduce
1. Have an instructor create an assignment that accepts submissions
2. Have a student start the submission process but not complete it.
3. Have the instructor grade the assignment as a non-submission
4. Have the student receive their grade and attempt to resubmit.

dovek:
### Status
**READY**

### Description
_Expected Behavior_
As an instructor, when I create an assignment that accepts submissions and resubmissions, it won't matter if a student has submitted before or not, once the assignment has been graded the first time then every thing submitted afterwards will be considered a resubmission. This will be reflected on the Grading Status page.

_Actual Behavior_
As an instructor, when I create an assignment that accepts submissions and resubmissions, a scope and method are in place to look for a submission and a grade in order for a submission to be considered a resubmission.

### Related PRs
N/A

### Todos
- [x] Add Tests

### Deploy Notes
N/A

### Gem dependencies
N/A

### Migrations
YES

1. Have an instructor create an assignment that accepts submissions
2. Have a student start the submission process but not complete it.
3. Have the instructor grade the assignment as a non-submission
4. Have the student receive their grade and attempt to resubmit.

* Submissions

======================
Closes #3899

### DIFF
--- a/app/models/submission.rb
+++ b/app/models/submission.rb
@@ -39,26 +39,6 @@ class Submission < ActiveRecord::Base
     .where.not(id: with_grade.where(grades: { instructor_modified: true }))
   end
 
-  # def submission_grade
-  #   if assignment.has_groups?
-  #     group.grade_for_assignment assignment if group.present?
-  #   else
-  #     student.grade_for_assignment assignment if student.present?
-  #   end
-  # end
-
-  # def resubmitted?
-  #   submission_grade && submission_grade.student_visible? &&
-  #   !graded_at.nil? && !submitted_at.nil? && graded_at < submitted_at
-  # end
-
-  # scope :resubmitted, -> {
-  #   includes(:grade, :assignment)
-  #   .where("grades.student_visible = true")
-  #   .where("grades.graded_at < submitted_at")
-  #   .references(:grade, :assignment)
-  # }
-
   scope :order_by_submitted, -> { order("submitted_at ASC") }
   scope :for_course, ->(course) { where(course_id: course.id) }
   scope :for_student, ->(student) { where(student_id: student.id) }
--- a/app/models/submission.rb
+++ b/app/models/submission.rb
@@ -39,6 +39,19 @@ class Submission < ActiveRecord::Base
     .where.not(id: with_grade.where(grades: { instructor_modified: true }))
   end
 
+  # def submission_grade
+  #   if assignment.has_groups?
+  #     group.grade_for_assignment assignment if group.present?
+  #   else
+  #     student.grade_for_assignment assignment if student.present?
+  #   end
+  # end
+
+  # def resubmitted?
+  #   submission_grade && submission_grade.student_visible? &&
+  #   !graded_at.nil? && !submitted_at.nil? && graded_at < submitted_at
+  # end
+
   scope :resubmitted, -> {
     includes(:grade, :assignment)
     .where("grades.student_visible = true")
@@ -108,10 +121,10 @@ class Submission < ActiveRecord::Base
 
   # this is transitive so that once it is graded again, then
   # it will no longer be resubmitted
-  def resubmitted?
-    submission_grade && submission_grade.student_visible? &&
-    !graded_at.nil? && !submitted_at.nil? && graded_at < submitted_at
-  end
+  # def resubmitted?
+  #   submission_grade && submission_grade.student_visible? &&
+  #   !graded_at.nil? && !submitted_at.nil? && graded_at < submitted_at
+  # end
 
   # Getting the name of the student who submitted the work
   def name
--- a/app/models/submission.rb
+++ b/app/models/submission.rb
@@ -29,10 +29,10 @@ class Submission < ActiveRecord::Base
   end
 
   scope :by_active_individual_students, -> do
-    individual
-      .joins(student: :course_memberships)
-      .where(student: { course_memberships: { active: true }})
-  end
+      individual
+        .includes(student: :course_memberships)
+        .where(student: { course_memberships: { active: true }})
+    end
 
   scope :ungraded, -> do
     includes(:assignment, :group, :student)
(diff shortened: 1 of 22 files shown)


app/services/creates_or_updates_submission/updates_resubmission_flag.rb
```diff
@@ -0,0 +1,33 @@
...
+      executed do |context|
+        assignment = context[:assignment]
+        submission = context[:submission]
+        submission.resubmission = true if find_individual_grade(assignment, submission) || find_group_grade(assignment, submission)
+        submission.save
+      end
+
+      private
+
+      def self.find_individual_grade(assignment, submission)
```
jonathoy: These methods can be cleanup up I think. We have a lot of duplication here and on line 11, since you're doing an `||` it's going to do two queries even though only one of them makes sense in the context.

A suggestion might be to do something like this:

```
...
submission.update(resubmission: true) if current_grade_exists?(assignment, submission)
# update also saves so you can get rid of the explicit `submission.save`
...

private

def self.current_grade_exists?(assignment, submission)
  if assignment.is_individual?
    Grade.where(assignment_id: assignment.id, student_id: submission.student_id).student_visible.present?
  else
    Grade.for_group(assignment, submission.group).student_visible.present?
  end
end
```

I left out the the checks for `submission.student_id.nil?` and `submission.group.nil?` because the submission model has a validation that would have caught these conditions already.

app/services/creates_or_updates_submission/updates_submission_params.rb
```diff
@@ -0,0 +1,18 @@
+module Services
+  module Actions
+    class UpdateSubmissionParams
+      extend LightService::Action
+
+      expects :assignment, :submission
```
jonathoy: Minor thing, but since you never use the assignment in this action you shouldn't have to expect it.

spec/models/submission_spec.rb
```diff
@@ -169,6 +169,26 @@
     end
   end
 
+  describe ".resubmitted" do
+    before { Submission.destroy_all }
+
+    it "returns only submissions where resubmitted is true" do
+      submitted_submission = create(:submission, assignment: assignment, student: student)
```
jonathoy: There's a lot going on in this test and I think for the sake of separating concerns, this probably shouldn't be checking that the service changes anything. What the service is supposed to do should be a test that exists in the service spec. If you were to write a test for the scope, it should only ensure that the scope returns submissions where the resubmission flag is true.

Since the scope is just  `scope :resubmitted, -> { where(resubmission: true) }`, I think that it would be fine for you to omit the scope test in `submission_spec` and just add tests elsewhere.

spec/models/submission_spec.rb
```diff
@@ -169,6 +169,26 @@
...
+      grade = create(:student_visible_grade, assignment: assignment, student: student)
+      grade.graded_at = grade.created_at
+      grade.save!
+      Services::CreatesOrUpdatesSubmission.creates_or_updates_submission assignment, submitted_submission
+
+      expect(Submission.resubmitted).to eq [submitted_submission]
+    end
+
+    it "returns only submissions where resubmitted is true" do
+      submitted_submission = create(:submission, assignment: assignment, student: student)
```
jonathoy: 👍

spec/models/submission_spec.rb
```diff
@@ -274,9 +294,9 @@
 
   describe ".resubmitted" do
     it "returns the submissions that have been submitted after they were graded" do
-      grade = create(:student_visible_grade, submission: submission, graded_at: 1.day.ago)
+      grade = create(:student_visible_grade, assignment: assignment, submission: submission, student: student, graded_at: 1.day.ago)
       submission.submitted_at = DateTime.now
-      submission.save
+      Services::CreatesOrUpdatesSubmission.creates_or_updates_submission assignment, submission
```
jonathoy: Same here, you probably shouldn't be calling the service in this test.

spec/models/submission_spec.rb
```diff
@@ -303,16 +323,16 @@
...
       group = create(:group, assignments: [submission.assignment])
       group.students << [student1, student2]
-      grade1 = create(:student_visible_grade, submission: submission, student: student1, graded_at: 1.day.ago)
-      grade2 = create(:student_visible_grade, submission: submission, student: student2, graded_at: 1.day.ago)
+      grade1 = create(:student_visible_grade, assignment: assignment, student: student1, group: group, graded_at: 1.day.ago)
+      grade2 = create(:student_visible_grade, assignment: assignment, student: student2, group: group, graded_at: 1.day.ago)
       submission.submitted_at = DateTime.now
       submission.group_id = group.id
-      submission.save
+      Services::CreatesOrUpdatesSubmission.creates_or_updates_submission assignment, submission
```
jonathoy: See prior comment

app/models/submission.rb
```diff
@@ -108,10 +122,10 @@ def will_be_resubmitted?
 
   # this is transitive so that once it is graded again, then
```
jonathoy: This is something that raises a question: There is a comment here that says that it's a transitive expression, so that it's no longer a resubmission after it's been graded and submitted. Did this still need to apply in the places where you replaced this method?

evanstraub:
QA Testing
Assignments - with no resubmission
As a student
- [ ] Start assignment, saves as draft, submit assignment
- [ ] Start assignment, wait a while, submit assignment

Assignments - with resubmissions
- [ ] just submit the assignment
- [ ] submit assignment, grade assignment, resubmit assignment
- [ ] start assignment, save as draft, submit assignment, grade assignment, resubmit assignment
- [ ] submit assignment, grade assignment, resubmit assignment, regrade assignment, resubmit again
- [ ] start assignment, save as draft, return later, submit assignment, grade assignment, start resubmission, save assignment, return later, resubmit.
- [ ] start assignment, save as draft, return later, submit assignment, grade assignment, start resubmission, save assignment, return later, resubmit, regrade, resubmit

jonathoy:
Leaving the branch `3899-simplify-resubmissions` for now for future reference if we decide to revisit this work. In the interim, the minimal set of changes needed to address the bug has been merged via #3961

PR #3944, opened 2018-04-02
Submissions after accept until

### LINKED ISSUE #3943: Submissions can still be submitted after assignment is closed

jonathoy:
### Status
READY

### Description
As reported by Evan, it was possible for a student to submit an assignment even though the due date and accept until date were in the past.

This PR restricts the modification or creation of submissions on the controller itself (for methods `new`, `create`, `edit`, and `update`) and returns an error if the assignment is not open.

Submissions

======================
Closes #3943

### DIFF
--- a/app/controllers/submissions_controller.rb
+++ b/app/controllers/submissions_controller.rb
@@ -61,9 +61,11 @@ class SubmissionsController < ApplicationController
     respond_to do |format|
       if submission.update_attributes(submission_params.merge(submitted_at: DateTime.now)) && Services::DeletesSubmissionDraftContent.for(submission).success?
         submission.check_and_set_late_status! unless submission.will_be_resubmitted?
-        path = @assignment.has_groups? ? { group_id: submission.group_id } :
-          { student_id: submission.student_id }
-        redirect_to = assignment_submission_path(@assignment, submission, path)
+        
+        redirect_to = assignment_submission_path @assignment,
+          submission,
+          @assignment.has_groups? ? { group_id: submission.group_id } : { student_id: submission.student_id }
+
         if current_user_is_student?
           send_notification(submission.id, submission_was_draft) if @assignment.is_individual?
           redirect_to = assignment_path(@assignment, anchor: "tabt1")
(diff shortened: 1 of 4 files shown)


jonathoy:
We have so many partials and so much view logic that I'm not able to find where exactly one might be seeing a button to submit after the due date, so I just restricted it from the controller side.

app/controllers/submissions_controller.rb
```diff
@@ -81,29 +86,38 @@ def update
...
-    else
-      NotificationMailer.updated_submission(submission_id).deliver_now
-    end
+  def find_assignment
+    @assignment = current_course.assignments.find params[:assignment_id]
+  end
+
+  def ensure_assignment_open?
+    redirect_to assignment_path(@assignment), alert: "The assignment is no longer open for submissions" \
+      unless @assignment.open?
```
acousticrobot: Do you want to return true here?

app/controllers/submissions_controller.rb
```diff
@@ -81,29 +86,38 @@ def update
...
   end
 
-  def ensure_editable(submission, assignment)
+  def ensure_editable?(submission, assignment)
     redirect_to assignment_path(assignment, anchor: "tabt1"),
```
acousticrobot: I don't think these should end in a question mark. Even though they return a boolean, they have side effects (redirects), so they don't fit the boolean method pattern, and I don't think I've ever seen a before_action method use question marks before.

app/controllers/submissions_controller.rb
```diff
@@ -81,29 +86,38 @@ def update
...
 
   private
 
-  def send_notification(submission_id, submission_was_draft)
-    if submission_was_draft
-      NotificationMailer.successful_submission(submission_id).deliver_now
-    else
-      NotificationMailer.updated_submission(submission_id).deliver_now
-    end
+  def find_assignment
```
acousticrobot: 👍 Nice cleanup here!

app/controllers/submissions_controller.rb
jonathoy: I didn't think it was needed? It doesn't look like we return true explicitly on any of the role-based filters such as `ApplicationController#ensure_admin?`

app/controllers/submissions_controller.rb
jonathoy: I can change this if it's preferred, but it follows the pattern of methods such as `ApplicationController#ensure_student?`, `#ensure_not_impersonating?`, etc which are all used for `before_action` filters that redirect.

PR #3951, opened 2018-04-05
Angularize Student Table on Assignment Show Page

jonathoy:
### Status
**READY**

### Description
animated
For individual assignments. Loads student table and info asynchronously and in batches of 25.

- Includes the addition of a common Angular directive to make table headers sortable and some refactoring to the shared team selector component.
- Adds shared service that maintains search criteria for an Angularized table
- Adds new API for releasing grades (note that old grade release route can likely be removed after group assignment refactor is finished)
- Adds API for fetching student information, with option to pull data in batches

**Notes to this approach: We add extra API routes to duplicate logic that was previously handled on page submits. We need to investigate and do some cleanup once this work is more throughly completed.**

### Gem dependencies
N/A

### Migrations
NO

Test the various conditions for which information might be displayed for a student on the assignment show table. The table headers provide a good basis possible test cases.

Assignment show page for individually-graded assignments

======================
Part one of fixing assignment show page to improve performance #3880


jonathoy:
Request: once all students are loaded, sort the collection by `last_name`.

maskib:
This looks good! I've found one glitch that needs fixing - sorting by "grade" or "score" does not work.

PR #3955, opened 2018-04-09
Don't clear out uploader media

### LINKED ISSUE #3952: Badge Icons Not Being Properly Copied During Course Copy
Narrowed down issue to an obscure [bug](https://github.com/carrierwaveuploader/carrierwave/pull/1544#ref-issue-311745163) with Carrierwave:

Requires us to update Carrierwave to ~> 1.0

jonathoy:
### Status
**READY**

### Description
Removes added lines that were meant to clear out the uploader media on the copy model prior to copying the actual media in S3.

Due to a bug in Carrierwave, the mounted uploader retains a reference to the original model instead of the copy and this in turn causes issues.

======================
Interim solution to #3952

### DIFF
--- a/app/models/assignment.rb
+++ b/app/models/assignment.rb
@@ -335,7 +335,6 @@ class Assignment < ActiveRecord::Base
 
   # Copy assignment media
   def copy_media(copy)
-    copy.update(media: nil, remove_media: true)
     remote_upload(copy, self, "media", media.url)
   end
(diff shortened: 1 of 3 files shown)


jonathoy:
This is an interim "fix" to prevent the original models from being altered. The long-term solution for resolving the actual copying of these files will come at a later time.

PR #3971, opened 2018-04-27
Student index improvements

jonathoy:
### Status
**READY**

### Description
Refactors view as a front-end component in Angular, for async and batch loading.

Observe that when you visit the see all students page, all of them load in batches with their intended information. Search and team filters should also work properly.

Students index page

======================
Addresses performance concerns for student index, as referred to in #3880


jonathoy:
When you deselect a selected team filter, the list does not repopulate. Presumably this is because the option value for the default filter is not null.

evanstraub:
I'm not sure how/if this is a thing - in a large class, the browser pretty much completely hangs as it is i loading at a certain point.  It almost looks like the page has crashed, but it will load eventually.  This is really only going to be an issue in really large classes like TG.  It's seeming to hang around 525 and then jumps to all 900

jonathoy:
I've fixed bottom10, but the lag issue is because there's too much rendered HTML on the page. The options there are to either add pagination or to not show the table results until all results are completely loaded (could still show the progress indicator).

jonathoy:
Based on our discussion, we are going to merge the batch loading work for now while we continue to evaluate adding pagination abilities.

PR #3977, opened 2018-05-09
3975 Google Create Account

### LINKED ISSUE #3975: Google Create Account
Make it so that if a new user attempts to authenticate through Google for the first time, that user has an account created.

dovek:
### Status
**READY**

### Description
When a user comes to GradeCraft for the first time and the user doesn't have an account and attempts to login via Google, the application will catch that the user doesn't have an account and make one for the user.

### Related PRs
N/A

### Deploy Notes
N/A

### Gem dependencies
N/A

### Migrations
NO

Attempt to log in to Gradecraft via the button on the dashboard as well as through the trial account page. Both routes should direct you as needed to a page where you are asked whether you are an student or instructor. The latter choice will lead you to the course creation page and log you in.

Login

======================
Closes #3975


jonathoy:
Documenting some questions that came up regarding workflow and design:

1. Do we want to direct the user to the external account creation form where they have to input a password if the intent is that they will never need to use that password to log in? I can imagine a workflow more similar to how LTI works where if an account does not already exist for the email, it will create one behind the scenes.

2. How might we restrict the account creation process? When I think about this functionality, I think of a few scenarios that the user could take to get to the Google login page.

- **Option 1**: The user simply clicks on the "Sign in with Google" button on the login screen.
Do we want to even allow this if we cannot infer what role they will have in Gradecraft? This works fine for creating a new `user` account, but upon successfully creating the account they will be presented with an error stating that they don't belong to any courses. What then?

If we allow this, we may want to check if there are any unintended side effects to allowing it to happen on Umich (edge case, but I can currently go to `umich.gradecraft.com` and click on Admin -> Sign in with Google.

- **Option 2**: A user is directed to the Sign in with Google page via a course invite (email). If this were the case, we could probably infer the user's role and also create a course membership for them after they sign in the the first time.

This brings about more questions: This probably works fine for students - under what conditions can an instructor create a new Gradecraft account via Google sign on? If they don't have an existing account, would you also go and create a blank course shell for them? If so, this then mirrors the trial account creation process which you may not want to happen on Umich.

app/controllers/user_sessions_controller.rb
```diff
@@ -15,6 +15,10 @@ def instructors
     @user = User.new
   end
 
+  def confirmation
+    @user = User.new(username: params[:email], email: params[:email], first_name: params[:first_name], last_name: params[:last_name])
```
jonathoy: We mentioned defining strong params here but we aren't using them. You can optionally change this to the following **if** you didn't change the method before so that the attributes are stored in the session and not in the params:

```
@user = User.new user_params
```

app/controllers/users_controller.rb
```diff
@@ -54,6 +54,9 @@ def new_external
       redirect_to new_user_path
     else
       @user = User.new
+      @first_name = params[:first_name]
```
jonathoy: Is this still needed or was this introduced in the early stages of development?

app/controllers/users_controller.rb
```diff
@@ -175,6 +188,10 @@ def activate_set_password
     redirect_to root_path, alert: "Invalid activation token. Please contact support to request a new one." and return unless @user
   end
 
+  def activate_via_google
+    redirect_to "/auth/google_oauth2/" and return
```
jonathoy: I don't think an extra method needs to exist just to redirect. Where you refer to it in the email template, you can just do something like:

```
# "http://www.example.com/auth/google_oauth2"
"#{app.root_url}auth/google_oauth2"
```

app/views/user_mailer/activation_needed_email.html.haml
```diff
@@ -7,3 +7,6 @@
 
 %p{ :align => "center" }
 = link_to "Activate Your Account", activate_set_password_user_url(@user.activation_token).html_safe
+
+%p{ :align => "center" }
+= link_to "Activate Your Account Through Google", activate_via_google_user_url(@user.activation_token).html_safe
```
jonathoy: It doesn't seem like we need to pass the activation token anymore.

app/controllers/users_controller.rb
```diff
@@ -73,6 +76,16 @@ def create_external
     end
   end
 
+  def create_via_google
+    @user = User.new(username: params[:email], email: params[:email], first_name: params[:first_name], last_name: params[:last_name])
```
jonathoy: This can be done via strong params, if desired.

app/controllers/users_controller.rb
```diff
@@ -73,6 +76,16 @@ def create_external
...
+    if @user.save
+      @user.activate!
+    else
+      render :new_external
```
jonathoy: This save probably shouldn't fail, but if it does we may just want to direct them to an error message otherwise the workflow will be confusing.

app/views/user_sessions/_form2.html.haml
```diff
@@ -0,0 +1,10 @@
+.login
+  %h1 Welcome to GradeCraft
+  %p It looks like you are trying to authenticate with Google, but you do not have an account. Are you trying to login as a Student or an Instructor?
+  = simple_form_for @user, :url => create_via_google_users_path, :method => :post do |f|
+    = hidden_field_tag :email, params[:email]
+    = hidden_field_tag :username, params[:username]
```
jonathoy: Optionally, you can remove some clutter by not having to set the username since on the backend you can just make it the same as the email.

app/views/user_sessions/instructor_confirmation.html.haml
```diff
@@ -0,0 +1,5 @@
+- content_for :public_class, 'public-with-background'
```
jonathoy: Is this view needed?

app/views/users/new_external.html.haml
```diff
@@ -6,9 +6,9 @@
       %h1.bold First, let's create your account:
 
       = simple_form_for @user, url: create_external_users_path, method: :post do |f|
-        .form-item= f.input :first_name
-        .form-item= f.input :last_name
-        .form-item= f.input :email
+        .form-item= f.input :first_name, input_html: { value: @first_name }
```
jonathoy: Simple form should populate these fields automagically without the `input_html` options if in the render method you have pre-populated `@user`. That way you also don't need to create `@first_name`, `@last_name`, and `@email`

config/routes.rb
```diff
@@ -275,9 +275,11 @@
     member do
       get :activate
       get :activate_set_password
+      get :activate_via_google
       get :resend_activation_email
       put :manually_activate
       post :activate_set_password, action: :activated
+      post :activate_via_google, action: :activated
```
jonathoy: We have two routes here that link to the same action. Either one of these routes are redundant or there is breaking logic here for `activate_set_password`.

config/routes.rb
```diff
@@ -323,12 +326,14 @@
   get :login, to: "user_sessions#new", as: :login
   get :logout, to: "user_sessions#destroy", as: :logout
   get :reset, to: "user_sessions#new"
-  resources :user_sessions, only: [:new, :create, :destroy] do
+  get :confirmation, to: "user_sessions#confirmation", as: :confirmation
```
jonathoy: Routes got a little bit messy here - some extras were added that don't need to be. Note that what I have below will change the names that you called them but it's resourceful and cleaner.

```
resources :user_sessions, only: [:new, :create, :destroy] do
  collection do
    get :confirmation
    get :student_confirmation
    get :instructors
  end
end

...

# revert the change you made on line 336 where you added confirmation, since I'm not seeing a confirmation method in the passwords controller
resources :passwords, except: [:new, :destroy, :index, :show]
```

PR #3979, opened 2018-05-10
3772 predictor weights

### LINKED ISSUE #3772: Weighted assignments in predictor interactive update
### Expected Behavior
In predictor: Completed assignments in a weighted category should have their color code update with changes to the weights on the Assignment Type

### Actual Behavior

 If I clear out all weights, the grade (no longer included) still stays green. The reverse is also true - if I weight an assignment type, the grades stay greyed out until I refresh. (Total points in the barchart update appropriately)

image

See #3547

acousticrobot:
### Status
**READY**

### Description
Fixes a couple issues with weighting in the predictor.

  * removes a note about weights not being available in classes that have no weights.
  * Always colors grades as green, even when there are no weights.

As discussed with Evan, the more elaborate solution of changing the color based on the weights as I described in the original ticket would take more work, and might not be appropriate.

======================
Closes #3772


jonathoy:
@evanstraub Since I'm unsure of the scope of the work that was discussed here, could you verify that this looks good? From a technical perspective, I'm happy to merge this if ready.

PR #3980, opened 2018-05-10
remove fixed height from rubric

acousticrobot:
### Status
**READY**

### Description

Evan had remarked that rubrics are always really big when doing demonstrations, even if there is very little description in the box.

@jonathoy please just confirm that it is OK to remove this line, and that it doesn't mess anything up with learning objectives which is when it was added.


jonathoy:
This looks good to me. At some point, I'd love to ideally clean up the shared styling between rubrics and learning objectives so that this doesn't happen again.

PR #3981, opened 2018-05-11
3563 uncomment onboarding access on every page

### LINKED ISSUE #3563: Onboarding Partials only when needed

acousticrobot:
### Status
**READY**

### Description

PR #3711 attempted to make the onboarding slideshow more performant by only loading it when the student had not yet seen it. The issue is, there is an access button to the slideshow which allows the student to view it again at any time.  The most performative and simplest way to make this work is to have the slideshow preloaded as Marie originally designed it.

In this PR, I have uncommented the slideshows so that they are available again to students and faculty, and they load when a student logs into a course for the first time.

I tried to run benchmarks on the same page with and without the slideshow, and noticed no meaningful differences with or without the slideshow.*

======================
Closes #3563 
Closes #3711

### Time Trials

*I reloaded the dashboard 8 times each with and without the slideshow.  The difference between each trial was actually greater than any difference with or without the slideshow. Loading the dashboard varied between 6.5 and 12 seconds, with both the fastest and slowest times recorded without the slideshow. All times were taken after an initial load to discount any differences due to caching. 

Load times for dashboard without slideshow 8 trials, ordered:
[6423, 9457, 9842, 10657, 11037, 11588, 11781, 12295] 
average: 10385ms 

Load times for dashboard with slideshow, ordered :
[6656, 7138, 9709, 9967, 10972, 11424, 11722, 11919] 
average: 9938ms

### DIFF
--- a/app/assets/javascripts/onboarding.js
+++ b/app/assets/javascripts/onboarding.js
@@ -115,7 +115,9 @@ if (m && p) {
     mClose.addEventListener('click', modalClose);
 
     // close modal by modal action btn click/hit
-    mAction.addEventListener('click', modalClose);
+    if (mAction) {
+      mAction.addEventListener('click', modalClose);
+    }
 
     // close modal by keydown, but only if modal is open
     document.addEventListener('keydown', modalClose);
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -208,6 +208,10 @@ class User < ActiveRecord::Base
     name.downcase == another_user.name.downcase
   end
 
+  def onboarded?(course)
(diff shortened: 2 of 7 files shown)


jonathoy:
This appears to be causing some Javascript issues in console:

screen shot 2018-05-18 at 2 32 13 pm

jonathoy:
@acousticrobot When you have a chance, could you rebase this against master to see if you can reproduce any Javascript errors that were the result of this branch?

PR #3992, opened 2018-06-05
Remove deprecated Mongod argument

jonathoy:
### Status
**READY**

### Description
While trying to set up Gradecraft locally, it was found that the `--rest` argument is deprecated in its current use in the Procfile.

This results in a failure if we ever want to upgrade Mongodb or we accidentally install the wrong version of Mongodb for the next local setup (currently the workaround is to ensure that we `brew install mongodb@3.4`)

This PR removes it, since it does not appear that we need this argument at all.

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -2,4 +2,4 @@ web: bundle exec puma -p 5000
 redis: redis-server --port $REDIS_PORT
 resque_scheduler: bundle exec rake resque:scheduler
 resque_worker: bundle exec rake resque:work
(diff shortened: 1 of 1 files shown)


jonathoy:
@acousticrobot I tried to `git blame` this to the original commit where the argument was added to see if I could find out the origin and I noticed that the commit was made by you.

It's been some time now, but do you have any thoughts or remembrance about why this was needed?

PR #3998, opened 2018-06-08
Update to Ruby 2.5.1

### LINKED ISSUE #3916: Address Ruby 2.2 EOL

jonathoy:
### Status
**READY**

### Description
Updates Ruby to latest (2.5.1)
Updates relevant gems

**NOTE:** Once this work is merged, all developers working on the project should update Bundler to 1.16.2 to avoid Gemfile.lock conflicts and will need need to install and set Ruby 2.5.1 using their preferred choice of Ruby version manager

```
gem update bundler -v 1.16.2

# using rbenv
rbenv install 2.5.1
rbenv local 2.5.1 (or rbenv global 2.5.1)
```

### Deploy Notes
The environment-specific base images must be built with Ruby 2.5.1 prior to deploying this work. See base_image repo for additional details.

It is also a good idea for developers on the project to run `gem update --system` while they are updating bundler and Ruby to ensure that the latest security updates are installed. 

So many things.

======================
Closes #3916, relates to #4016


jonathoy:
Specs passed?!

jonathoy:
Investigate at some point if we can remove `gem 'angularjs-rails', '~> 1.4.2'` entirely and swap out for `gem 'rails-assets-angular-animate', '1.3.15'`, since at a glance it appears that we try to load two different versions of Angular at once.

PR #4002, opened 2018-06-13
Revise paginated loading for Canvas API

### LINKED ISSUE #4003: Canvas Grade Import Stuck If Multiple Batches
Unsure if the Canvas API changed.....

But we previously operated under the assumption that if there were multiple [pages](https://canvas.instructure.com/doc/api/file.pagination.html), the linked URL in the header with `rel=next` would look something like this:

```
/submissions?page=1&per_page=3
```

where now it looks like this:

```
/submissions/page=bookmark:WzD4DQ&per_page=1
```

Since on our end we fetched pages asynchronously by incrementing the page number, this functionality now is broken.

jonathoy:
### Status
**READY**

### Description
Feeling like this is a change with the Canvas API but not 100% sure....

Previously our expectation was that if we paginate via the API and there were more pages to be fetched, the provided link in the header would contain something to the effect of `page=1&per_page=10`.

In this scenario, we would previously fetch additional pages by incrementing the page number for the next call. However, now the page number returned by Canvas in the links header now resembles `page=bookmark:WzD4DQ&per_page=1`.

This PR changes how the params are passed around from the server-side, to the client, and back to server between AJAX calls so that it no longer operates simply off of the assumption that the page is an int.

Canvas grade import
Canvas user import

======================
Closes #4003

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -370,11 +370,6 @@ Style/LineEndConcatenation:
                  line end.
   Enabled: true
 
-Style/MethodCallWithoutArgsParentheses:
-  Description: 'Do not use parentheses for method calls with no arguments.'
-  StyleGuide: 'https://github.com/bbatsov/ruby-style-guide#no-args-no-parens'
-  Enabled: true
-
 Style/MethodDefParentheses:
   Description: >-
                  Checks if the method definitions have or don't have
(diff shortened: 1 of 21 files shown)


jonathoy:
An observation that was made during this was that the bug caused a lot of server load when it continued to try and fetch the same page over and over again.

An optional thing that we could do that wouldn't necessarily be the worst would be to add the gem `rack-attack`. I tested this locally (super easy to implement) and set a throttle limit by IP, which basically prevents a given user from generating too many requests within some period of time.

It seems like the added overhead would be pretty minimal - what would be difficult to determine would be the proper throttling rate, as we don't want to cause problems for our users like TG who can hit 20+ requests in a short period of time when loading for instance all students for the course in batches.

jonathoy:
I'm not crazy - finally got a response from Instructure:

Thank you for your reply. I was able to do some more research and it does appear that the addition to bookmark was an intentional change that our team put into place in order to help the submission api return the information quicker as it will slow down if there are quite a few pages of data.

Since the addition of the bookmark within the pagination of the submissions api, you may need to adjust the calls you are using to not rely on an integer but work around the bookmark instead as this does appear to be an intended, permanent adjustment to the call.

Hopefully this information helps! If you do have any further questions, please let us know!

PR #4012, opened 2018-07-04
prevent menus from overlapping with tables

chcholman:
### Status
**READY**

### Description
A tiny UX improvement that adds a margin to prevent context menus from overlapping with the tables beneath


jonathoy:
screen shot 2018-07-05 at 10 54 28 am

Is this overlapping what you were trying to fix? I think this still appears for me on the assignment index page. What if we added `overflow: hidden` to `_layout.sass` instead?

```
.page-header
  padding: 0.5rem 1rem
  overflow: hidden
```

PR #4017, opened 2018-07-11
3990 assignment export

### LINKED ISSUE #3990: Assignment structure Export for moving assignments between courses
### Expected Behavior
Instructors should be able to get a full export of assignment structure to move/share between courses without having to make a copy and then be able to import the assignment structure.  (or at least a good portion of it).  Ideally, the format for the assignment structure export would be the same format as what is required for the import.

### Actual Behavior
The current structure is:  Assignment ID, Name, Assignment Type, Point Total, Description, Open At, Due At, Accept Until, Submissions Count, Grades Count, Created At, Learning Objectives

It leaves our "Assignment Purpose"  "Accepts Submissions" and "Required" (we could discuss if we need all of these).

Currently, the import file takes:  
Assignment Name, Assignment Type, Point Total, Description, Due Date (mm/dd/yyyy hh:mm:ss am/pm)

It would be helpful if these two files were closer aligned as well as some additional information that would make it a more seamless process for exporting and importing.
-- | -- | -- | -- | --

acousticrobot:
### Status
**IN DEVELOPMENT**

### Description

Aligns imports and exports, and adds additional fields that are missing in the current import.

All export fields that are not used in the import have been moved to the end, and will be ignored in the import. This allows exports from one class to be imported directly into another.

  * We need to verify the new export structure is OK with current users.
  * Verify that exports work as imports
  * Verify that the import sample file works as a starting point for creating an import.

======================
Closes #3990


jonathoy:
@acousticrobot Is this ready for review?

jonathoy:
- [ ] Dates are in various timezones
- [ ] In the export, return only the dd/mm/yyyy and no time. In the import, assume 11:59 for the time and investigate the timezone output
- [ ] Verify that the export can be directly imported without any modifications

PR #4023, opened 2018-07-27
Fix user form issues

### LINKED ISSUE #4022: In App, Upon error when adding new user to course, the form gets glitchy
### Expected Behavior
After "1 error prohibited this user from being saved:
Email is invalid", I would expect the page layout to be the same as it was before.
### Actual Behavior
Instead, the "communication design profile" section of the page repeats over and over again, literally 1,230 times!!

### Steps to reproduce
 In app.gradecraft.com, add a new user to a course, and input an invalid email. Then, scroll down to see all of the repeats of the "communication design profile" section.

jonathoy:
### Status
**READY**

### Description
Fixes an issue with the user form where if you triggered a form error while trying to create or update an existing user, it would render one subset of form inputs for every course membership the user belonged to.

======================
Closes #4022


maskib:
This seems to work now. I am getting a notice of two errors instead of one, and I'm not sure what the second error actually means: 

screen shot 2018-07-30 at 1 42 33 pm

PR #4078, opened 2018-08-29
Upgrade gems

jonathoy:
### Status
**READY**

### Description
Updates various gems including Rails from 5.2.0 to 5.2.1 and most notably, Redis to 4.0.2.

The latter was performed in anticipation for our move from using memcached as an application cache to Redis.

======================
Closes #[ISSUE NUMBER]


jonathoy:
Upgraded `scrypt` as well, since we appear to be getting more and more frequent segmentation faults in Codeship during our CI builds. A quick Google search shows that this generally is due to issues related to gems with C-extensions and scrypt is one of those gems.

PR #4079, opened 2018-08-29
Grade scheme element setup

jonathoy:
### Status
**READY**

### Description
The autosave functionality can be a little weird if you are setting things up from scratch. Since it won't save until the entire form is valid, it can become a little confusing for the user if they are anticipating changes to be saved automatically after some period of time.

This PR aims to address that by making the autosaves more granular, in the sense that any single element will be saved onBlur as soon as it is considered valid.

Grade scheme elements

======================
Part 2 of #4081

### DIFF
--- a/app/assets/javascripts/angular/directives/grade_scheme_elements/element.coffee
+++ b/app/assets/javascripts/angular/directives/grade_scheme_elements/element.coffee
@@ -1,57 +1,50 @@
 # Renders a single grade scheme element in the grade scheme element mass edit form
 # Note: This directive can exist only as a child of the gradeSchemeElementsMassEditForm
-@gradecraft.directive 'gradeSchemeElement',
-['GradeSchemeElementsService', 'DebounceQueue', '$timeout', (GradeSchemeElementsService, DebounceQueue, $timeout) ->
-  {
-    scope:
-      gradeSchemeElement: '='
-      index: '='
-    require: '^^gradeSchemeElementsMassEditForm'
-    templateUrl: 'grade_scheme_elements/element.html'
-    restrict: 'E'
-    link: (scope, element, attrs, gseForm) ->
-      _validateElements = () ->
-        GradeSchemeElementsService.validateElements()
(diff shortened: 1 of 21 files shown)


jonathoy:
Two questions come up here:

- Do we want the ability to have elements reorder themselves in the list according to their point threshold? If so, it doesn't make sense to have the button to `+ Add a Level Below`.
- This PR makes the autosaving more granular so that it will persist a single element as soon as it's valid. Should we direct attention to the user if they try to leave the page and they haven't yet resolved one or more invalid elements?

jonathoy:
@maskib Michael, would you kindly help me with some QA testing on this work? It should be in staging now. Feel free to try to break it any way you can and let me know if you have suggestions for improving the user experience. The intent for this work was to improve how the autosave functionality works.

Feel free to also do some regression testing when starting from scratch as well and setting up grade scheme elements when there are none (via the setup form).

Thanks!!!

maskib:
I added a bunch of levels below the starter level “not on the board yet”. After refreshing the page, they all have the same “not on the board yet” name, which I think is fine, but they also all have the “remove” button greyed out, which seems like a bug.

maskib:
If you start from scratch and have it create a bunch of different grade levels for you, but don’t fill them all out before refreshing the page, any that you didn’t add point values to will disappear. I think there should be an ability to save even for grade levels that do not have a threshold value yet. Or at least there should be some sort of warning letting you know that grade levels without point thresholds will disappear if you navigate away.

maskib:
Can we make the “level name” text field wider and push over the “points” text field? There’s a bunch of free space to the right of that that we could be using.

maskib:
Maybe there should be an alert confirming that you actually want to delete the grade element that you clicked "remove" on?

jonathoy:
@maskib Agh! I can't reproduce that scenario where you had multiple zero elements, even though I saw you do it...

chcholman:
@jonathoy Can you help me figure out what the next step is for this PR? Does it need further review/testing?

PR #4080, opened 2018-08-30
Add interim save button

jonathoy:
### Status
**READY**

### Description
As an intermediate solution, adds a button to manually trigger a save when editing grade scheme elements. Follow up in response to some observed issues relating to the autosaving of work.


maskib:
This seems to work fine for me!

PR #4101, opened 2018-09-13
Update Froala to latest, use rails-assets

jonathoy:
### Status
**READY**

### Description
Updates Froala to the latest release and removes the minified files in the project, using instead assets from rails-assets.org (for now).

Anywhere we use a froala editor (e.g. submissions, grading feedback, etc.)


jonathoy:
The latest version of Froala includes FontAwesome 5. We use 4 everywhere else, so we may want to do some testing to ensure that they don't cause issues. We can look into what it would take to move completely off of our FA 4 dependency.

PR #4143, opened 2018-11-19
Admin Page For Retroactively Triggering Unlock Checks

jonathoy:
### Status
**IN DEVELOPMENT**

### Description
This PR provides an interface for which an admin can go and manually trigger the process for checking whether unlock conditions have been met. The typical use case for this is when an instructor has gone and midway altered the conditions for an unlock after the course has started.

### Gem dependencies
N/A

### Migrations
NO

Visit /admin/tasks/unlocks while logged in as an admin. The route will not exist if visiting the route as an unauthenticated user or as someone without the necessary role.

The logic for doing this has been extracted into a helper class, in the event that we want to build this in so that it is automatically triggered on a condition change. For now, we are making this a manual process since it has a potential for being expensive to perform.

Unlocks

======================
Closes #[ISSUE NUMBER]


jonathoy:
Per Barry and Cait's suggestion, we might want to think of some text to add to the unlock conditions section so that instructors are aware that changing conditions midway through a course can cause side effects.

jonathoy:
@evanstraub This is a quick implementation for the ability to re-trigger unlocks but it should work for your needs. Did you still have any thoughts about possibly adding text to warn instructors about changing unlocks midway through a course? Otherwise I may just merge this.

PR #4148, opened 2018-11-28
added auto award and email notification functionality to LO model

### LINKED ISSUE #4139: Ensure LO compatibility with unlock emails and auto awarded badges
Once #4113 is merged, we need to make sure that we also are sending notifications on unlocks and that unlocked badges are auto-awarded when the condition is based upon the completion of a learning objective.

ebarroso7:
### Status
**Ready**

### Description
Added the ability to have a learning objective unlockable condition that automatically awards a badge.
Also added email notification to the achievement of a learning Objective

### Related PRs

### Todos
- [x] More user testing for emails and functionality -- Make sure students are emailed upon completion of a learning objective
- [x] Email notification saying that the user has completed the learning objective
- [x] QA / User testing. See below or ask me (Erik) if you have any questions
- [x] Add rspec tests

### Migrations
YES -- Made migration to add email_learning_objective_achieved to course_memberships & achieved to learning_objectives_cumulative_outcomes

Create or Edit a badge to have a learning objective as the unlockable condition. Underneath while editing check the 'Auto award on unlock' box. Then have a student achieve the learning objective. This results in the badge getting awarded and email notifications sent automatically.

In addition to the badge getting auto awarded, this PR also includes functionality around emailing a user when a learning objective is achieved for the first time. 

To ensure an email is sent upon the completion of a Learning Objective (LO) the email setting must be checked on /users/edit_profile (reachable by clicking on your name in the top right corner and clicking 'my account'). 

Now, the achievement of a LO should also email the student saying it has been achieved.

Things to make sure they work properly --
* Achievement of LO sends email, achieved by either a number of points achieved or assignments completed.  
* receiving a proficiencies for ONE assignment that achieves TWO learning objectives sends TWO emails
* Achieving a learning objective automatically awards a badge and sends appropriate amounts of emails -- Achieving LO by a grade should send grade email, LO achieved email, badge unlocked email, badge awarded email
* Unchecked 'email learning objective achieved' should not send an email upon the completion of a LO

Learning objective model. Email notifications. grade model

======================
Closes #4139 #4033


maskib:
At this point, neither Shaelyn or me were able to get the automatic email notification, unfortunately. Checked my spam folder as well - nothing there, either.

maskib:
Good to go - the expected emails were found in the GC staging account!

app/controllers/users_controller.rb
```diff
@@ -246,6 +246,7 @@ def update_profile
     @user = current_user
 
     up = user_params
+    puts(up.to_h)
```
jonathoy: You'll probably want to remove this

app/views/users/edit_profile.html.haml
```diff
@@ -65,6 +65,8 @@
           = p.input :email_grade_notifications
           - if current_course.has_team_challenges?
             = p.input :email_challenge_grade_notifications
+          - if current_course.has_learning_objectives?
```
jonathoy: It might be a safer idea to use `#uses_learning_objectives?` here for now.

app/models/user.rb
```diff
@@ -192,6 +192,10 @@ def email_challenge_grade_notifications?(course)
     return true if course_memberships.where(course: course).first.email_challenge_grade_notifications?
   end
 
+  def email_learning_objective_achieved?(course)
```
jonathoy: Unless I'm missing something, you aren't actually sending these emails based on whether this is true or not.

app/models/user.rb
```diff
@@ -192,6 +192,10 @@ def email_challenge_grade_notifications?(course)
...
+    return true if course_memberships.where(course: course).first.email_learning_objective_achieved?
```
jonathoy: Your logic here is consistent with the other methods that relate to notifications being sent, but I think there's actually a bit of a misnomer here with the method name/return value. If `course_membership.email_learning_objective_achieved` is `true`, this method will return also return `true`. If not it actually skips that line and returns a default value of `nil`, thus making `email_learning_objective_achieved?` an implied boolean return method that doesn't always return a boolean.

This _could_ have some side effects depending on how it's being used.

Really this line can just be:

```
course_memberships.find_by(course: course).email_learning_objective_achieved?
```

FYI that you can also condense `where().first` down to `find_by()`

app/models/learning_objective.rb
```diff
@@ -30,6 +31,17 @@ def completed?(student)
     progress(student) == COMPLETED_STATUS
   end
 
+  def first_time_achieved?(student)
+    lo_cumulative_outcome = LearningObjectiveCumulativeOutcome.for_user(student.id).for_objective(self.id).first
+    if self.completed?(student) && !lo_cumulative_outcome.achieved
+      lo_cumulative_outcome.achieved = true
+      lo_cumulative_outcome.save
+      return true
+    else
```
ebarroso7: Is this else necessary ?  I want to make sure it returns false in the instance that the if false, and do not want it to return the lo_cumulative_outcome variable

maskib:
Hello, 

This appears to mostly be working with one exception: when I turned off all of my email notifications for the student I was using for this test, one email still showed up in the gc.staging email interceptor: 

Hi Michael,
You have successfully unlocked the badge Auto award test 4 (no email) in your class Michael Skib - testing!
Play On!

Everything else works nicely. One super nitpicky thing I'll point out is this: 
screen shot 2018-12-12 at 5 08 29 pm

The badge was awarded at 5:04 but the email says it was awarded at 10:04? Not sure if that's an issue...just weird...

Thanks!

PR #4149, opened 2018-11-28
Learning objective assignments completed bug

ebarroso7:
### Status
**READY**

### Description

- This PR fixes a bug that has to do with the column 'Assignments Completed' now changed to 'Assignments Submitted'. 

- Also fixes bug where "Learning Objectives Graded" column was not displaying any information.  -- Really shows Assignments that have earned proficient status.

Previously, there would be a difference between the number of reported assignments completed and what the student had actually completed, this was because the service was counting assignments that had been saved as a draft. So the number being shown to students and staff were the number of assignments completed and the number of draft assignments that had been started and saved.

Right now the code is setup so that the students and staff see the number of assignments submitted, excluding drafts. While an admin user still sees the assignments and draft assignments. 

**QUESTION**: Should table be changed so that the admin sees the same number of assignments submitted as students and staff (this would exclude draft assignments from being counted in the admin view) ??

**QUESTION**: Change column "Learning Objectives graded" to "Proficiencies earned" ??

### Todos
- [x] Talk with Evan about admin and staff view discrepancies 
- [x] If we do want to change it, check if it effects other places in the application that call the submission service
- [x] Change "Learning Objectives graded" to more accurately show what is being displayed "Proficiencies earned" ?? 

### Migrations
NO

Visit the Learning Objectives page for a class that has linked assignments. Discrepancy can be made and seen in the admin page when a student has saved a draft to an assignment and has not submitted it. 

This PR also includes the fix of "Learning Objectives graded" column that previously did not display anything.

Learning Objective and Submission Services


evanstraub:
I definitely think we want admin to see the same thing as instructors/students see - it's too confusing otherwise (I know I would never remember that my admin view is different from students/instructors).  It would make troubleshooting issues a nightmare.

maskib:
Everything looks as described above for me, thank you!

PR #4159, opened 2018-12-11
remove access for u-m faculty to add courses independently

chcholman:
### Status
**READY**

### Description
This PR removes Umich staff member's ability to add courses independently - we thought we'd already done this, but there was one button remaining that they could access! This was causing the crazy Canvas ID bug. 

### Migrations
NO

Prior to this, if the faculty members found their way to the courses index, there was a "New Course" button we didn't intend them to have. As a staff member who is NOT an admin, you should no longer be able to see that.

### DIFF
--- a/app/controllers/courses_controller.rb
+++ b/app/controllers/courses_controller.rb
@@ -24,7 +24,7 @@ class CoursesController < ApplicationController
 
   def index
     @courses = current_user.courses.includes(:earned_badges)
-    @can_create_courses = current_user_is_admin? || (!Rails.env.beta? && current_user_is_staff?)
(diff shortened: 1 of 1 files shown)


chcholman:
@jonathoy So, my read of the current state is that app users can't do that right now anyways - yes? Since it's checking to see if it's not-beta + staff of any kind who can see the button? But I agree that's not ideal. Will confirm with Evan, my initial inclination is that staff on app should be able to create courses, although then be limited in their next actions by whether or not they've got licenses available. Thoughts?

chcholman:
@jonathoy confirmed with Evan, good to merge and then when we're integrating licensing revisit

Issue #4167, opened 2018-12-20
Build Document Size Validation into Analytics::Event


jonathoy:
Currently we check if an underlying event document has exceeded the maximum allowable size imposed by MongoDB prior to enqueueing the Resque job - see `app/event_loggers/pageview_event_logger.rb`

It might be better if we just add a custom validator class and hook to the underlying Rails model represented by Mongoid (`lib/analytics/event.rb`)

If it has exceeded the maximum size, we could probably continue to throw the `EventLogger::Exceptions::DocumentSizeExceededError` and have resque-retry ignore it.

This might help alleviate the following exception that we are seeing in Rollbar by cutting down on the overall number of MongoDB read transactions (without upping pool size in the config):

```ruby
Timeout::Error: Timed out attempting to dequeue connection after 1 sec.
```

jonathoy:
This is contingent upon our usage of MongoDB as a data store for events

maskib:
@chcholman Did we build this validation when we removed mongo?

PR #4168, opened 2018-12-20
Disable doc size check for now

jonathoy:
### Status
**READY**

### Description
I'm not seeing any evidence of document sizes exceeding the max size yet in production, and this check effectively doubles the number of MongoDB read operations right now. I'm temporarily disabling this logic for now so that we can think about this later, either in the form of the work described in #4167 or by moving off MongoDB entirely.


jonathoy:
If there is a doc size exceeded error moving forward, I've already specified in resque-retry that we should ignore `Mongo::Error::OperationFailure`s

PR #4176, opened 2019-01-07
updated ruby version in gemfile

ebarroso7:
### Status
**READY**

### Description
Dockerfile was updated recently to reflect this change, builds were failing in quay because the gemfile was trying to use ruby 2.5.1 instead of 2.5.3

Will need to update the production Dockerfile to have the ruby version be 2.5.3 in base image repo

Also updated scrypt 3.0.6 from 3.0.5 due to compile error

### Deploy Notes
Local ruby versions may need to be updated to 2.5.3 also.
easy to use rbenv to do this version updates

### Gem dependencies
Gemfile -- ruby version changed from 2.5.1 to 2.5.3

### Migrations
NO

Gemfile

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -97,7 +97,7 @@ gem 'sanitize'
 gem 'sassc-rails'
 
 # secure crypt hashing library stronger than bcrypt or PBDBF2
-gem 'scrypt', '3.0.5'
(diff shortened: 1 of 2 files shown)


jlost:
README.md should also be updated to reflect the new Ruby version.

PR #4185, opened 2019-01-11
added a check so that locked assignments do now show in lo table

### LINKED ISSUE #4119: Inconsistent assignment visibility when assignment is locked
### Expected Behavior
AS A STUDENT: When an assignment is set to be visible when locked, but the name is not set to be visible, the display of this assignment would be consistent between the "assignments" view, and the view of assignments that meet a certain learning objective. 

### Actual Behavior
AS A STUDENT: When an assignment is set to be visible when locked, but the name is not set to be visible, the behavior is as expected on the assignments page, but I can still see the name of the assignment on the page displaying assignments that meet a certain LO, and if I click the name of the assignment, I get a "server error".

### Steps to reproduce
As an instructor, Create a new assignment, set it to meet a certain LO. Set the visibility settings to assignment visible when locked, but make sure that the other visibility settings are unchecked. Then, as a student look at the LO page for the LO you assigned this assignment to, and you should see the name of the assignment in that list. Try clicking it and see if you get a server error!

ebarroso7:
### Status
**READY**

### Description
When an assignment was locked, and not visible to the student, it would show up hidden in the assignments page, but would show up if it was linked to a learning objective. 

Now a locked assignment will not show up in a learning objective in the students view

### Todos
- [x] User Testing

### Migrations
NO

Students view of Learning objectives

======================
Closes #4119

### DIFF
--- a/app/assets/javascripts/angular/templates/learning_objectives/objectives/show.html.haml
+++ b/app/assets/javascripts/angular/templates/learning_objectives/objectives/show.html.haml
@@ -98,7 +98,8 @@
                              "data-sortable-fn"=>"loShowCtrl.sortByAssessment"}
 
       %tbody
-        %tr{"ng-repeat"=>"assignment in loShowCtrl.linkedAssignments | orderBy : loShowCtrl.sortable.predicate : loShowCtrl.sortable.reverse"}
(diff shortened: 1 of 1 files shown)


maskib:
Works as intended!

PR #4198, opened 2019-01-25
Update Google omniauth

### LINKED ISSUE #4197: Update omniauth
@Kyle Dove sent us a notification at one point that we might need to look into some changes in GC with the impending Google+ shutdown. I didn’t think it affected us but it turns out that the underlying API does pull some data from Google+. It looks like they’ve addressed it in the latest version of `omniauth-google-oauth2` so Google sign in should not be affected so long as we update before March.

jonathoy:
### Status
**READY**

### Description
- Updates `omniauth-google-oauth2` to latest version to prevent issues caused underlying Google+ API dependency
- Updates `bundler` to version 1.17.3

### Notes

Developers will want to `gem install bundler -v 1.17.3` after this is merged

======================
Closes #4197

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,7 +7,7 @@
 FROM quay.io/gradecraft/secrets:production
 MAINTAINER Nagumalli Sundeep <nagumals@umich.edu>
 ENV DOCKER_FIX random
-RUN gem install -v 1.17.1 bundler
+RUN gem install -v 1.17.3 bundler
 RUN mkdir /gradecraft
 WORKDIR /gradecraft/
 ADD Gemfile /gradecraft/Gemfile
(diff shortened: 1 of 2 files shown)


ebarroso7:
@jonathoy Will we want to update the docker file to update the bundler gem version to 1.17.3 also ?

jonathoy:
@ebarroso7 Yes to be consistent this should be updated in the Dockerfile as well as in Codeship during builds. I haven't yet made the changes

jonathoy:
@ebarroso7 I went ahead and updated Codeship builds to use 1.17.3. You will probably want to merge and deploy this soon because logging in via Google will no longer work properly in March.

ebarroso7:
@jonathoy Awesome! thank you

PR #4200, opened 2019-01-25
Students last activity date now sorting by date instead of alphabetically

ebarroso7:
### Status
**READY**

### Description
As the instructor viewing all the students in a class, the last activity column was sorting the dates alphabetically (Friday first) instead of by the actual date. 

Now the column is actually sorting students by the date they last signed in. However, students who have no last activity (haven't signed in) are sorted to the top of the list.

### Todos
- [ ] Confirm the side affect (No login = top of activity) is okay 

### Migrations
NO

Visit the students page as an instructor and click on the 'Last Activity' column


ebarroso7:
@jonathoy  I didn't think it would make a difference either... Looking through different courses with more student login data it looks like its working this way without the code (as you said). I'm wondering if it was reported this way because the day of the week is shown first ? 
This was just from Laurie so I don't think there is a github issue to check the details of this. Seems good to me without the changes.

jonathoy:
@ebarroso7 I've encountered this before and you will probably be able to replicate the bug in Laurie's course. It's usually because it's sorting on the wrong date format on the Angular side. I wouldn't doubt that it is a bug but you should confirm that you can reproduce on a larger dataset.

ebarroso7:
@jonathoy Just checked Talent Gateway again and it is sorting by date. It is still putting the students that have never logged in on top of the sort. She was having some other problems that we couldn't recreate while we were sitting next to her, I wonder if this is similar to those. 
I think this is fine to close?

jonathoy:
@ebarroso7 Feel free to if you think it's can be closed

PR #4205, opened 2019-02-05
Self graded assignments

### LINKED ISSUE #4179: Creating a Self-graded assignments gives an option of using a rubric (but they cannot be used)
### Expected Behavior
If I choose an assignment to be "student-logged", there should not be grading options that are unavailable to me displaying. In particular, there should not be a rubric available to use, since there's no functionality for students to self-log by using a rubric. For reference, here's what a self-logged assignment page would look like:

screen shot 2019-01-28 at 5 25 13 pm

### Actual Behavior
When an instructor chooses an assignment to be "student-logged," he/she can still create a rubric and it appears that they rubric would be used to assess for the self-graded assignment. The rubric cannot be used by the student for self-logging, self-logging is rather more like a "pass/fail" grading style.

### Steps to reproduce
Create an assignment that is both student-logged and contains a rubric. View that assignment as a student, and you'll see something like the screen shot above.

ebarroso7:
### Status
**READY**

### Description
This resolves a bug regarding self logged assignments and grading rubrics.

Before, if you created an assignment and marked it as student logged there was still an option to grade according to a rubric. This does not make sense because there should not be any rubrics associated with self-logged assignments. 

Now:
As a student, if you go to the description of a self-logged assignment you have the option to say you've completed the assignment (and reward full points), without having the option to use a rubric.

As a professor, when editing an assignment if you select student_logged in the 'details' of an assignment you will no longer be able to add / copy a rubric to the assignment. 

Restrictions have been applied to assignments that do not allow the instructor to make an assignment self graded and accept submissions. 

Assignments that are self-logged no longer appear in the grading status table for instructors 

### Todos
- [ ] Test in staging

### Migrations
NO

Create an assignment and make it self_logged, now there should be no associations with a rubric and this assignment.

* Assignments and rubrics 

======================
Closes #4179

### DIFF
--- a/app/models/submission.rb
+++ b/app/models/submission.rb
@@ -38,12 +38,14 @@ class Submission < ApplicationRecord
   scope :ungraded, -> do
     includes(:assignment, :group, :student)
     .where.not(id: with_grade.where(grades: { instructor_modified: true }))
+    .where.not(assignment_id: Assignment.where(student_logged: true))
   end
 
   scope :resubmitted, -> {
     includes(:grade, :assignment)
     .where("grades.student_visible = true")
     .where("grades.graded_at < submitted_at")
+    .where.not(assignment_id: Assignment.where(student_logged: true))
     .references(:grade, :assignment)
   }
 
--- a/app/models/submission.rb
+++ b/app/models/submission.rb
@@ -38,14 +38,14 @@ class Submission < ApplicationRecord
   scope :ungraded, -> do
     includes(:assignment, :group, :student)
     .where.not(id: with_grade.where(grades: { instructor_modified: true }))
-    .where.not(assignment_id: Assignment.where(student_logged: true))
+    .where.not(assignments: {student_logged: true})
   end
 
   scope :resubmitted, -> {
     includes(:grade, :assignment)
     .where("grades.student_visible = true")
     .where("grades.graded_at < submitted_at")
-    .where.not(assignment_id: Assignment.where(student_logged: true))
+    .where.not(assignments: {student_logged: true})
     .references(:grade, :assignment)
   }
 
--- a/app/assets/javascripts/angular/directives/grades/edit.coffee
+++ b/app/assets/javascripts/angular/directives/grades/edit.coffee
@@ -28,7 +28,7 @@
         assignment = AssignmentService.assignment()
         return "" if !assignment
 
-        if assignment.is_rubric_graded == true
+        if assignment.is_rubric_graded == true && !assignment.student_logged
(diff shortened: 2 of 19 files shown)


ebarroso7:
@maskib  Could you give this some user testing in staging when you have a chance? Let me know if you need any more guidance

maskib:
After creating a rubric with criterion levels, if you:
1. Change the assignment to be self-logged
2. Change it back so that it's not self-logged

An extra copy of each criterion, and each level within each criterion, is created. E.G.

maskib:
screen shot 2019-02-12 at 10 37 49 pm

ebarroso7:
Hey, @maskib could you give this another check?

maskib:
Thanks Erik! This might be out of scope for this PR, but I wanted to let you know that it's also possible to mark an assignment as self-logged that accepts resubmissions. But it's not possible (at least, it doesn't look that way to me) for students to self-log a second time, which they would have to do once they resubmit, because once you self-log, the "i have completed this assignment" button disappears.

maskib:
I'll also add, though I'm not sure it's in the scope of this PR, that it's possible for assignments to be self-logged and allow resubmissions, but this is a bit of a conflict, because once you click the button that self-logs your grade, it disappears, and you won't be able to self-log again once you've resubmitted. Also, assignments that allow for resubmissions still show up in the instructor "grading status" page, as though they need grading, at which point the instructor can overwrite the score a student gave themself. Happy to talk through this if what I'm writing doesn't  make much sense!

app/models/submission.rb
```diff
@@ -38,12 +38,14 @@ class Submission < ApplicationRecord
   scope :ungraded, -> do
     includes(:assignment, :group, :student)
     .where.not(id: with_grade.where(grades: { instructor_modified: true }))
+    .where.not(assignment_id: Assignment.where(student_logged: true))
```
jonathoy: Hey @ebarroso7, I wanted to point out something that I noticed here and in the change below. This is something that is really subtle but can potentially result in a non-performant query. You are using `Assignment` with a capital A here instead of `assignment` with lowercase a, and this actually makes a big difference in how the ORM will generate the underlying SQL query.

You are essentially asking here for submissions where not assignment_id in (select * from assignments where student_logged = true). This means that an extra query will be made just to fetch **all** assignments that are student logged, rather than filtering on only the joined assignments.

maskib:
@ebarroso7 This looks good to me with the exception of the bug I pointed out, thanks! I will create a separate github issue for that bug.

PR #4232, opened 2019-04-09
Add Maintainer role to LTI connection

### LINKED ISSUE #4076: Add "Maintainer" as an LTI role that will be transformed to "Instructor" in GC
Sakai offers users the ability to create project sites, where the roles available are "Maintainers" and "Participants." We should interpret "Maintainer" as "Instructor" in GradeCraft.

ebarroso7:
### Status
**READY**

### Description
Some pathways into Gradecraft use an LTI connection where the role of the user is determined based on their role from the LTI. Currently we really only support Instructors who have a role of "Instructor" from their LTI. 

We've found that Sakai calls their role "Instructor" a "Maintainer", so when an instructor is logging into Gradecraft using an LTI connection provided by Sakai they are translated to have a role of student in Gradecraft when they are really an instructor.

This fix aims to add a check to the LTI consumer to allow "Maintainer" to have the role of "instructor" when added to Gradecraft. 

### Todos
- [x] Test in staging to make sure LTI connection isn't broken
- [x] Use https://ltiapps.net/test/tc.php  to see what LTI response is back
- [ ] Push to production, confirm LTI connection isn't broken & see if professor can login using LTI / Sakai 
- [x] Confirm the role on the Sakai side is "Maintainer" not "Maintain" 

### Documentation
Documentation added as a result of the testing work done to using the LTI consumer emulator: 
[Here](https://github.com/UM-USElab/Developer-Documentation/wiki/Testing-LTI-connections-on-staging)

### Deploy Notes
When this is tested in staging we need to make sure the previous LTI connections are not broken (students and instructors). Then we can add it to production and see if it fixes the issue.

LTI connection 

======================
Closes #4076


ebarroso7:
https://sakai.screenstepslive.com/s/sakai_help/m/13946/l/604894-what-are-permissions-and-roles 

Looking at this source and going to the "Default participant roles" section, the role in Sakai seems to be called "Maintain"

Issue #4121, opened 2018-10-09
Assignment unlock conditions do not copy when copying an assignment


maskib:
### Expected Behavior

Assignment unlock conditions should copy when copying an assignment.

### Actual Behavior

Assignment unlock conditions do not copy when copying an assignment.

### Steps to reproduce
Create an assignment, set any unlock condition you want. Copy that assignment, and check to see if the unlock conditions are there in the copied assignment.

chcholman:
➤ Erik Barroso commented:

@michaelskib  could you test this guy too? 
https://github.com/UM-USElab/gradecraft-development/pull/4214

chcholman:
➤ Michael Skib commented:

Unlock conditions appear to be copying correctly now!

PR #4234, opened 2019-04-12
Refactor copy uc

meekray:
### Status
Ready

### Description
Currently, the logic for copying unlock conditions of an assignment is in the Assignments Controller. These commits include a refactor where this logic will now take place in a function in the Assignments model instead, thus improving code reusability and readability. 

### Related PRs
This refactors code that was originally written as part of  #4121 

Assignments Model
Assignments Controller


ebarroso7:
Hey @meekray, I made those changes I requested above but still want you to review it and give it some more testing. 

I moved the method higher in the assignment model to put it underneath `copy_with_prepended_name` because thats the method used in the assignment_controller just before your `copy_unlock_condition` call. This is to help with tracing / readability.

I also added a new method to the assignment model to check if an assignment has an unlock condition.

PR #4236, opened 2019-04-17
Unlock visibility

### LINKED ISSUE #4171: Unlock visibility conflicts
### Expected Behavior
As an instructor, I expect that if an assignment is set to be invisible, then it will not appear in the list of unlock requirements/keys. 

### Actual Behavior
However, this setting is being overridden by the visibility controls established in the unlock conditions. So if an invisible assignment is set to show in the unlock conditions, it will show.

### Steps to reproduce
Create an assignment that is set to invisible. Make it a key for another assignment. Go look at the requirements for that assignment - the 'hidden' assignment will appear.

meekray:
### Status
**Ready**

### Description
It is expect that if an assignment is set to be invisible, then it will not appear in the list of unlock requirements/keys. This pull request fixes this, as a link-less statement saying "You have not unlocked this assignment" will appear in place of the link.

### Migrations
NO

Create an assignment that is set to invisible. Make it a key for another assignment. Go look at the requirements for that assignment - the 'hidden' assignment will no longer appear, and is instead replaced by the link-less statement. Furthermore, hovering over the lock icon in the assignments list will also display this message.

Assignments
- Description view
- Icon view

Closes #4171

### DIFF
--- a/app/models/unlock_condition.rb
+++ b/app/models/unlock_condition.rb
@@ -45,9 +45,9 @@ class UnlockCondition < ApplicationRecord
         description = "#{ condition_state_do } in the #{ condition.name } #{unlockable.course.assignment_term} Type"
       end
     elsif condition_type == "LearningObjective"
-        description = "#{ condition_state_do } the #{ condition.name } #{unlockable.course.learning_objective_term.singularize}"
+      description = "#{ condition_state_do } the #{ condition.name } #{unlockable.course.learning_objective_term.singularize}"
     else
-        description = "#{ condition_state_do } the #{ condition.name } #{ condition_type }" 
+      description = "#{ condition_state_do } the #{ condition.name } #{ condition_type }" 
     end
     description += condition_date_sentence(condition_date_timezone) if condition_date.present?
     description
--- a/app/models/unlock_condition.rb
+++ b/app/models/unlock_condition.rb
@@ -29,6 +29,11 @@ class UnlockCondition < ApplicationRecord
     check_condition_for_each_student(group)
   end
 
+  def check_assignment_condition_visible
+    assignment = Assignment.find(self.condition_id)
+    return assignment.visible
+  end
+
   # Human readable sentence to describe what students need to do to unlock this
   def requirements_description_sentence(condition_date_timezone=nil)
     if condition_type == "Course"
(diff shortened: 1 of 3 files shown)


app/models/unlock_condition.rb
```diff
@@ -40,9 +45,9 @@ def requirements_description_sentence(condition_date_timezone=nil)
         description = "#{ condition_state_do } in the #{ condition.name } #{unlockable.course.assignment_term} Type"
       end
     elsif condition_type == "LearningObjective"
-      description = "#{ condition_state_do } the #{ condition.name } #{unlockable.course.learning_objective_term.singularize}"
+        description = "#{ condition_state_do } the #{ condition.name } #{unlockable.course.learning_objective_term.singularize}"
     else
-      description = "#{ condition_state_do } the #{ condition.name } #{ condition_type }"
+        description = "#{ condition_state_do } the #{ condition.name } #{ condition_type }"
```
ebarroso7: Could you remove the extra tab in the two description lines @meekray

app/views/assignments/_description.haml
```diff
@@ -73,8 +73,11 @@
         - presenter.assignment.unlock_conditions.each do |uc|
           - if !uc.condition_type == "Learning Objective"
             %li= link_to uc.requirements_description_sentence(current_user.time_zone), uc.condition
-          - else
-            %li= link_to uc.requirements_description_sentence(current_user.time_zone), learning_objectives_objective_path(uc.condition_id)
+          - elsif uc.condition_type == "Assignment"
+            - if !uc.check_assignment_condition_visible
+              %li= "You have not unlocked this assignment."
+            - else
```
ebarroso7: It looks like the `- else` and accompanying `%li` lines are indented one tab too far. 

Now it is nested under the `elseif` line on 76 so it would actually only appear if the uc.conditon_type is Assignment. We want that `- else` on 79 to display regardless of `uc.condition_type` but also if it the unlock condition assignment is visible.

I think the best way to fix this and not duplicate code would be to change the `-elseif` statement on 76 to include the `if` condition on 77. Something like: 
`- else if uc.condition_type == "Assignment" && !uc.check_assignment_condition_visible `

(^ Didn't test this line or anything so not sure if its correct )

app/views/assignments/index_student/components/_assignment_icons.haml
```diff
@@ -10,8 +10,11 @@
         .small.italic In order to unlock it you must:
         %ol
           - assignment.unlock_conditions.each do |condition|
-            %li= condition.requirements_description_sentence(current_user.time_zone)
-
+            -if condition.condition_type == "Assignment"
```
ebarroso7: Looks like these conditionals will need the same changes as comment on above

ebarroso7:
Hey @adishy, I made the requested changes to this PR do you think you could pull it down and do some testing for me? 
Let me know if you need help with that

adishy:
Will do!

PR #4262, opened 2019-05-20
Admin user not listed in /staff

### LINKED ISSUE #4259: It is possible to add an admin staff member to an individual course, but then that admin does not appear in the list of course staff
It is possible to add an admin staff member to an individual course, but then that admin does not appear in the list of course staff.

Steps to reproduce:

1. Go to "https://staging.gradecraft.com/staff"
1. Click "+ new staff member"
1. Add name, email, and select "admin" from the dropdown list of roles.
1. Click "create user"
1. Observe whether the new user has been added to the list of course staff.

┆Attachments: <a href="https://trello-attachments.s3.amazonaws.com/5a6f2020d4ad818ef1fffe12/5cdf0cd379ce3580ee0316dc/7489713fd5baf4089e3c7c62f9df2721/Screen_Shot_2019-05-17_at_3.05.33_PM.png">Screen Shot 2019-05-17 at 3.05.33 PM.png</a>

adishy:
### Status
**READY**

### Description
An admin user can be added to a course, but would not appear in the staff list for that course. The issue was probably the "admin" role not being listed in models/user.rb > with_role_in_courses

### Migrations
NO

1. Create a new staff member by visiting /staff and clicking the new staff member button, and enter details for the new user and select "Admin" from the dropdown menu
2. The new Admin staff member will not be listed under staff for the course, however, they will be listed in /users as an admin

* /staff 
* Viewing staff for a particular course

======================
Closes #4259


ebarroso7:
@adishy, Tracing through this PR makes me want to think a little bit more about what we are adding here.

I'm not sure how dated the comment is - if it still applies to the codebase, but in the `models/course.rb` file there is a comment above the `staff` method that says it does not include 'admins'.  Seeing this makes me want to know if adding 'admin' into this method will throw things off in other places in the application. I'm not sure where it would come into play but I could see a page that instructors are able to go to and see their staff members for the course which would then include "admin". There are a lot more admins on the production sites (app/umich) than our local environment. 

The comment in the `models/course.rb` file makes a comparison to the `is_staff?` method. However, this is in the `models/user.rb` file so it would not apply to the task of filtering through a course to find its staff members.

A decision we'll want to make is if we actually want admins to be displayed on this page and if we do, will instructors be able to see them also? Like I said there are about 20 admins in each course on app and umich that would be less informative to instructors.

ebarroso7:
We're not moving forward with this feature change at this time

PR #4283, opened 2019-05-29
Student Dashboard progress bar customization in course settings

### LINKED ISSUE #4306: A setting should exist in the "Advanced" tab in the course settings page for instructors to manually set the upper limit value for the progress bar in the student dashboard

adishy:
### Status
**READY**

### Description
* The upper limit of the progress bar displayed on the student dashboard can be too large, causing a student's progress to appear very small in comparison to the course total while it might be the case that many points in the course are optional. Now a new course setting has been added which can be used to set the student dashboard progress bar's upper limit to a custom value, and can be set back to 0 for the default upper limit.

### Migrations
NO

1. As an instructor go to the "Advanced Settings" tab in the course settings page (i.e. /courses/<id>/edit). Under the "Student Settings" list, you can now enter a custom value for the upper limit and save the course settings.
2. As a student, log in and select the course, you will now see that the progress bar's upper limit has changed to the value that was set in the course settings page.

* Student dashboard (directives/analytics/points_breakdown.coffee)
* Students controller (controllers/api/students_controller.rb)
* Course model (for adding validation for the full_points field used to store the custom value for the progress bar's upper limit) (models/course.rb)

======================
Closes #4306

### DIFF
--- a/app/controllers/api/students_controller.rb
+++ b/app/controllers/api/students_controller.rb
@@ -35,6 +35,5 @@ class API::StudentsController < ApplicationController
     else 
       @course_potential_points_for_student = current_course.full_points
     end
-
   end
 end
--- a/app/controllers/api/students_controller.rb
+++ b/app/controllers/api/students_controller.rb
@@ -29,6 +29,12 @@ class API::StudentsController < ApplicationController
     @student = student
     @assignment_types = current_course.assignment_types
     @earned_badge_points = @student.earned_badge_score_for_course current_course
-    @course_potential_points_for_student = current_course.total_points + @earned_badge_points
+  
+    if current_course.full_points.nil? || current_course.full_points == 0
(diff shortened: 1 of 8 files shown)


maskib:
@adishy @ebarroso7  This works great! I want to point out one thing however. The number furthest to the right on the progress bar gets cut off in certain cases. In particular, when you set it to a value such as 1000, 2000, 3000, 4000. Basically whenever the number furthest to the right is at the very edge of the graph gets cut off and this seems to happen with those sorts of rounded numbers (also with 300, 400, 500, etc). Not sure if this was happening before the change but I wouldn't be surprised. This is me being as nitpicky as possible :) if we think this is acceptable, then that's fine! 

Screen Shot 2019-06-07 at 12 37 55 PM

ebarroso7:
Looks great! :D thanks

PR #4293, opened 2019-06-04
Prevent internal server error when viewing analytics for a course with no assignment types

### LINKED ISSUE #4289: Going to the "analytics" page in an unpublished course leads to a server error
Screen Shot 2019-06-03 at 2 49 46 PM

### Expected Behavior
Either there wouldn't be an analytics link for an unpublished course, or it would go to a blank page, with some sort of notification stating that content will display once the course starts.

### Actual Behavior
Going to the "analytics" page in an unpublished course leads to a server error.

### Steps to reproduce
1. In a course that hasn't been published, click "analytics" in the left nav.

adishy:
### Status
**READY**

### Description
* When viewing the analytics page for a course with no assignment types an internal server error was caused due to an uninitialized object access in models/breadcrumb_trail.rb > info_per_assign
* Now a conditional check has been added to prevent the server error and a message is displayed if the course has no assignment types or if the course has no assignments

### Migrations
NO

1. As an admin create a new course
2. Click on the "Analytics" link on the navigation sidebar. There will be no internal server error when the /per_assign page is displayed and there will be a message that indicates that there are no assignment types in the course

* models/breadcrumb_trail.rb
* Viewing assignment analytics (/per_assign) (i.e. views/info/per_assign.html.haml)

======================
Closes #4289

### DIFF
--- a/app/models/breadcrumb_trail.rb
+++ b/app/models/breadcrumb_trail.rb
@@ -341,9 +341,7 @@ class BreadcrumbTrail < Croutons::BreadcrumbTrail
 
   def info_per_assign
     dashboard
-    if objects[:course].assignment_types.length > 0
-      breadcrumb(objects[:assignment_types].first.course.assignment_term + " Analytics")
-    end
+    breadcrumb(objects[:course].assignment_term + " Analytics")
   end
 
   def institutions_edit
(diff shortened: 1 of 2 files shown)


app/models/breadcrumb_trail.rb
```diff
@@ -341,7 +341,9 @@ def info_grading_status
 
   def info_per_assign
     dashboard
-    breadcrumb(objects[:assignment_types].first.course.assignment_term + " Analytics")
+    if objects[:course].assignment_types.length > 0
```
ebarroso7: lets clean this up

ebarroso7: cleaned!

PR #4304, opened 2019-06-05
It should be possible to select a team for a new user during their creation

### LINKED ISSUE #4308: Removing a staff member who is a team leader from a course does not also remove them from their team
### Expected Behavior
* When removing a staff member from a course, if they are a leader for a team in the course, they should also be removed from the team.

### Actual Behavior
* Currently, removing a staff member from a course who is a leader for a team does not remove them from their team.

### Steps to reproduce
* As an admin, create a staff member and assign them as the leader for a team
* Delete the staff member
* When viewing all the teams in the course (i.e. /teams), the staff member who was removed is still listed under the "TA" column with their team.

adishy:
### Status
**READY**

### Description
* When creating a new user as an admin, it should be possible to add the new user to a team if the course has teams.
* Now a dropdown has been added to the /users/new page under the course section that lists all the teams in the course if the course has teams. If the new user is a student, they are added to the list of the team's students. If the user is a GSI or professor they are added to the list of the team's leaders.
* Also, now deleting a staff member will also remove them from their team

### Migrations
NO

1. As an admin, create a new user by visiting /users/new in a course with teams
2. Select a team for the user by selecting a team from the dropdown. When the user is created they are a member of the team. If no options from the dropdown are selected, the new user is not added to any team.

* Adding a new user (/users/new)
* controllers/users_controller.rb
* The course view when creating a new user (/views/users/_courses.html.haml)
* The model course_membership.rb to add an attr_accessor called team_in_course to store the new user's team in the course
* courses_helper.rb now includes a helper function to list all the names of the teams

======================
Closes #4308, #4317

### DIFF
--- a/app/controllers/course_memberships_controller.rb
+++ b/app/controllers/course_memberships_controller.rb
@@ -56,10 +56,31 @@ class CourseMembershipsController < ApplicationController
     end
   end
 
+  def remove_leaders_from_teams(course_membership)
+    if !course_membership.user.is_staff?(course_membership.course)
+      return false
+    end
+
+    staff_member = course_membership.user
+
+    course = course_membership.course
+
+    staff_member.team_leaderships_for_course(course).each do |team_leadership|
+      team = team_leadership.team 
+      team.leaders = team.leaders.reject {|member| member["id"] == staff_member.id}
+      team.save
+    end
+
+    return true
+  end
+
   def destroy
     course_membership = current_course.course_memberships.find(params[:id])
-    Services::CancelsCourseMembership.call course_membership
+    
+    remove_leaders_from_teams(course_membership)
 
+    Services::CancelsCourseMembership.call course_membership
+   
     respond_to do |format|
       format.html { redirect_to session[:return_to], notice: "#{course_membership.user.name} was successfully removed from course." }
       format.json { head :ok }
--- a/app/controllers/course_memberships_controller.rb
+++ b/app/controllers/course_memberships_controller.rb
@@ -57,10 +57,6 @@ class CourseMembershipsController < ApplicationController
   end
 
   def remove_leaders_from_teams(course_membership)
-    if !course_membership.user.is_staff?(course_membership.course)
-      return false
-    end
-
     staff_member = course_membership.user
 
     course = course_membership.course
@@ -70,14 +66,14 @@ class CourseMembershipsController < ApplicationController
       team.leaders = team.leaders.reject {|member| member["id"] == staff_member.id}
       team.save
     end
-
-    return true
   end
 
   def destroy
     course_membership = current_course.course_memberships.find(params[:id])
     
-    remove_leaders_from_teams(course_membership)
+    if course_membership.user.is_staff?(course_membership.course)
+      remove_leaders_from_teams(course_membership)
+    end
 
     Services::CancelsCourseMembership.call course_membership
    
--- a/app/helpers/courses_helper.rb
+++ b/app/helpers/courses_helper.rb
@@ -19,8 +19,7 @@ module CoursesHelper
   end
 
   def available_teams(course)
-    teams = course.teams.pluck("name")
-    return teams
+    course.teams.pluck("name")
   end
 
   def bust_course_list_cache(user)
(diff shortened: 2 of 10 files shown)


maskib:
This works well! I would make two changes:
-  The header above the dropdown menu used to select the section to add users to says "select team" but it should say "select ("term for "section")"
- This dropdown menu also now exists on the "edit user" pages. I think that's fine, but if you try to change which section users are in from that page and click "update user", it doesn't actually change the section they're in. The "New Staff Member (x) was successfully updated!" still shows up after that too so people will think that they have in fact changed that user's section.

I'm also wondering if it's worth changing the term "ta" on the "sections" page to something more like "section leader" to be more general? Not sure if that matters, that's just a thought. Thank you!!

maskib:
@adishy I'll also just add that I recognize that in many places the terms "sections" and "teams" are used interchangeably, and that we might want to standardize this across the board in the future. What users typically see is "section" or whatever term they're using for "section", so I think we should stick with that for users. Thanks again!

adishy:
@maskib Great catch on both the term_for point as well as about the edit user page! I've updated the title to be the singular version of whatever the term for teams are in the course and changed the page so that the "Select Team" dropdown only appears while creating a user (i.e. /users/new) and not when editing an existing user.

adishy:
@maksib The requested changes are now on staging!

maskib:
@adishy @ebarroso7 Great, thanks! I want to ask: what happens if somebody adds a new staff member to the wrong section in the course or if the staff member needs to change sections midway through the course? Would we have to remove them from the course and re-add them? Possibly not an  issue, I just want to confirm whether this is the case, and if we're okay with that.

adishy:
@maskib Currently the interaction for changing the section of an existing staff member is as follows:

1. Visit the /teams page as admin and choose the team the staff member is currently in and edit the team. Uncheck the staff member's name from the "Leaders" section.
2. Go back to the /teams page and choose the new team for the staff member and edit the team. Check the staff member's name in the "Leaders" section for the new team. Note that it is possible add one staff member as the leader for multiple teams.

I do agree that this interaction is a bit tedious :/

maskib:
@adishy Got it, thanks! It's a little tedious, but I can't imagine it would have to happen often. Thanks!

app/controllers/course_memberships_controller.rb
```diff
@@ -56,10 +56,31 @@ def reactivate
...
+    end
+
+    return true
+  end
+
   def destroy
     course_membership = current_course.course_memberships.find(params[:id])
-    Services::CancelsCourseMembership.call course_membership
+    
+    remove_leaders_from_teams(course_membership)
```
ebarroso7: I think it would make the logic a little easier to read in this area if you pulled out the conditional check that the user is a staff member from within the `remove_leaders_from_team` to within the destroy function.

so within the `destroy` method it would be: 

if course_membership.user.is_staff?(course_membership.course)
      remove_leaders_from_teams(course_membership)
end

app/controllers/course_memberships_controller.rb
```diff
@@ -56,10 +56,31 @@ def reactivate
...
+
+    course = course_membership.course
+
+    staff_member.team_leaderships_for_course(course).each do |team_leadership|
+      team = team_leadership.team 
+      team.leaders = team.leaders.reject {|member| member["id"] == staff_member.id}
+      team.save
+    end
+
+    return true
```
ebarroso7: Not sure if you need this `return true` here ?

app/controllers/course_memberships_controller.rb
adishy: I believe that particular change was made to keep codeclimate happy, I do agree it was a bit unreadable so it has been updated.

app/controllers/course_memberships_controller.rb
adishy: Updated the function to not include this.

app/helpers/courses_helper.rb
```diff
@@ -18,6 +18,11 @@ def available_roles(course)
     roles
   end
 
+  def available_teams(course)
+    teams = course.teams.pluck("name")
+    return teams
+  end
```
ebarroso7: with ruby's implicit returns, I think you can shorten this to just `course.teams.pluck("name")`  
not setting or returning `teams`

PR #4310, opened 2019-06-11
Export assignment submissions in a specified date range in the course data exports page

### LINKED ISSUE #4315: It should be possible to export a specified date range for certain course data export types
### Expected Behavior
* For some courses such as the Talent Gateway Online Community course, the ability to export assignment submissions within a specified date range is required.

### Actual Behavior
* Currently, this is not possible and exporting assignment submissions will export all assignment submissions from the creation of the course, which is not required.

### Steps to reproduce
* Visit the /downloads page as an instructor. It is not possible to specify a date range for exporting the assignment submissions course export type.

adishy:
### Status
**IN DEVELOPMENT**

### Description
* When exporting assignment submissions for some courses, the data to export is too large and thus it is useful to specify a date range which can be used as a filter to specify which assignment submissions to export. 
* Now, a form has been added in the /downloads page for the "Assignment Submissions" download link to download only a specified date range if desired (or all the data if no dates are specified)

### Migrations
NO

1. As an instructor, visit the /downloads page, and for the "Assignment Submissions" link, specify a date range and a column to apply the date range to
2. Click the link and open the attachment in the email. The data exported should be within the date range for the specified column

* Course data exports (i.e. /downloads)
* Added a script for the form for the "Assignment Submissions" link (assets/javascripts/downloads.js)
* SubmissionExporter now uses a date range
======================
Closes #4315


maskib:
@adishy This works great so far! One thing to change potentially is that if one of the dates is blank or incomplete, it will cause a 500 server error. Another idea is that it might be better if the date fields retain the values you just used after you click the button to trigger the data export. That way you can see what you just did, in case you're not 100% positive you exported the right date range. Does that make sense?

adishy:
@maksib The requested changes have now been made on staging!

maskib:
@adishy @ebarroso7 Export does not appear to send if one of the date fields for the date range is left blank.

PR #4319, opened 2019-06-18
Badge description should not disappear when attaching badge icon

### LINKED ISSUE #4203: Attaching an image to badge prior to save deletes description text
### Expected Behavior
When creating a badge, a user should be able to write a description and continue to edit the creation form and retain description information while interacting with other pieces of the form

### Actual Behavior
The instructor can write a description for the badge and when he/she clicks the "Attach Icon" the description for the badge disappears.  This happens even after the description text has been marked as "autosaved."  This only occurs on the initial badge creation.  Editing an existing badge by taking out the text and changing the icon picture does not replicate the issue.

### Steps to reproduce

adishy:
### Status
**READY**

### Description
* When attaching an icon for a badge as an instructor, the description text disappeared as the changes to the description were not saved, and saving an icon reset the description input value. 
* Now attaching a badge icon does not make the description text disappear

### Migrations
NO

1. As an instructor create a badge and provide a description for the badge
2. Attach an icon for the badge. The description input retains the value provided instead of resetting its value to become blank.

* Badge edit template (i.e. templates/badges/edit_basics.html.haml)

======================
Closes #4203

### DIFF
--- a/app/controllers/api/badges_controller.rb
+++ b/app/controllers/api/badges_controller.rb
@@ -31,14 +31,15 @@ class API::BadgesController < ApplicationController
   end
 
   def show
+    @badge = Badge.find(params[:id])
   end
 
   # PUT api/badges/:id
   def update
     @badge = Badge.find(params[:id])
-
     if @badge.update_attributes badge_params
-      render "api/badges/show", :locals => { :badge => Badge.find(params[:id]) }, success: true, status: 200
+      @badge = Badge.find(params[:id])
+      render "api/badges/show", success: true, status: 200
     else
       render json: {
         message: "failed to save badge",
@@ -51,7 +52,7 @@ class API::BadgesController < ApplicationController
   def create
     @badge = current_course.badges.new(badge_params)
     if @badge.save
-      render "api/badges/show",:locals => { :badge => @badge }, success: true, status: 201
+      render "api/badges/show", success: true, status: 201
     else
       render json: {
         message: "failed to save badge",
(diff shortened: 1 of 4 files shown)


maskib:
@adishy @ebarroso7  I'm not sure if this is related to the code you just pushed or if it was a preexisting issue, but after you add a badge icon image, if you change the "points awarded" value, the badge icon image disappears.

adishy:
@maksib That sounds like it's probably related, will take a look!

maskib:
@adishy to be clear, it just disappears on that particular page, the 'edit badge' page. if you go to the 'badges' page, the image still appears there.

adishy:
@maskib The updated version is now on staging!

app/controllers/api/badges_controller.rb
```diff
@@ -38,6 +38,7 @@ def show
   def update
     @badge = Badge.find(params[:id])
     if @badge.update_attributes badge_params
+      @badge = Badge.find(params[:id])
```
ebarroso7: @adishy I'm wondering why the badge instance variable is set again? I tried to test it without this locally and I couldn't tell if it was working or not because my badge icon wouldn't upload.

adishy: There was a weird issue with the scope of the badge instance variable where the JSON view would not get the updated badge instance variable after calling update_attributes and caused the badge to have the default badge icon url. Finding the badge again seems to solve this problem and the view gets the updated version of the badge instance variable.

ebarroso7: works for me! thanks

Issue #3978, opened 2018-05-09
Add column describing visible state to Assignment Structure export


chcholman:
### Expected Behavior
As an instructor, I would like to know if an assignment is visible to students by default via the assignment structure export file. Answers here could be: Yes, No, and If Unlocked

chcholman:
➤ adishy commented:

https://github.com/UM-USElab/gradecraft-development/pull/4320

chcholman:
➤ adishy commented:

https://github.com/UM-USElab/gradecraft-development/pull/4320

PR #4320, opened 2019-06-19
Course data exports additional features

### LINKED ISSUE #3854: Remove deactivated students from Final grade download

adishy:
### Status
**READY**

### Description
* Adds a column in the "Assignment Structure" export that describes the student visibility of the assignment
* In the "Final Grades" export, exports only activated students

### Migrations
NO

1. Visit the Course Data Exports page (i.e. /downloads) as an instructor and download the export "Final Grades". Only activated students will be exported.
2. Visit the Course Data Exports page (i.e. /downloads) as an instructor and download the export "Assignment Structure". There will be an additional column that describes the student visibility of the assignment.

* Final Grades exporter (exporters/course_grade_exporter.rb)
* Assignment Structure exporter (exporters/assignment_exporter.rb)

======================
Closes #3854, #3978

### DIFF
--- a/app/exporters/assignment_exporter.rb
+++ b/app/exporters/assignment_exporter.rb
@@ -54,14 +54,14 @@ class AssignmentExporter
   def visibility(assignment)
     if assignment.unlock_conditions.exists?
       if !assignment.visible_when_locked
-        return "If Unlocked"
+        return "Hidden when locked"
       end
       
-      return "Yes (even if assignment locked)"
+      return "Visible when locked"
     end
 
-    return "Yes" if assignment.visible
+    return "Visible" if assignment.visible
       
-    return "No"
+    return "Hidden"
   end
 end
--- a/app/exporters/assignment_exporter.rb
+++ b/app/exporters/assignment_exporter.rb
@@ -52,7 +52,7 @@ class AssignmentExporter
   end
 
   def visibility(assignment)
-    if assignment.unlock_conditions.length > 0
+    if assignment.unlock_conditions.exists?
       if !assignment.visible_when_locked
         return "If Unlocked"
       end
--- a/app/exporters/assignment_exporter.rb
+++ b/app/exporters/assignment_exporter.rb
@@ -53,11 +53,11 @@ class AssignmentExporter
 
   def visibility(assignment)
     if assignment.unlock_conditions.length > 0
-      if assignment.visible_when_locked
-        return "Yes (even if assignment locked)"
-      else
+      if !assignment.visible_when_locked
         return "If Unlocked"
       end
+      
+      return "Yes (even if assignment locked)"
     end
 
     return "Yes" if assignment.visible
--- a/app/exporters/assignment_exporter.rb
+++ b/app/exporters/assignment_exporter.rb
@@ -15,7 +15,7 @@ class AssignmentExporter
 
   # These headers are not used for import
   ADDITIONAL_HEADERS = [
-    "Assignment Id", "Created At", "Submissions Count", "Grades Count", "Learning Objectives"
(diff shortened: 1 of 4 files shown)


app/exporters/assignment_exporter.rb
```diff
@@ -49,4 +50,18 @@ def formatted_date(date)
     return nil if date.nil?
     date.in_time_zone(@user.time_zone).strftime("%m/%d/%Y")
   end
+
+  def visibility(assignment)
+    if assignment.unlock_conditions.length > 0
```
ebarroso7: I would prefer that this line uses `assignment.unlock_conditions.exists?` rather than the `.length > 0`

adishy: Just pushed the change :)

app/exporters/course_grade_exporter.rb
```diff
@@ -4,7 +4,7 @@ class CourseGradeExporter
   def final_grades_for_course(course)
     CSV.generate do |csv|
       csv.add_row baseline_headers
-      course.students.order_by_name.each do |student|
+      course.students.where(activation_state: "active").order_by_name.each do |student|
```
ebarroso7: making a note here to investigate activation_state some more... looking on prod there are three: "active", "pending", and nil. 
I'm also not sure where this is used and if it addresses the issue in #3854.

adishy: Updated to use the course membership of the student to check whether they have been deactivated in the course.

app/exporters/course_grade_exporter.rb
```diff
@@ -4,7 +4,7 @@ class CourseGradeExporter
   def final_grades_for_course(course)
     CSV.generate do |csv|
       csv.add_row baseline_headers
-      course.students.order_by_name.each do |student|
+      course.students.order_by_name.reject{ |student| !student.course_memberships.find_by(course_id: course.id).active }.each do |student|
```
ebarroso7: @adishy good fix, but i still think we can do better.

I'm thinking we make this its own method, or scope -- something like `active_students`

so you can make the line as clean as `course.active_students.order_by_name.each do |student|`

adishy: Good idea! Will change it to use this :)

adishy: An active_students scope has been added in models/user.rb to get active students in a course.

app/exporters/assignment_exporter.rb
```diff
@@ -49,4 +50,18 @@ def formatted_date(date)
...
     date.in_time_zone(@user.time_zone).strftime("%m/%d/%Y")
   end
+
+  def visibility(assignment)
+    if assignment.unlock_conditions.exists?
+      if !assignment.visible_when_locked
+        return "If Unlocked"
+      end
+      
+      return "Yes (even if assignment locked)"
```
ebarroso7: probably more clear to say "Yes, even when locked"

app/exporters/assignment_exporter.rb
```diff
@@ -49,4 +50,18 @@ def formatted_date(date)
     return nil if date.nil?
     date.in_time_zone(@user.time_zone).strftime("%m/%d/%Y")
   end
+
```
ebarroso7: I'm thinking the column options should be like:
"Visible when unlocked"
"Visible when locked"
"Visible"
"Hidden"

ebarroso7: changed my mind!! Good night sleep always does it...
"Hidden when locked"

adishy: This has been updated as specified :)

PR #4108, opened 2018-09-19
Add Assignment-Level Overview of Learning Objectives for Instructors

jonathoy:
### Status
**READY**

### Description
Adds an additional tab when viewing an assignment that shows a high-level overview of learning objective assessments.

screen shot 2018-09-19 at 11 18 08 am

### Impacted Areas in Application
Learning objectives

======================
Closes #[ISSUE NUMBER]

PR #4329, opened 2019-07-02
Learning objectives should have at least two levels

### LINKED ISSUE #4114: Require that a learning objective have at least two levels
_Originally posted by @maskib in https://github.com/UM-USElab/gradecraft-development/pull/4108#issuecomment-426011639_

adishy:
### Status
**READY**

### Dev note:
up     20190702175922  Add default level to learning objective level
switching between branches turn down migration with:

`rake db:migrate:down VERSION=20190702175922`

### Description
* A learning objective should have at least two learning objective levels. Currently it is possible to have a learning objective with no learning objective levels or one learning objective level.
* Now, creating a learning objective creates two learning objectives that cannot be deleted, one with the minimum proficiency level and one with the maximum proficiency level. Additional learning objective levels can be created and deleted as necessary however every learning objective will have at least two learning objective levels.

### Related PRs
`add-lo-assignment-overview` [#4108]

### Deploy Notes
* A new boolean value called default_level has been added to the LearningObjectiveLevel model for checking whether a learning objective level was created when the learning objective was created or whether the level as created by the user.

### Migrations
YES

1. As an instructor, create a learning objective.
2. The learning objective will have two levels one with the minimum proficiency level set and one with the maximum proficiency level set. It should not be possible to delete these two levels, but it is possible to add and delete more learning objective levels if required by the instructor.

* Creating learning objective (i.e. models/learning_objective.rb,  controllers/api/learning_objectives/objectives_controller.rb)
* Learning objective edit page

======================
Closes #4114


maskib:
@adishy Should it be possible to delete the two initially created levels if you create more levels?

adishy:
@maskib 
That's an interesting question! The issue does not explicitly describe any particular desired behavior regarding this.
* The primary problem this PR tries to address is that it is possible to create learning objectives with less than two levels which is not useful for grading. The current interaction creates two default undeleteable levels to sort of implicitly highlight this to the user and ensure that they do not end up with less than two levels in any objective. In my head, it made sense to me that there would be two "permanent" levels that cannot be deleted and any number of "temporary" levels that can be created and deleted as desired.
* I would be happy to discuss any alternative interactions for this with you, especially if you believe that the ability to create the default levels would be expected behavior for the instructors!

maskib:
@adishy I think it's fine this way, it might just be a little confusing to users why they can delete some LO levels and not others, but that's just me thinking of everything that could possibly be an issue. :) How bout we merge this as is, since this matches the request that was made?

app/models/learning_objective.rb
```diff
@@ -120,6 +120,25 @@ def check_unlockables(student)
...
+    maximum_proficiency = LearningObjectiveLevel.new
+    maximum_proficiency.flagged_value = LearningObjectiveLevel.flagged_values.key(0)
+    maximum_proficiency.name = "Maximum Proficiency Level"
+    maximum_proficiency.description = "Level with maximum proficiency"
+    maximum_proficiency.course_id = course_id
+    maximum_proficiency.default_level = true
+    
+    levels.push(minimum_proficiency)
+    levels.push(maximum_proficiency)
+
```
ebarroso7: @adishy not sure how this was running / put into staging while this is missing an `end` tag ?

ebarroso7:
@adishy, I like how you're creating the default levels but I feel like those levels should be able to be deleted also. 

My suggestion would be to create the levels as default, but not set them as the default levels in the database. Then your check would pull in all the levels and not let them delete the level if they're are only two levels returned. 

Let me know if you need some more clarification on that ! (:

app/models/learning_objective.rb
adishy: Could be that this was removed when a merge conflict was resolved when merging master-efs.

adishy:
* Created a new branch for this issue [here](https://github.com/UM-USElab/gradecraft-development/pull/4368)

PR #4331, opened 2019-07-10
Prevent a learning objective level from being deleted if it has been used to assess students

### LINKED ISSUE #3938: Block LOs from being deleted if they've been assessed
There's a possibility where learning objectives and levels are set up and one or more students are assessed.

If the instructor goes back and deletes a learning objective level after grading, we are now left with a "hanging" outcome with a broken reference.

Most notably, this will result in a template error for the following endpoints that are used to obtain LO outcomes:

`API::LearningObjectives::OutcomesController#index`
`API::LearningObjectives::OutcomesController#outcomes_for_objective`

If this occurs, it's possible that this person may not see any outcomes for any objective.

A question we may want to ask is - should outcomes be destroyed if the level is deleted?

adishy:
### Status
**READY**

### Description
* It should not be possible to delete a learning objective level if it has been used to assess a student (i.e. if an instance of LearningObjectiveObservedOutcome exists with a link to the level). Currently, any learning objective level can be deleted, which can lead to a "hanging" outcome where instances of LearningObjectiveObservedOutcome reference levels that do not exist.
* Now, if the level has been used to assess students, a message is displayed asking instructors to reassess students with a different level before deleting it.

### Migrations
NO

1. As an instructor, create an assignment with a linked learning objective, create a level for the learning objective. Grade a student and award the learning objective to the student with the level that was created.
2. Edit the learning objective and delete the learning objective level created. A message is shown that asks to reassess students with a different level as the level being deleted is currently being used to assess students.

* Learning objective levels controller (i.e. controllers/api/learning_objectives/levels_controller.rb)

======================
Closes #3938


adishy:
* This should also consider the case where a deactivated student has been assessed using a learning objective level that is being deleted. Currently the issue is that allowing levels that are linked to only deactivated students to be deleted leads to a hanging outcome where LearningObjectiveObservedOutcome instances reference the deleted level..

maskib:
@adishy This does prevent LO levels from being deleted if that level has been received by any number of students, but you can still delete the LO levels by deleting the LO entirely. Should we prevent this from happening as well?

adishy:
@maskib 
Good question! The ability to entirely delete the learning objective even though it has been used for assessing students should not be a problem. 
* This is as this PR actually aims to reduce database inconsistencies (i.e. invalid references to nonexistent learning objective levels) caused by deleting  learning objective levels. Currently even if a learning objective level that has been used to assess students is deleted, learning objective cumulative outcomes and learning objective observed outcomes maintain references to this deleted level. Deleting a learning objective completely deletes any related learning objective cumulative outcomes and learning objective observed outcomes for all students, and thus is not a concern for this particular problem.

maskib:
@adishy Ok thanks, that's good to know!

PR #4330, opened 2019-07-09
Allow students to delete student-logged grades for an assignment

adishy:
### Status
**READY**

### Description
* When a grade has been logged by a student for an assignment, it should be possible for them to delete the grade. Currently this is not possible and the instructor is required to delete the grade logged by the student if they want to resubmit the grade.
* Now a setting has been added under the "Advanced" tab in Course Settings that enables students to delete student logged grades for an assignment. A button that enables students to delete the grade they logged appears after they submit a grade. After they delete the grade they logged they can resubmit their grade as normal. They are allowed to repeat this as long as the assignment is open.

image

image

### Deploy Notes
* A boolean value called delete_student_logged_grade has been added to the Course model to track whether or not the instructors allow students to delete student logged grades.

### Migrations
YES

### Steps to Test or Reproduce
1. As an instructor, create an assignment and mark it as student logged. Ensure that the "Delete Student Logged Grades" setting  under the "Advanced" tab in Course Settings is checked.
2. As a student submit a grade for the assignment. Visit the submitted grade for the assignment (i.e. visit /assignments/\<id\>), you will see a button that says "Delete self-logged grade". Clicking the button will delete the grade that was submitted and will enable you to submit a grade for the assignment again.

### Impacted Areas in Application
* Course Model 
* Assignment Grades controller (i.e. controllers/assignments/grades_controller.rb)
* Course settings
* Viewing a grade as a student 

======================
Closes #4237

PR #4350, opened 2019-08-12
Change checks for select dropdown on student logged grade page to prevent errors

adishy:
### Status
**READY**

### Description
* Changes checks in behaviors/self_grade_outcomes_select.js to prevent errors. Currently, the checks caused errors when checking for the dropdown menus as they would return null. 
* Now, the dropdowns are checked with ids and the "Submit" self-logged grade button is disabled only if the raw points dropdown or the pass fail dropdown and the "Submit" self-logged grade button is present on the page.

### Related PRs
students-delete-self-logged-grade (#4330)

### Migrations
NO

1. As an instructor, create a student logged assignment and create outcome levels for the assignment.
2. As a student, submit the assignment and view the assignment page under the "My Submission" tab. It should not be possible to submit the student logged grade without selecting an outcome level

* behaviors/self_grade_outcomes_select.js

======================

### DIFF
--- a/app/assets/javascripts/behaviors/self_grade_outcomes_select.js
+++ b/app/assets/javascripts/behaviors/self_grade_outcomes_select.js
@@ -1,10 +1,13 @@
 //Selectively disables the "Submit" button for student logged assignments, if the
 //assignment has outcome levels and no outcome level is selected.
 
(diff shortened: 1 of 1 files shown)


ebarroso7:
Also merging to master-efs

PR #4368, opened 2019-09-18
Learning objectives should always have at least two levels

### LINKED ISSUE #4114: Require that a learning objective have at least two levels

adishy:
### Status
**READY*

### Description
* A learning objective should have at least two learning objective levels. Currently it is possible to have a learning objective with no learning objective levels or one learning objective level.
* Now, creating a learning objective creates two learning objective levels, one with the minimum proficiency level and one with the maximum proficiency level. Additional learning objective levels can be created and deleted as necessary however it is ensured that every learning objective will have at least two learning objective levels as deletion of learning objective levels is prevented once there are two learning objective levels remaining.

### Related PRs
branch | PR
------ | ------
`add-lo-assignment-overview` | #4108

### Migrations
NO

1. As an instructor, create a learning objective.
2. The learning objective will have two levels one with the minimum proficiency level set and one with the maximum proficiency level set. It should not be possible to delete levels if there are only two levels (it should be possible to add and delete when there more than 2 levels, but there should always be at least 2 remaining). 

* Creating learning objective (i.e. models/learning_objective.rb,  controllers/api/learning_objectives/objectives_controller.rb)
* Learning objective edit page

======================
Closes #4114

### DIFF
--- a/app/controllers/api/learning_objectives/levels_controller.rb
+++ b/app/controllers/api/learning_objectives/levels_controller.rb
@@ -49,6 +49,12 @@ class API::LearningObjectives::LevelsController < ApplicationController
       end
     end
 
+    if @objective.levels.length - 1 < 2
+      render json: { message: "Cannot delete levels as there should be at least two levels for a learning objective.", success: false },
+      status: 500
+      return
+    end
+
     @level = @objective.levels.find params[:id]
     @level.destroy
 
--- a/app/controllers/api/learning_objectives/levels_controller.rb
+++ b/app/controllers/api/learning_objectives/levels_controller.rb
@@ -49,7 +49,9 @@ class API::LearningObjectives::LevelsController < ApplicationController
       end
     end
 
-    if @objective.levels.length - 1 < 2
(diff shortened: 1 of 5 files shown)


app/controllers/api/learning_objectives/levels_controller.rb
```diff
@@ -49,6 +49,12 @@ def destroy
       end
     end
 
+    if @objective.levels.length - 1 < 2
```
ebarroso7: @adishy would prefer this to be a little more clear (not having to subtract within the conditional). 

either set a variable like `deleted_levels_count = @objective.levels.length - 1` then check for that in the conditional 

or make it check <= 2 or < 3

ebarroso7:
Screen Shot 2019-09-19 at 3 56 57 PM


Something is still a little off with this PR, when I go to create a new learning objective the levels are not there until you try to save the learning objective

PR #4235, opened 2019-04-15
Adding image button to menubar for froala

chcholman:
### Status
**IN DEVELOPMENT**

### Description
This PR fixes the image button in Froala such that students and faculty can now add in-line image uploads

### Todos
- [ ] Assess security regarding user-generated uploads
- [ ] Update/Add Documentation

### Migrations
NO

### Impacted Areas in Application
List general components of the application that this PR will affect:

* All forms with Froala

======================
Closes #4210

PR #4369, opened 2019-09-19
Upload inline images in Froala

### LINKED ISSUE #4210: Make it possible to add an inline image in Froala

adishy:
### Status
**IN DEVELOPMENT**

### Description
* It should be possible for instructors and students to upload an inline image in the Froala editor. Currently uploading images does not work and is disabled. Now, uploading images is allowed and is uploaded to a new directory in the uploads folder (files/uploads/froala_uploads) and can be inserted as inline images in the editor.

### Related PRs

branch | PR
------ | ------
image-button | #4235 

### Migrations
NO

1. As an instructor or student access a page that has the Froala editor and allows images to be uploaded. Click the image button in the editor and upload an image. 
2. The image that is uploaded should be inserted as an inline image in the editor and it should be possible to view and edit the uploaded image even after navigating away and then back to the page where the image was uploaded in.

* Uploading inline images in the Froala editor

======================
Closes #4210


adishy:
The following pages from the Froala documentation website were used:
* https://www.froala.com/wysiwyg-editor/v2.0/docs/framework-plugins/rails
* https://www.froala.com/wysiwyg-editor/docs/server/ruby/image-upload

ebarroso7:
@adishy  Not quite working on my machine. 

I started with an assignment type inline image and that seems to work, but then I go to an assignment to test there and thats where it falls off.

I'll be able to add one into the top text box "Assignment Description" and it will stay in the text box, then if I go to add one into the "Assignment Purpose" text box it will upload as if its there, then flash away in a few seconds. 

When I go to save with the assignment with an "assignment description" image inline and no image in the "assignment purpose" the assignment saves, but the image does not.  Looking at the specific  assignment page, the image put into the assignment type is shown, but not the image I put in the Assignment Description. 
When i return to edit the assignment, the image within the assignment description text box is now broken, displaying the broken image box. 

After trying to upload the image to the Assignment Description text box and looking within `files/uploads` its only showing one item within that folder (im assuming its the image that worked for the assignment_type). 

Looking at the code, i'm wondering if we can get these to go within a course folder ? like 'files/uploads/<course_id>/froala_uploads'

adishy:
Thanks for testing that!

A few other concerns:

* Security of uploads: It might be possible to deduce the URLs of files uploaded by other users
* Changing upload directory to courses: Do we want to upload files for a course in a separate folder called froala_uploads or do we want to place it within one of the subfolders too and the directory structure will need to be decided

maskib:
Not a big deal at all, i just noticed that adding an inline image can affect the format of the page underneath the editor. For example, on the badges page: 
Screen Shot 2019-10-10 at 12 16 06 PM

If the image height is great enough it pushes down the "submit" button to right above the footer, instead of there being space between them. 

Not sure if this is worth addressing?

maskib:
This mostly looks good, there is a rollbar error I'd like some feedback on though: 

https://rollbar.com/gradecraft/gradecraft-development/items/4750/

One inline image I added to a badge description disappeared, and I'm wondering if that corresponds to this error? the badge ID was 1503 in course 962. Any thoughts @ebarroso7 @adishy ?

PR #4370, opened 2019-09-20
Attachment directory fix

### LINKED ISSUE #4363: File Attachment directory testing
After the move to EFS from S3 we noticed that file attachments could not be found if the course number was changed. 

I have since changed how the directories are built on staging to remove user editable data, so they should be found regardless of user input.

ebarroso7:
### Status
**READY**

### Description
This holds the changes to find the locations of the attachments after they've been moved by the rake task.  Finishing the work in removing s3 and having a better system in place using the EFS

### Deploy Notes
Run the rake tasks first:

`rake move_attachment_directories:move_extra_attachments[true]`
`rake move_attachment_directories:delete_empty_directories[true]`

### Migrations
NO

======================
Closes #4363

### DIFF
--- a/app/controllers/courses_controller.rb
+++ b/app/controllers/courses_controller.rb
@@ -70,7 +70,6 @@ class CoursesController < ApplicationController
       change_current_course @course.id
       redirect_to action: :edit, id: @course.id and return
     end
-    @attachments = @course.has_attachments?
     authorize! :update, @course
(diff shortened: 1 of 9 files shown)


ebarroso7:
I've added `lib/tasks/course_number.rake` into master-efs but its still showing as changed in this PR.

PR #4371, opened 2019-09-20
Remove mongo merge

### LINKED ISSUE #4098: Growing Document Sizes in MongoDB Aggregates
We are writing some aggregate analytics into MongoDB and one of them is our course pageviews. For some of our larger and long-running courses, we continue to append to the same document which has caused us to reach the maximum document size (16mb) allowed by MongoDB. The effect that we are noticing in the app is that the `pageview_event_logger` job is failing in Resque.

A query of `db.course_pageviews.findOne({"course_id": 110})` can be performed in production which will return an example of an offending document.

According to the official MongoDB [docs](https://docs.mongodb.com/manual/reference/limits/#BSON-Document-Size):

>BSON Document Size The maximum BSON document size is 16 megabytes.
>
>The maximum document size helps ensure that a single document cannot use excessive amount of RAM or, during transmission, excessive amount of bandwidth. To store documents larger than the maximum size, MongoDB provides the GridFS API. See mongofiles and the documentation for your driver for more information about GridFS.

With this information, it's probably not a good idea for us to try and increase this limit. We should look into potentially using [GridFS](https://docs.mongodb.com/manual/core/gridfs/) which is intended to chunk out large documents, or investigate other potential long-term solutions.

The team behind Mongoid for Ruby appears to have already a gem for implementing [GridFS](https://github.com/mongoid/mongoid-grid_fs).

chcholman:
### Status
**IN DEVELOPMENT**

### Description
This PR removes Mongo and Resque and installs Sidekiq to handle background jobs. 

### Todos
- [x] Add Tests

### Deploy Notes
To get this to pass Codeship, we'll need to remove the references to Resque and Mongo in the build settings. To take advantage of the cost savings possible, we'll need to remove Mongo from the setup in OpenShift.

### Gem dependencies
- ADDED Sidekiq
- REMOVED Mongo, Resque, resque-scheduler

### Migrations
NO

LOCALLY:
Checkout branch, restart server, confirm that a grade awarded to a student 1) updates their total course score and 2) a job is visible as successfully processing at http://localhost:5000/sidekiq

STAGING: 
Need to do QA testing on every background job

* Analytics tracking
* Exporters
* Background jobs (score updating, predictor logging)

======================
Closes #4098, #4400  #4402, #4399


app/jobs/predictor_event_job.rb
```diff
@@ -0,0 +1,8 @@
+class PredictorEventJob
+  include Sidekiq::Worker
+  sidekiq_options retry: 10, queue: 'low'
+
+  def perform(data)
+    Analytics::Event.create({ event_type: "predictor" }.merge(data))
```
ebarroso7: @chcholman is this Analytics call in a little early ? should it be held or commented out until we merge in Ahoy work?

causes this rollbar error: https://rollbar.com/gradecraft/gradecraft-development/items/4787/?item_page=0&#traceback

ebarroso7: CH: looks like a mongo artifact 
EB: Will remove

PR #4388, opened 2019-10-11
Removed username column in sample CSV

### LINKED ISSUE #4374: Remove "username" field from the student import template .csv
Because we aren't using the username field to generate usernames (we're using the first portion (before the @ sign) of email addresses), we would like to remove that column from the .csv template for importing students to reduce unnecessary confusion for our users.

adishy:
### Status
**READY**

### Description
* The username column should not be included in the sample students import CSV as it is not used while importing students. Now, the username column has been removed from the sample students import CSV.

One thing to note about this PR is that it removes the username column from csv import, this in turn will remove the ability for umich instructors to enter in only the unique id (username) of a student. They will have to use their full email (including the '@umich.edu')

### Migrations
NO

1. Visit the users import page (i.e. /users/import) and download the sample CSV
2. The CSV that is downloaded should not include the username column

* CSV for importing users

======================
Closes #4374


maskib:
@adishy 

Currently, using this file to import users isn't working. Is this because the import still expects there to be usernames for the students in the .csv?

adishy:
@maskib 
Yup, that was exactly what it was doing, sorry I overlooked that.
It should be fixed now :)

maskib:
No problem! Looks like it working for me now too, thanks! :)

ebarroso7:
@adishy Code looks great, but lets fix up the specs in this folder /spec/importers/user_importers/csv_student_importer_spec.rb` before i merge it in

PR #4396, opened 2019-10-18
Allow students to be imported if there are capitalization differences in the student's section name

### LINKED ISSUE #4375: Students won't import if there are capitalization differences between the section name in the import .csv and in the app
### Expected Behavior
I would expect that having a section name that is identical except for capitalization would do either 1 of 2 things: it would import those students into the course in a separate section with that name, or it would add them to the existing section where the capitalization is different. 

### Actual Behavior
The students do not import. 

### Steps to reproduce
1. Create a section in your course with a title like "Section 1a"
2. Download the student import template .csv and add some students. In the "team (ie. section) column, give them a team name identical to the one already in your course, but capitalize it differently (e.g. in this case, Section 1A)
3. Import the .csv

adishy:
### Status
**READY**

### Description
* When importing students from a CSV file, the section names should be case insensitive, but currently are not. Now, the case is ignored for section names when importing students with sections 

### Migrations
NO

1.   Change the case of the section name when importing students using the CSV import template and import it as a CSV user import (i.e. /users/import)

* Importing students using a CSV file

======================
Closes #4375


ebarroso7:
One thing to note about this PR is that it removes the username column from csv import, this in turn will remove the ability for umich instructors to enter in only the unique id (username) of a student. They will have to use their full email (including the '@umich.edu')

PR #4415, opened 2019-11-20
Ahoy!

chcholman:
### Status
**READY**

### Description
This branch adds a new utility to track events within the application. Ahoy is highly flexible, while also being quite simple in its approach. To inspect events in the console, use either `Ahoy::Event` or `Ahoy::Visit` as the starting point.

### Todos
- [x] Test tracking in staging
- [x] re-add mailer link tests
- [x] track user_id on visits
- [x] Update/Add Documentation https://github.com/UM-USElab/Developer-Documentation/wiki/Ahoy-Tracking 
- [ ] Update Sample Data

### Gem dependencies
- ADDED ahoy_matey
- ADDED ahoy_email

### Migrations
YES

Log into the application, confirm that there's a new Ahoy::Visit record. Same for submitting an assignment, getting a grade, logging out, predicting things. 

* All pages. Analytics, phew!

======================


app/controllers/application_controller.rb
```diff
@@ -88,7 +88,9 @@ def event_session
   protected
 
   def track_action
-    ahoy.track "Application Event", request.path_parameters
+    if !request.format.json?
+      ahoy.track "Application Event", request.path_parameters
```
chcholman: 💯

spec/mailers/notification_mailer/earned_badge_awarded_spec.rb
```diff
@@ -78,10 +78,6 @@
         it "declares a doctype" do
           should include "DOCTYPE"
         end
-
-        it "contains the badges url" do
```
chcholman: curious - why is this removed in this PR?

spec/mailers/exports_mailer/submissions_export_spec.rb
```diff
@@ -59,7 +59,6 @@
     describe "html part body" do
       subject { html_part.body }
       it_behaves_like "a complete submissions export email body"
-      it_behaves_like "a submissions export email with archive data"
```
chcholman: Why is this removed in this PR?

spec/mailers/notification_mailer/earned_badge_awarded_spec.rb
ebarroso7: @chcholman  - When an email is sent with tracking ahoy ( I think by default ) the ahoy::message has attributes about what / if links are clicked within the email. To accomplish this extra tracking, ahoy changes the url links within the html body of an email to have a really long string. 

So when the test looks for the badges url  "/badges" it fails because the link in the html body of the email is actually something like "/badges/1wzu9asdfqasdfijo9p50iu" 

The link to the badge page still works in the email. 

Also .text (apposed to .html) emails are unchanged by ahoy message tracking so that test is not removed.

spec/mailers/exports_mailer/submissions_export_spec.rb
ebarroso7: I was seeing the same results as described above for badges & the test for a badge URL for this email & test

ebarroso7:
Currently facing the same issue as identified here: https://github.com/ankane/ahoy_email/issues/66 

but have been unable to implement their solution ):

PR #4430, opened 2020-01-03
Score recalculator

ebarroso7:
### Status
**READY**

### Description
The goal of this PR is to add more stability to our score recalculation. To do this I added a job that runs weekly to trigger the score recalculator job on all active courses, and an additional check for challenge grades to the nightly score recalculator job. 

It is hard to see what is happening in this PR with the job Class files because they were renamed more than re-written shown in the comparison of these files. 

For clarity:
"RecalculateActiveCourseScoresJob" was renamed to "RecalculateRecentlyUpdatedScoresJob"  

#### Notes on the new, more accurately named "RecalculateRecentlyUpdatedScoresJob" : 

1) Complexity of the `updated_challenge_grades.each` loop
- Needed to get the students ID and course ID to add to the `unique_grades` array 

2) Within the challenge grade model a new method was made called `contributes_to_student_score?`
- Making this method gave me the idea of moving the attribute `add_team_score_to_student` from the course to the challenge 
- This would require more work on the front end and migrations to change the attribute location
- Would give instructors more opportunities to use this feature instead of always or never 

#### Notes on the new "RecalculateActiveCourseScoresJob" 
1) Runs on **all** active courses
2) Does **not** include deactivated students
3) Current active courses count:
- staging: 321
- umich: 793
- app: 2,300

### Migrations
NO

If you'd like to trigger these jobs, `foreman start` then open the rails console and run: 

`RecalculateActiveCourseScoresJob.perform_async` 
`RecalculateRecentlyUpdatedScoresJob.perform_async`

Reset local DB to have grades updated within the last 24 hours with: 
Rake db:drop
bundle exec rails db:create
bundle exec rails db:sample

### Times jobs are scheduled to run: 
RecalculateRecentlyUpdatedScoresJob -- every morning 2 am

RecalculateActiveCourseScoresJob -- every Saturday morning at 3 am


ebarroso7:
Work on allowing the syllabus to be removed was done on this branch then cherry-picked onto master-efs so should be ignored / not included in this PR -- its already on branches master-efs, production, and beta.gradecraft.com

app/models/challenge_grade.rb
```diff
@@ -17,6 +17,11 @@ class ChallengeGrade < ApplicationRecord
   validates_presence_of :team_id
   validates_numericality_of :raw_points, :final_points, :adjustment_points, allow_nil: true, length: { maximum: 9 }
 
+  def contributes_to_student_score?
```
chcholman: 👍

app/jobs/recalculate_active_course_scores_job.rb
```diff
@@ -1,24 +1,18 @@
 require "sidekiq-scheduler"
```
chcholman: Really glad we're cleaning these up and making the naming clearer

app/views/courses/_course_details.haml
```diff
@@ -35,6 +35,8 @@
         = f.file_field :syllabus
         - if current_course.syllabus.present?
           .italic= link_to "Current Syllabus", current_course.syllabus_url
+          = f.check_box :remove_syllabus
```
chcholman: nice, helpful thing to add!

PR #4446, opened 2020-02-11
Export assignment submissions in a specified date range in the course data exports page

### LINKED ISSUE #4315: It should be possible to export a specified date range for certain course data export types

adishy:
### Status
**IN DEVELOPMENT**

### Description
* When exporting assignment submissions for some courses, the data to export is too large and thus it is useful to specify a date range which can be used as a filter to specify which assignment submissions to export. 
* Now, a form has been added in the /downloads page for the "Assignment Submissions" download link to download only a specified date range if desired (or all the data if no dates are specified)

### Migrations
NO

1. As an instructor, visit the /downloads page, and for the "Assignment Submissions within a date range" link, specify a date range and a column to apply the date range to
2. Click the link and open the attachment in the email. The data exported should be within the date range for the specified column

* Course data exports (i.e. /downloads)
* Added a script for the form for the "Assignment Submissions" link (assets/javascripts/downloads.js)
* SubmissionListExporter now uses a date range
======================
Closes #4315


app/assets/javascripts/downloads.js
```diff
@@ -0,0 +1,71 @@
...
+
+        var EndDate = _Input.value;
+
+        StartDateInputs.forEach(function(_Input){
+            var MinimumStartDate = new Date(EndDate);
+            MinimumStartDate.setDate(MinimumStartDate.getDate() - parseInt(_Input.getAttribute("date_range_days")));
+            _Input.setAttribute("min", MinimumStartDate.toISOString().split("T")[0])
+        })
+    });
+});
```
ebarroso7: Would be great to sit down and work through this file together, I'm not exactly sure what all is going on in here. 

It looks like you might be trying to set up the date inputs from the JavaScript side to a date time format Ruby can read ? I think we might be able to find some built in JS functions that convert it to a universal timestamp that would then be able to be read by ruby's Date class.  To hopefully reduce complexity and the nested loops. 

A general critique I have about this file is to maybe name it more specifically (even though i know the page is just called "downloads") I think something like export_downloads.js would be more helpful. This would then have to get changed in the application.js file and maybe the bottom of this file also app/views/downloads/index.html.haml ?

app/assets/stylesheets/pages/_downloads.sass
```diff
@@ -1,2 +1,26 @@
 .download_type
   margin-top: 30px
+
+.calendar_indicator i
+  position: absolute
+  margin-top: 17px
+  margin-left: 5px
+  padding: 0.5px 0px 0px 0px
+
+.datetimepicker
```
ebarroso7: Is it possible to do something for this page that would use a datetime picker that more closely resembles the others in the application? I'm thinking like the one on the course settings page for term start and end.

app/controllers/info_controller.rb
```diff
@@ -122,9 +122,48 @@ def multiplier_choices
     @teams = current_course.teams
   end
 
+  def start_end_dates(course)
```
ebarroso7: Another file we can work on together on along with `downloads.js`. 

Should be able to clean it up pretty quickly, mostly removing the 30 days permitted constraint

config/environments/development.rb
```diff
@@ -79,7 +79,7 @@
   # Debug mode disables concatenation and preprocessing of assets.
   # This option may cause significant delays in view rendering with a large
   # number of complex assets.
-  config.assets.debug = true
```
ebarroso7: I know I rushed you to push this and make the PR so i'm just noting this as something we'll want to remember to change back :)

app/assets/javascripts/downloads.js
adishy: * I have renamed the file as you suggested to export_downloads.js in the current version and we now use Jquery UI datepickers!
* The core purpose of this code is to append GET parameters to the submission list download link, I think this was done so as to minimize changes to the exports page/template.

app/assets/stylesheets/pages/_downloads.sass
adishy: * The Jquery UI datepickers are already styled so the datepickers should now resemble ones in other parts of Gradecraft.

app/exporters/submission_list_exporter.rb
```diff
@@ -1,8 +1,14 @@
...
-  def export(course)
+  def export(course, start_date, end_date, field)
+    puts "Start: #{start_date}"
+    puts "End: #{end_date}"
+
     CSV.generate do |csv|
       csv << baseline_headers(course)
-      course.submissions.submitted.find_each(batch_size: 500) do |submission|
+      
+      query = (field == "graded_at") ? course.submissions.submitted.includes(:grade).where("grades.graded_at" => start_date.beginning_of_day..end_date.end_of_day).references(:grade) : course.submissions.submitted.where(field => start_date.beginning_of_day..end_date.end_of_day)
```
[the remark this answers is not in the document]
adishy: Removed the "graded at" field and any queries related to it in the current version!

PR #4489, opened 2020-04-22
Webpacker asset pipeline

ebarroso7:
### Status
**HOLD**

### Description
EB: Creating PR to compare how Jon attempted to add webpack / webpacker


ebarroso7:
Using a different approach to get Vue into the application that does not require removal of the asset  pipeline

PR #4512, opened 2020-06-03
Added Scores to Learning Objectives Tab in Assignments

### LINKED ISSUE #4459: Score column in learning objectives tab of a graded assignment
### Expected Behavior

Feature request from Laurie at TG.

Want to be able to determine if an assignment has been graded but a proficiency for learning objective has not been awarded. 

To do this, we were thinking of adding a column that shows the student's score for the assignment. This way a grader would know if there was a score, but no proficiency awarded that they should edit the grade and add a proficiency level. I would also expect that a pass/fail assignment says "pass" or "fail" (course term_for pass/fail) in this column. 

Should make sure that both proficient & not proficient markers are displayed in the learning objectives column, and that this is displayed if there are multiple learning objectives associated with one assignment. 

Something else I'm thinking about is if we should hide / show the score if the grade is not yet visible to the student -- is the score displayed if the grade is not yet visible to student ? 

### Actual Behavior

Currently, nothing is displayed next to a student who has been graded but not awarded a proficiency level for the associated learning objective.

Screen Shot 2020-02-20 at 10 59 37 AM

### Steps to reproduce

- Link an assignment to a learning objective
- Go to the assignment page
- Go to Learning objectives tab

cfengjin:
### Status
**READY**

### Description
Added assignment and appropriate functions to LOInstructorOverviewCtrl so the LO tab can access student scores. Added score column in LO tab (only if LOs exist for that assignment).

Also removed outcomes in the same haml page since there is no matching function in the respective Angular directive.

### Migrations
NO

See link to issue.

Visibility of scores to instructors in learning objectives tab of every assignment.

======================
Closes #4459


app/assets/javascripts/angular/templates/learning_objectives/instructor_overview.html.haml
```diff
@@ -26,7 +30,11 @@
           %a{"ng-href"=>"{{student.student_path}}"} {{student.first_name}}
         %td
           %a{"ng-href"=>"{{student.student_path}}"} {{student.last_name}}
-        %td{"ng-repeat"=>"outcome in loInstructorOverviewCtrl.outcomes(student.id) track by outcome.id"} {{outcome}}
```
ebarroso7, line 29: @cfengjin I see this is the line you're referencing as the column you deleted but just want to confirm that it also looks like there is not a header associated with the column? So that doesn't need to be deleted because it was never there? 

I can't really tell what this is trying to do, but maybe past developers went with the `outcomeForObjective` method to achieve what they set out to do with this, thoughts?

cfengjin: It does seem that `outcomeForObjective` was ultimately chosen to display the outcomes for each learning objective instead of `outcomes`.

I could've been clearer, but the primary reason I deleted that column was because the `outcomes` method doesn't exist in `loInstructorOverviewCtrl` and it didn't produce anything when I was testing in my browser. There also wasn't a header associated with the deleted column, as you mentioned.

PR #4521, opened 2020-06-18
Group Rejection and Formation

dbrucknr:
### Status
**IN DEVELOPMENT**

### Description
PR attempts to remedy an issue located here:

https://trello.com/c/TUHL6TdY/72-rejecting-a-group-as-an-instructor-does-not-allow-students-to-form-a-new-group

I have added a button that allows a student to view the status and/or feedback of the instructor of their proposed group.

Development TODO - test and understand the rejection pattern in place

### Todos
- [ ] Add Tests
- [ ] Update Sample Data

### Migrations
NO

In the Trello card, there exists a series of steps I used to find the groups/:id page. Those pages are what is being referenced for feedback.

* assignments
* groups

======================
Closes # N/A


app/views/assignments/_description.haml
```diff
@@ -38,8 +38,10 @@
 
 - if current_student.present?
   - @group = current_student.group_for_assignment(presenter.assignment)
+
   - if presenter.group_assignment?
-    - if @group.present?
+    - if @group.present? && @group.approved != 'Rejected'
```
ebarroso7, line 43: @dbrucknr This is the line I am referring to. Seems like we might want something like `@group.approved == 'Rejected'`   

You'll also want to check what the partial "groups/status_display" renders that is just on line 45 below this.  To me that looks like it would be the little button / div span thing that says on it "approved" or "rejected" and maybe thats just not showing up properly because the condition `@group.present?` doesn't hit it? -- I could be wrong those are just my thoughts right now. 

yeah... when i look, at app/views/groups/_status_display.haml it seems to have something trying to do this in the line `- elsif @group.rejected?`  where I don't think we actually have any method `.rejected?` on a group. so that line could be changed to the `@group.approved == 'Rejected'` (or whatever you find works for that) 

Remember the general issue was saying theres no way for students to know that the group proposal has been rejected, but also no way for them to resubmit another group proposal. 

If thats something they can do from the group status page (the link you've added on line 44) then maybe that link should be moved into the Rejected conditional in the app/views/groups/_status_display.haml file.

app/views/groups/_status_display.haml
```diff
@@ -1,13 +1,16 @@
...
+
+- elsif @group.approved == 'Rejected'
   .alert-box.alert
     %i.fa.fa-times-circle.fa-fw
     %span.bold Status: Rejected
+  %div
+    = link_to glyph(:users) + "Create Another #{term_for :group}", new_group_path, class: "button"
 - else
-  .alert-box.notice
+  .alert.box.notice
```
ebarroso7, line 13: I'm guessing this should still be `alert-box` not `alert.box`

app/views/groups/_status_display.haml
```diff
@@ -1,13 +1,16 @@
 - if @group.approved?
-  .alert-box.success
+  .alert.box.success
```
ebarroso7, line 2: I'm guessing this should still be `alert-box` not `alert.box`

app/views/assignments/_description.haml
```diff
@@ -38,9 +38,10 @@
 
 - if current_student.present?
   - @group = current_student.group_for_assignment(presenter.assignment)
+  = render partial: "groups/status_display", locals: { group: @group }
```
ebarroso7, line 41: Looks like you're just a little off in the location of this line... this will try to display the group status partial, but you're only within the conditional that checks if the current_student is present. 

move this down beneath ` - if @group.present?` and it will only display the partial for assignments that are marked as group assignments

app/views/groups/_status_display.haml
dbrucknr: This class issue was causing some trouble - I could see the div in the console, and in the elements on the page, but it wouldn't render in the page.

app/views/assignments/_description.haml
dbrucknr: I will make adjustments soon! Thanks Erik!

PR #4525, opened 2020-06-25
Add conditional that catches the #destroy method event for a course

dbrucknr:
### Status
**IN DEVELOPMENT**

### Description
I added an 'unless' statement that catches a context in which a course destroy event was returning a frozen hash and breaking the course destroy event.

The issue is also referenced in the QA Trello board at this link:
https://trello.com/c/Prsjiujp/91-trying-to-delete-a-course-with-teams-results-in-an-error-and-fails-to-delete

### Todos
- [x] Add Tests
- [x] Make sure other team functions remain intact

### Migrations
NO

* courses - destroy event(s)
*teams

======================


dbrucknr:
The unless statement I added does appear to break other aspects of the ranking - when a team is destroyed, rather than a course, the ranks no longer work.

PR #4527, opened 2020-07-02
increasing time to timeout

chcholman:
### Status
**IN DEVELOPMENT**

### Description
We've had GC timing people out after 60 minutes and this has produced a variety of issues for learners and instructors. We'd like to change this to 24 hours. 

### Migrations
NO


chcholman:
@ebarroso7 I believe this is the right line, but potentially in the wrong file - the dockerfile references `RUN cp ./secrets/production.rb /gradecraft/config/environments/production.rb`. but this file doesn't exist. Thoughts on where we should be making this change?

